PR #11, opened 2017-02-17
Nullptr is casted to false in modern compillers

patrickelectric:


lovettchris:
Our coding guidelines require we spell out the nullptr equality tests.  Can you provide more info why this changes is needed?

patrickelectric:
@lovettchris it's just a matter of readability and because nullptr have an implicit conversions to bool.
If the coding guidelines require that you can close this PR.

ps: Where is the guidelines and style conventions ?

lovettchris:
We are mostly following these guidelines: https://google.github.io/styleguide/cppguide.html.

PR #16, opened 2017-02-18
Add rpclib as a submodule and compile directly from source

8W9aG:
This fixes the OSX linker issues when dealing with rpclib.

* Removes the binary distributions of rpclib and replaces them with a source compile
* Adds a submodule to rpclib

As a sidenote, I'm not sure whether you want to use the git subtree or submodule workflow here.

PR #43, opened 2017-02-22
[WIP] Linux Build

bkueng:
This adds linux build support, including a build script that handles the boost and eigen dependency.

There are symbol inconsistencies due to the [GCC Dual ABI](https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html) with respect to C++11 between GCC and Clang (or more importantly libc++ and libstdc++). And since the Unreal Engine builds with clang, we should make sure that we build everything with clang here as well to avoid running into problems. This also means we need to build boost from source and cannot use the system's installed version. The build script takes care of that.

I needed to add rpclib to the unreal build script, because there were unresolved symbols otherwise (https://github.com/Microsoft/AirSim/commit/6157b2e474b449483cda94988af80cbbdc3c79cb). I wonder how this could have worked on Windows without it?

To make this complete, it needs handling of the rpclib. I'm waiting for https://github.com/Microsoft/AirSim/pull/16 to be merged first.

I tested this on Fedora, but should work on Ubuntu too.


sytelus:
I've merged this PR. However, please note that we may have broken previous Linux build while making a change so that rpclib is now submodule. There is a POST_BUILD command in CMakeLists.tst that should copy build output of rpclib to AirLib/deps but needs to be tested yet.

PR #55, opened 2017-02-26
Fixes OS X issues when attempting to build

8W9aG:
This PR fixes various issues encountered when attempting to build AirSim for OS X, and borrows some ideas from this PR: https://github.com/Microsoft/AirSim/pull/16

* Fixes cmake syntax issues
* Fixes commands being used to copy files around (and defining the target from an invalid place according to cmake)
* Links to rpclib from the cmake project
* Fixes various issues in MavLinkCom not compiling sources or using the wrong arguments in the Utils.cpp


lovettchris:
Nice fixes, thanks.

lovettchris:
I had to make some fixes to get Linux build to work, please test to make sure I didn't break anything in the OSX build.  Specifically, I moved the EXECUTABLE_OUTPUT_PATH so all binaries are now in cmake/output/bin, and I now include CommonSetup() in the top level CmakeLists.txt so that AIRSIM_ROOT is defined.

PR #67, opened 2017-02-28
Some updates in build.sh

patrickelectric:


lovettchris:
Probably should also check that clang and clang++ are installed right?

patrickelectric:
@lovettchris done

Issue #36, opened 2017-02-21
Add support for Mac OSX


kentaccn:
Any support for mac dev env?

sytelus:
Please see this:https://github.com/Microsoft/AirSim/pull/15

8W9aG:
This makes it compile again: https://github.com/Microsoft/AirSim/pull/55

sytelus:
Cool! Are you able to run sim on OSX now?

PR #15, opened 2017-02-18
Add OSX compiler support

8W9aG:
This introduces some minor changes which enables AirSim to be compiled on OSX successfully.

Tested on:
Xcode 8.2.1
OSX 10.12.3

I have yet to test whether this works, but thought I'd get that going after this is in.

PR #80, opened 2017-03-05
Workaround for missing sem_timedwait on os x

ghexp:
Follows MIT licensed http://stackoverflow.com/a/37324520
Trying to keep #15 and #36 going

### DIFF
--- a/MavLinkCom/include/MavLinkSemaphore.hpp
+++ b/MavLinkCom/include/MavLinkSemaphore.hpp
@@ -5,6 +5,10 @@
 #define MavLinkCom_Semaphore_hpp
 #include <memory>
 
+#ifdef __APPLE__
+#include <semaphore.h>
+#endif
+
 namespace mavlinkcom
 {
 	/*
@@ -27,13 +31,16 @@ namespace mavlinkcom
 		void wait();
 
 		// wait for a given number of milliseconds for one call to post.  Returns false if a timeout or EINTR occurs.
-		// If post has already been called then timed_wait returns immediately decrementing the count so the next 
+		// If post has already been called then timed_wait returns immediately decrementing the count so the next
 		// wait will block.  Throws exception if an error occurs.
 		bool timed_wait(int milliseconds);
-	private: 
+	private:
 		class semaphore_impl;
 		std::unique_ptr<semaphore_impl> impl_;
 	};
 }
 
+#endif
+#ifdef __APPLE__
+int sem_timedwait(sem_t *sem, const struct timespec *abs_timeout);
 #endif
--- a/MavLinkCom/src/MavLinkSemaphore.cpp
+++ b/MavLinkCom/src/MavLinkSemaphore.cpp
@@ -8,6 +8,10 @@
 using namespace mavlinkcom;
 using namespace common_utils;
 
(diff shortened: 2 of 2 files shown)


clovett (commented):
Thanks for that.

lovettchris:
Actually, I found a simpler implementation using mach, see [http://pastebin.com/pm78BJPi](http://pastebin.com/pm78BJPi).

Turns out mach has semaphore_timedwait.  Can we use this instead?

ghexp:
Found that one too, but appears to be from [GPL project](https://github.com/tumi8/vermont/blob/master/LICENSE) Vermont:
https://github.com/tumi8/vermont/blob/master/src/osdep/osx/sem_timedwait.h
https://github.com/tumi8/vermont/blob/master/src/osdep/osx/sem_timedwait.cpp

lovettchris:
[http://pastebin.com/pm78BJPi](http://pastebin.com/pm78BJPi) is just a snippet showing how to use a system call, can you verify this works on OSX, if so, I will switch to using semaphore_timedwait using the mach mach_timespec_t.

ghexp:
Code sample compiles. If you're switching, note that sem_init / sem_destroy (unnamed semaphore) compiles but is also unsupported at run time on OS X, requiring named semaphores instead sem_open/sem_close.   
http://stackoverflow.com/questions/1413785/sem-init-on-os-x
http://blog.superpat.com/2010/07/14/semaphores-on-linux-sem_init-vs-sem_open/

clovett:
Interesting, that stackoverflow also recommends…

https://developer.apple.com/reference/dispatch/1452955-dispatch_semaphore_create

I wonder which one works better on OSX…


Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

ghexp:
dispatch_semaphore_create is tied to dispatch queues and tasks which are even less POSIX than the limited support for the standard that OS X already provides. "Because Grand Central Dispatch manages the relationship between the tasks you provide and the threads on which those tasks run, you should generally avoid calling POSIX thread routines from your task code." 

https://developer.apple.com/library/content/documentation/General/Conceptual/ConcurrencyProgrammingGuide/ThreadMigration/ThreadMigration.html#//apple_ref/doc/uid/TP40008091-CH105-SW18

clovett:
Got it, sounds like we need an implementation of MavLinkSemaphore that works with named semaphores for OSX then…

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

lovettchris:
I found yet another set of semaphore classes on OSX:

https://developer.apple.com/reference/coreservices/1585569-mpcreatesemaphore
https://developer.apple.com/reference/coreservices/1585586-mpdeletesemaphore
https://developer.apple.com/reference/coreservices/1585722-mpwaitonsemaphore
https://developer.apple.com/reference/coreservices/1585713-mpsignalsemaphore
...

So I added code for that, just wish I had a way to test it...

See change #[4c0df2fe97f692732dd6514b9d696497b95e65d4](https://github.com/Microsoft/AirSim/commit/4c0df2fe97f692732dd6514b9d696497b95e65d4)

PR #98, opened 2017-03-09
Fix forward declaration scope issues in semaphore class on macOS

jamuus:
This fixes building on macOS using gcc-6 for me. Not sure if this is how it should be.


lovettchris:
Thanks, I prefer a different fix to this which makes the OSX code consistent with the other platforms.  See [this change# 462021081babae6e26035a6c6d43d76d17d1e637](https://github.com/Microsoft/AirSim/commit/462021081babae6e26035a6c6d43d76d17d1e637).

PR #110, opened 2017-03-14
DroneServer: Fix settings load

meee1:
this fixs the settings load


lovettchris:
Hey, cool, thanks for fixing that !  Things are a bit crazy at the moment... juggling too much, need more coffee :-) Or maybe it was too much coffee :-)

PR #162, opened 2017-04-17
Multithreaded approach for recording camera images

saihv:
This PR proposes modifications to the existing camera recording process by moving it to a separate thread.

The native Unreal implementation of ReadPixels() contains a call to the function FlushRenderingCommands(), which effectively blocks the game thread until the camera image has been rendered, thus reducing the frame rate to unplayable levels. To get around it, this version has a separate thread created as an Unreal FRunnableThread type. Inside this thread, the drone's camera is accessed periodically, and the FPV render command is queued into the render thread. It also has a custom implementation of a 'RenderCommandFence', which checks on the status of this render, and once it's done, compresses the image and saves the screenshot to the usual log folder, along with the ground truth (similar to the earlier implementation). All of this is fully Unreal based; so I expect it to work on other OS's as well, but I haven't tested it. 

On my computer, (i5-6600K, Nvidia GTX 1070, Win 10) this results in ~10 Hz of image logging rate whereas the game runs at an almost constant ~120 FPS (in a very simple environment) while recording.

Changes to existing code: 

1. HDR has been disabled in the FPV render target settings. This halves the size of the texture from ~7 MB to ~3.5 MB, thus making the render process quicker, whereas leaving HDR on seems to be putting too much load on the GPU. Also, the gamma in image settings has been bumped up to fix the lighting (caused by turning off HDR). (Still checking on whether a more optimized version can allow HDR to be on)

2. Methods such as APIPCamera::getCaptureComponent() and ASimModeBase::getPhysicsBody() have been exposed as public so that the new thread can access them. 

Please let me know if you have any suggestions to improve this further. 

EDIT: The first few commits on this branch can be ignored, the repo's been merged with a recent version before these new modifications were added.


sytelus:
@saihv - this is great contribution. Thanks!

PR #197, opened 2017-05-03
Fix compiler warnings

andre-nguyen:
Basically took it from here http://stackoverflow.com/questions/3417837/what-is-the-best-way-to-supress-unused-variable-x-warning

The 3 ways of silencing the unused param warnings are:
1. (What I did) Not giving a name to the parameter. We can keep it as a comment though for documentation purposes. 
2. [Define a template function that does nothing](https://herbsutter.com/2009/10/18/mailbag-shutting-up-compiler-warnings/) `template<class T> void ignore( const T& ) { }`
3. Cast to void (and maybe define a macro do to is). `(void)variable;`

The change to `RecordingThread.cpp` is for a warning about having `&&` in a `||`.

Edit: Looking at the Travis log it looks like I missed some, but its a start!

### DIFF
--- a/AirLib/include/common/StateReporter.hpp
+++ b/AirLib/include/common/StateReporter.hpp
@@ -52,7 +52,7 @@ public:
 
     //write APIs - heading
     //TODO: need better line end handling
-    void startHeading(string heading, uint heading_size, uint columns = 20)
+    void startHeading(string heading, uint /*heading_size*/, uint /*columns*/ = 20)
     {
         ss_ << "\n";
         ss_ << heading;
--- a/AirLib/include/common/UpdatableObject.hpp
+++ b/AirLib/include/common/UpdatableObject.hpp
@@ -16,7 +16,7 @@ public:
     virtual void update() = 0;
     virtual ~UpdatableObject() = default;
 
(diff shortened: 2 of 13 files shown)


lovettchris:
Thanks for attempting to fix these annoying compiler warnings.  Unfortunately, this particular fix causes C++ intellisense to be less useful:

image

Instead I would prefer the following to be added to a common header, let's use common/common_utils/Utils.hpp:
`template <class T> inline void unused(T const & result) { static_cast<void>(result); }`

This template seems to work better across all the compilers we care about (CLang, GCC and VC++).

And please remove this separate fix: 
`Unreal/Plugins/AirSim/Source/RecordingThread.cpp `
I need to have a conversation with Shital first about what is the correct logic here...

andre-nguyen:
Sure, let me know when you confirm which method you want to use.

Closing for now.

lovettchris:
The template <class T> inline void unused(T const & result) { static_cast<void>(result); } is the way we want to fo.

My other comment is about a separate compiler warning you were fixing in Unreal/Plugins/AirSim/Source/RecordingThread.cpp that was not related to unused parameters.

PR #200, opened 2017-05-04
Fix unused parameter warnings in AirLib

andre-nguyen:
@lovettchris As discussed. Should fix all the ones in AirLib, I didn't touch the ones in MavLinkCom.

Wasn't quite sure if you wanted `unused` to be in the `common_utils` namespace though.


lovettchris:
Thanks!

PR #203, opened 2017-05-08
Update linux_build.md

Mayankm96:
The previous instructions lead to confusion while using cmake for AirSim


lovettchris:
Thanks for the idea, I prefer not to assume AirSim is in $HOME.  So I found another way to expand the help around running getlibcxx.  It now says this:

Next you will need the latest version of libc++ library from llvm. You can get that by running the following script which you will find in the `cmake` folder that is included in your AirSim repo:

````
getlibcxx.sh
````
This will clone the llvm-source and build it in an llvm-build folder, then it will install the built libraries.  This ensures you have the right version of llvm for your platform.

Issue #212, opened 2017-05-12
Unable to calibrate the Joystick Controller


Mayankm96:
I am not sure whether I am missing something but the documentation seems a bit unclear for me. I am trying to use an emulator for Logitech Rumblepad 2. The controller works fine since I used it to try out a game on the PC. However, when I am trying to calibrate the controller on QGroundControl, I have been unsuccessful with it so far. 

```
Launch QGroundControl and you should see a new Joystick tab under stettings:

<image> 

Now calibrate the radio, and setup some handy button actions. For example, I set mine so that the 'A' 
button arms the drone, 'B' put it in manual flight mode, 'X' puts it in altitude hold mode and 'Y' puts it in
position hold mode. I also prefer the feel of the controller when I check the box labelled "Use exponential 
curve on roll,pitch, yaw" because this gives me more sensitivity for small movements.]
```

Why do I need to calibrate the radio if I am using a joystick? And more importantly how should I go about it?

andre-nguyen:
>Why do I need to calibrate the radio if I am using a joystick? 

Because just like a regular RC, QGC needs to know the min/max of all channels as it will differ from one gamepad to another.

>And more importantly how should I go about it?

Starting from scratch:
1. Launch px4 SITL in a terminal, it will wait for a UDP connection on a certain port.
1. Launch Unreal Editor and press play. Once the quad spawns, a connection should be made to px4 SITL which will unblock the autopilot and start executing.
1. Start QGC (caution tricky)
    1. Click the white/purple Q in the top right to open the settings.
    1. Add a UDP comm link on port 14540 with the localhost ip (127.0.0.1)
    1. Connect using that link. You should see at the top the sattelite count become valid and the battery voltage too. If you don't see this, then stop now because the Mavlink connection is invalid.
    1. Click the gears icon next to the Q icon and you should get a bunch of menus including the joystick configuration menu.
    1. The rest should be pretty straight forward, click calibrate and follow the instructions on the screen.

Flying:
1. Assign a button to arming and disarming.
2. Click on "Manual" at the top right and change the mode to "Position".
3. Arm by pressing the button from step 1.
4. Throttle up to take off.

Mayankm96:
Is there a way to build PX4 SITL in Windows 8.1? BashOnWindows seem to require Windows 10 and I have seen the documentation on the PX4 website but it seems to not work. 

I am really sorry for bothering so much with this. Thank you for your help!

lovettchris:
[...]

PR #224, opened 2017-05-18
Update xbox_controller.md

Mayankm96:
* Adding more description to how to use the contoller as shown in issue #[212](https://github.com/Microsoft/AirSim/issues/212)
* For HIL, you don't need a PX4 connected to the serial port which is a bit confusing as the MavLinktest step is redundant for that. The PX4 SITL works fine. 
*  The XBox360ce might need to be checked.

### DIFF
--- a/docs/xbox_controller.md
+++ b/docs/xbox_controller.md
@@ -1,12 +1,58 @@
 # XBox Controller
 
-To use an XBox controller with AirSim follow these steps:
+To use an XBox Controller with AirSim follow these steps:
 
 1. Connect XBox controller so it shows up in your PC Game Controllers:
 
 ![Gamecontrollers](images/game_controllers.png)
 
-2. Launch QGroundControl and you should see a new Joystick tab under stettings:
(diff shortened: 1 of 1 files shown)


andre-nguyen:
> For HIL, you don't need a PX4 connected to the serial port which is a bit confusing as the MavLinktest step is redundant for that. The PX4 SITL works fine.

Careful, HIL means hardware in the loop so by definition you need a pixhawk connected to be doing HIL simulations :wink:

Mayankm96:
@andre-nguyen.. That's what I thought too! But so far I haven't connected the PX4 and it's running just fine for me :P

andre-nguyen:
It's running "fine" because you aren't doing HIL, you are doing SITL. They are two different things.

lovettchris:
I don't really agree with all these edits.  And it is too redundant with px4.md and sitl.md.  The xbox controller page was trying to provide only information relevant to setting up xbox controller.

Issue #252, opened 2017-06-02
MavLinkCom does not work on OS X


nacho-carnicero:
Hi, 

I've been trying to use the MavLinkCom library on my Mac without any success. I wanted to test the MavLinkTest main program and check if it worked. I compiled it by performing `cmake MavLinkCom` and then `make` and then ran the executable that was created in `cmake/output/bin`. Nevertheless when I try to connect to a simulation that is running (PX4 SITL and jMAVSim) through a UDP client socket with flag `-udp:127.0.0.1:14550` I get a socket error as follows:
```bash
./MavLinkTest -udp:127.0.0.1:14550
Connecting to offboard drone at address 127.0.0.1:14550
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: MavLinkConnectionImpl: Error sending message on connection 'drone', details: UdpClientPort socket send failed with error: 56
```

Error code 56 is issued when the socket is already connected and you try to connect it again (you can check it [here](https://www.ibm.com/support/knowledgecenter/SSLTBW_1.13.0/com.ibm.zos.r13.cs3cod0/syserret.htm)).

This does not happen if I connect through a `-server` connection.

Besides, I've tried to compile and execute this on both Windows and Linux, and it works just fine with both client and server connections. I guess then it's an OS X specific issue. I've been working on a fix, I have to test it and I will issue a Pull Request for this

lovettchris:
Is another app already using port 14550?  For example, QGroundControl ?

nacho-carnicero:
No, no app is using it. I've issued Pull Request #253 , I really think it's something specific to OSX and the way it handles UDP client sockets.

nacho-carnicero:
Also I was thinking, since MavLinkCom can be used on its own for other projects, maybe it would be useful to put all the MavLinkCom cmake files in that folder so that it is more self-contained?

lovettchris:
The only reason I didn't do that is so folks didn't accidentally clobber the hand crafted visual studio *.vcxproj files by running cmake on windows.  But I don't feel too strongly about it.  

Can you run MavLinkTest app from terminal and talk to PX4 using -serial:/dev/ttyACM0 ?  Then can you run MavLinkTest with  -serial:/dev/... and -proxy 127.0.0.1:14550 and connect to QGroundControl without AirSim ?

lovettchris:
Oh, saw your pull request, looks like you fixed it then? Thanks.

nacho-carnicero:
[...]

PR #253, opened 2017-06-02
Fix OSX specific UDP socket issue

nacho-carnicero:
This solves Issue #252.

The main problem here was that when the socket is created, it is connected to the server through the `connect()` function. Since UDP is a connectionless protocol the connect() step is not needed for client sockets, and there are specific functions for sending and receiving messages (`sendto()` and `recvfrom()`). In those functions the address of the destination socket is declared since in the nominal case no connection has been made before between the client and the server, so the client does not know to whom it has to send the message. 
In the case of AirSim, I guess the connection is mandatory in order to create the `hil_node_` as far as I have understood from the `MavLinkDroneController.cpp` file where the connection to PX4 is made. So if we are running PX4 in SITL mode , we need to connect to it through UDP in order to be able to send/receive the MavLink HIL messages. Nevertheless OSX does not like having a UDP client socket already connected to the server and then trying to send/receive information through it using the UDP-specific functions `sendto()` and `recvfrom()`, it issues error 56 that according to [this page](https://www.ibm.com/support/knowledgecenter/SSLTBW_1.13.0/com.ibm.zos.r13.cs3cod0/syserret.htm) it means that the socket is being connected two times. 
This does not happen in Linux nor Windows, so I guess (and this is only a guess) that Linux and Windows have a check on the `recvfrom()` and `sendto()` functions to check if the socket is already connected, and if it is they do not try to connect it again. OSX does not seem to have this check and so if the socket is already connected, then by using `sendto` and `recvfrom` what we are doing is that we are connecting again a socket which is already connected to the server, thus resulting in a socket error.
I have solved this by modifying the the UDP client port in MavLinkCom to use the classic TCP `write()` and `read()` functions instead of `sendto()` and `recvfrom()` when we are running the code on OSX.

I have successfully tested the MavLinkTest program on OSX, on Windows and Linux nothing should be changed but a double check for sanity reasons would be nice.

Cheers, 

Nacho.

### DIFF
--- a/MavLinkCom/src/serial_com/UdpClientPort.cpp
+++ b/MavLinkCom/src/serial_com/UdpClientPort.cpp
@@ -157,7 +157,11 @@ public:
 		}
 
 		socklen_t addrlen = sizeof(sockaddr_in);
+		#if defined (__APPLE__)
+		int hr = ::write(sock, reinterpret_cast<const char*>(ptr), count);
+		#else
 		int hr = sendto(sock, reinterpret_cast<const char*>(ptr), count, 0, reinterpret_cast<sockaddr*>(&remoteaddr), addrlen);
+		#endif
 		if (hr == SOCKET_ERROR)
 		{
 			hr = WSAGetLastError();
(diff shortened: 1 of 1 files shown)


RVBldr:
Internal VSTF 7402.

Issue #254, opened 2017-06-02
Best way to fly AirSim drone from ground pilot perspective


Timebutt:
I have been playing around with a solution to be able to fly the drone in AirSim from a pilot's perspective. This can be interesting for many other people, as AirSim could excellently serve as a flight trainer as well. I have my official UAV pilot examination coming up shortly so could use some extra flight time ;)

_(yes, we need to perform different manoevres and display general drone flying 'prowess' in Belgium, in order to obtain the pilot's license)_

I present my current solution to the problem to start a discussion, and see if there maybe is a more elegant and user-friendly solution. Maybe a custom AirSim component or specific behaviour could natively provide this functionality instead of hacking around like I do now?

Here goes:
I first made sure there is no `Player Start` in the Unreal environment, and added the `BP_FlyingPawn`. After that, I added a traditional camera (so not the `BP_PIPCamera`)

![camera](https://cloud.githubusercontent.com/assets/10674081/26725884/e6f672de-47a0-11e7-989c-26c08d62e425.PNG)

Next I opened the Level Blueprint:

![open_level_blueprint](https://cloud.githubusercontent.com/assets/10674081/26726005/82a12738-47a1-11e7-80c7-f11e9152eef0.png)

There I added the following modules and connections to make sure that the new camera always center focuses on our drone, and that the newly added camera is used as the default camera. The white line arriving comes from `Event Tick` module, so that this code is continuously executed. A solution where I execute ´Set View Target with Blend` once (`Event BeginPlay`or `Respawn player_Event` for instance) didn't seem to work correctly. Having this code run every tick probably isn't the most efficient solution, but it works.

![blueprint](https://cloud.githubusercontent.com/assets/10674081/26725836/ae7d1930-47a0-11e7-8654-d7f978cff57b.PNG)

Result: flying as if you were flying the drone outside in real life!
(in the image the 'pilot' is higher up than the drone)

![flying](https://cloud.githubusercontent.com/assets/10674081/26726182/6a64d830-47a2-11e7-8212-dd0e64152aa8.png)

Is this the way to go, or can we improve this?
I will gladly write documentation to go with this.

lovettchris:
This looks great.  We have these modes in the ~\AirSim\Unreal\Plugins\AirSim\Source\CameraDirector.cpp.  We also have this already: ACameraDirector::inputEventGroundView which is bound to the backslash key:
`
[...]

PR #261, opened 2017-06-06
Added 'Pilot' view to AirSim

Timebutt:
I have written and tested code to add the 'Pilot' view functionality to AirSim, following thoughts and discussion in issue #254.


sytelus:
Sorry about the delay in reply. 
I'm still trying to understand "pilot view". Traditionally FPV view is the pilot view (i.e. as if you are sitting in the drone). The ground view is supposed to be about 6 feet from ground level. It simulated view of a person who is running after drone on the ground. Currently we don't do ground detection using ray trace (which is now super easy via function in AirBluePrintLib) so z is kind of fixed. But that's easy to fix.

Some comments on PR iteself:
- it would be good to squash the commits so its easy to review changes
- looks like changes in AirBlueprintLib.cpp and .h is mostly white spaces? It's hard to see changes in these files.
- Could you describe changes in binary file .uasset? I'd moved most of the BP code to C++ so there should be no need to change .uasset anymore.
- Probably not a good idea to change kbd assignments.
- Probably not a good idea to define macros like print(text)

Timebutt:
Hi @sytelus, thank you for your reply.

I will try to explain the issues I am having with the current views in AirSim and the reason why I added _Pilot view_ (I get that the name can be confusing but I didn't find anything better so far - a UAV _pilot_ to me, is the stationary person actually flying the drone). The current _Ground View_ has two problems for me:

1. The observer is always 6ft above 'ground level', which poses a problem in _Mountain Landscape_ as my takeoff point generally is a lot higher (see my comment visually explaining this [here]( https://github.com/Microsoft/AirSim/issues/254#issuecomment-306438359)). As the image shows, this far away from the drone it's almost impossible to fly still see it.
Granted: it seems adding ground detection using ray trace may fix this issue, and always place the observer 6ft above the actual terrain.

2. I use AirSim not only for my research, but also as a simulator to fly a drone and learn difficult maneuvers (figure 8, flying a POI around an object, ...). The _Ground View_ is not a realistic way of looking at the drone as the 'drone pilot': no pilot is running around at the same speed of the drone, always positioned perfectly beneath it. In reality, a UAV pilot chooses a comfortable and safe spot to fly and stays there for most of the flight. This is exactly the situation I have recreated here. 
No current view allows me to correctly evaluate the flight path I am doing as they are always fix, relative to the drone. Figure 8 for instance, I can't assess if my two circles were equal in size and if they touched nicely when the camera is always following the drone. An important exercise for the UAV pilot licence here that I am currently training in AirSim.

3. The _Ground View_ does not allow the user to position the observer, relative to the drone. My view does: if you move the camera director around, you basically position the UAV pilot. You could recreate a situation where the pilot is up the hill, a little further away from the drone looking down on things, or up close right next to the drone. _Ground View_ has no such flexibility.

Maybe properly naming the view could make things a lot clearer, but I'm not sure what would be more appropriate. Have I explained to some extent, what I envision and created this view for?

If it's not the AirSim intention to include a 'flight simulator' element, I understand and will withdraw the pull request. I just needed this function for myself, and thought maybe other could benefit from it too.

As for your PR comments, you are absolutely right. I think I crammed in too much stuff last-minute and didn't do a proper job of presenting it in a clean way. This is my first time committing to a project of this size. I will edit my pull request so that you can see clearly what the exact changes are, and add a screenshot of the changed binary file.

sytelus:
Sorry about delay in response. I'm still struggling to understand what you mean by "Pilot View". Do you mean, its same as ground observer mode but camera stays at same place (only orientation changes?).

szx0112:
This is exactly I am looking for. I view from the remote pilot while the drone is flying. I just wondering if this function has been added in the AirSim.

PR #327, opened 2017-07-07
Update AirSim.Build.cs to fix IWYU compiling error

ytgui:
disable IWYU option for UE 4.15.3 or later, which will cause compiling error.
more detail please visit https://forums.unrealengine.com/showthread.php?137015-4-15-C-Transition-Guide

Only modifications are line 59-61

screenshot from 2017-07-01 12-35-02

### DIFF
--- a/Unreal/Plugins/AirSim/Source/AirSim.Build.cs
+++ b/Unreal/Plugins/AirSim/Source/AirSim.Build.cs
@@ -56,6 +56,10 @@ public class AirSim : ModuleRules
 
     public AirSim(TargetInfo Target)
     {
+        // Disable IWYU option for UE 4.15.3 or later, which will cause compiling error.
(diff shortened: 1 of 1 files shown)


sytelus:
Thanks for your contribution. I've added the change in to master. I've also added file AirSim.Build.4.16.cs which you can use with Unreal 4.16 version.

BossKwei:
nice work!!

BossKwei:
@sytelus And, in [https://github.com/Microsoft/AirSim/blob/master/Unreal/Plugins/AirSim/Source/AirSim.Build.cs#L124](https://github.com/Microsoft/AirSim/blob/master/Unreal/Plugins/AirSim/Source/AirSim.Build.cs#L124), do you forget an `isLibrarySupported = true;` for linux?

PR #412, opened 2017-08-11
Corrected trig function in toEulerianAngle

gmgremillion:
Changed trig function from sin to asin in calculating pitch angle from quaternion in 'toEulerianAngle' in PythonClient.py. This is now consistent with [AirSim/AirLib/include/common/VectorMath.hpp](https://github.com/Microsoft/AirSim/blob/74b476dd881ebf819a8327fdeb0c8329ae170bc8/AirLib/include/common/VectorMath.hpp#L236)

### DIFF
--- a/PythonClient/PythonClient.py
+++ b/PythonClient/PythonClient.py
@@ -114,7 +114,7 @@ class AirSimClient:
                 t2 = 1
             if (t2 < -1.0):
                 t2 = -1.0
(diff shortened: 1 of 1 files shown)


sytelus:
Thank you for PR. This is now incorporated in this commit: https://github.com/Microsoft/AirSim/commit/3aa4195367eaab6bf1992d853daad8432de93215

PR #438, opened 2017-08-25
Update line 32 and 33

shuklam20:
The file location in line 32 is incorrect.
In line 33: UE4Editor is an executable so should be opened using "./UE4Editor" command


sytelus:
Thanks. Docs have been changed since this PR. I've added one related change in this commit: https://github.com/Microsoft/AirSim/commit/311a157443d8c17b243ccf5c05d424c196c28cda

PR #484, opened 2017-09-23
Update settings.md

abhay-venkatesh:
Wasted quite a bit of time because this was not mentioned anywhere. Apparently not supposed to delete the initial two entries.


abhay-venkatesh:
And a couple of subsequent typos.

sytelus:
I'm not sure about new example you have added which is only slighly different than cv_mode.py. Your suggestion to update settings doc is now added in this commit: https://github.com/Microsoft/AirSim/commit/e4f38291207b757d75808a4bc73e5b669dfdc146

PR #485, opened 2017-09-25
Add some try/catch robustness to hand bad settings values.

lovettchris:
Make Airsim a bit more robust to bad settings file input.


sytelus:
I'm not too sure about suppressing errors with logging in Settings. The Settings can be used in many context, such as headless and logging could have been redirected or not noticed. This can leave users very frustrated. I've made this change which suppresses error in UX and shows message instead: https://github.com/Microsoft/AirSim/commit/e4f38291207b757d75808a4bc73e5b669dfdc146. Not entirely perfect but closer to your goal.

PR #493, opened 2017-09-28
Fixe build errors; Add supports for GamePad; Allow rolling stop by adding a brake button

MattChanTK:
- Fixed build errors by adding the appropriate includes.
- Added supports for GamePad with Joysticks.
- Changed to use a separate button for brake to allow rolling stop.
- Fixed doc to use "DebugGame Editor" instead of "DebugGame"


Unreal/Plugins/AirSim/Source/Car/CarPawn.cpp
```diff
@@ -339,8 +345,19 @@ void ACarPawn::setupInputBindings()
...
+
+    UAirBlueprintLib::BindAxisToKey(FInputAxisKeyMapping("Footbrake", EKeys::SpaceBar, 1), this, 
+        this, &ACarPawn::FootBrake);
+
+    UAirBlueprintLib::BindAxisToKey(FInputAxisKeyMapping("Footbrake", EKeys::Gamepad_RightTriggerAxis, 1), this,
+        this, &ACarPawn::FootBrake);
+
+    UAirBlueprintLib::BindActionToKey("Reverse", EKeys::Down, this, &ACarPawn::OnReversePressed, true);
+    UAirBlueprintLib::BindActionToKey("Reverse", EKeys::Down, this, &ACarPawn::OnReverseReleased, false);
+    //UAirBlueprintLib::BindActionToKey("Reverse", EKeys::Down, this, &ACarPawn::OnReverseReleased, false);
```
tuzzer: This should be deleted.

sytelus:
I've merged these changes in this commit: https://github.com/Microsoft/AirSim/commit/550e76de7a3d105d1b99b11883c3a91789205d07. I'd to fix few things, for example, how reverse is handled because otherwise things were always stuck in 1st gear.

PR #499, opened 2017-09-30
fix default initialization of an object of const type error for GeoPoint

chickensouple:
Building Block Environment on Ubuntu 16.04 get error

"default initialization of an object of const type 'const
      simple_flight::GeoPoint' without a user-provided default constructor"


sytelus:
Thank you for your contribution. I've merged your change.

PR #530, opened 2017-10-11
Added column headers to logged file

shrimalmadhur:
- Initialized in SimModeBase so that we can extend it to support different parameters to log from different pawn


sytelus:
I've merged this PR: https://github.com/Microsoft/AirSim/commit/42bf6dda70fd9a2ab1320ba7954f6847af9eab79

We should probably do few changes, however...

1. Should we eliminate duplication of column list in derived classes? The SimModeBase provides the default and derived class should override only if its different?

2. appendColumnHeader could be  possibly be less verbose. See here: https://stackoverflow.com/questions/9277906/stdvector-to-string-with-custom-delimiter

shrimalmadhur:
@sytelus Agreed with both. will make the changes and send a PR.

PR #540, opened 2017-10-15
High poly suv

mitchellspryn:
Commit to use high-polycount SUV as car model instead of default go-kart.


sytelus:
Merged in this commit: https://github.com/Microsoft/AirSim/commit/1c398ba3887b33e84c83a335beb1173c85332c68

Issue #620, opened 2017-11-18
Deep Learning example


mitchellspryn:
We need an example of how to use AirSim to train deep learning models. I already have an working model and dataset; the tutorial needs to be written. 

The PR should be submitted by Monday (11/18).

sytelus:
Thanks for notifying. Looking forward to it.

mitchellspryn:
[PR is up](https://github.com/Microsoft/AirSim/pull/624).

mitchellspryn:
I've updated the PR with your suggestions.

PR #624, opened 2017-11-21
End to end deep learning tutorial

mitchellspryn:
In response to [issue 620](https://github.com/Microsoft/AirSim/issues/620), create a tutorial for how to use end-to-end learning to drive the car.


sytelus:
Thanks for the PR. Few changes I would suggest:

1. Reduce the file sizes: Current PR size is 23MB, most of it is because of driving_car.gif. It looks like this gif can be 6 MB. I will send gif separately,
2. Should we put the .h5 model and other large files mentioned in tutorial together? You can send me dropbox link and I can upload in to our releases page. 
3. I think tutorial doesn't seem to have step-by-step instruction on how to reproduce your results. Generally, the outline of the tutorial is (a) abstract (b) pre-req and setup (c) steps to reproduce your results (d) any observations/measurements (e) conclusion and future work. Basically the idea is that users should be able to follow instructions and reproduce your experiments. We are in process to get RL tutorial in the same framework as well. Please also feel free to add list of authors at the start of the tutorial. You can also fill in initial details now and send more updates in later PRs.
4. Reference your gif as `Car Driving` in markdown.
5. Lets rename AI folder to Tutorial folders so we can add other tutorials there such as using AirSim in headless mode etc.

mitchellspryn:
1) I don't quite understand your suggestion. Are you suggesting that I remove the gif entirely? I can do that. 

2) I included the sample model in the case that one did not want to train the model themselves from scratch. I can remove it if it's too large. 

3) I'm confused. All of the pieces that you suggest are included. (A) and (B) are in the README.md, running the notebooks will give us (C) and (D), and the last block in "step 2" answers step (E). 

4) If I'm reading suggestion (1) correctly, I'll be removing the gif.

5) Sounds good.

sytelus:
1. No, I'm not suggesting to remove the gif. We just need to make it smaller in size. I've sent you another version of gif which is much smaller.
2. We should include the sample model with rest of the large data instead of in your tutorial folder. We can zip up both and add in to releases.
3. Yes, we just need to note these steps in your readme. Many users won't know the purpose of all the files available in folder. So just add a section and provide a link to the next file user should be looking at and how they should use that file so they can obtain same results you have obtained.
4. No we don't need to remove gif.

mitchellspryn:
I understand now. I'll have the changes up this evening.

mitchellspryn:
Updated PR with suggestions.

aditya17a:
@sytelus Per our discussion yesterday, changes to this PR have been made, can you please approve and merge?

sytelus:
This tutorial along with example of Azure Batch is now available at aka.ms/ADcookbook

PR #670, opened 2017-12-09
fix px4 takeoff

lovettchris:
Fix behavior of PX4 drone when sitting on the ground in disarmed state.  Without this fix AirSim was generating crazy sensor readings for z-acceleration.  It is now nice and smooth and PX4 is much happier.  This allows PX4 to finishes sensor fusion and declare a home position, which in turns makes the "takeoff" command happy.


sytelus:
Merged as https://github.com/Microsoft/AirSim/commit/261a76fe8957e1cea000139d5da65c8e12d43a1d

PR #676, opened 2017-12-11
Fix compilation errors on CentOS 7.4

edowson:
This commit fixes compilation errors on CentOS 7.4 using gcc-7.2.1 and clang-5.0.1.


sytelus:
I've fixed the issue in VehiclePawnWrapper so those changes are no longer required. Rest of the change is now integrated in https://github.com/Microsoft/AirSim/commit/23e11d9e6568ac4192f2a849dba5cb3dc1efa50d.

PR #696, opened 2017-12-19
Python Car Client AttributeError Bugfix

kenneyd:
I had two issues with the python client for the car control, I got `AttributeError`'s for the `getCarState` function in the CarClient class in `AirSimClient.py`, and I had the same temp folder issue that I ran into in #695 

I fixed the attribute error by adding the `collision` and `kinematics_true`, and setting them to the same function used for the `MultirotorState` class. Used the same tempfolder approach here from the drone control PR mentioned above.

### DIFF
--- a/PythonClient/AirSimClient.py
+++ b/PythonClient/AirSimClient.py
@@ -161,7 +161,6 @@ class CarState(MsgpackMixin):
     collision = CollisionInfo()
     kinematics_true = KinematicsState()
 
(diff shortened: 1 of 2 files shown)


sytelus:
This is now fixed in https://github.com/Microsoft/AirSim/commit/b1ecc13f500a05d355dbe7aca65c200a3f4e0051

PR #713, opened 2017-12-28
Fix missing collision attribute of CarState

NeilMaude:
Added collision attribute, to allow hello_car to run - simple, but demo wouldn't run without it...


sytelus:
This is fixed here now: https://github.com/Microsoft/AirSim/commit/b1ecc13f500a05d355dbe7aca65c200a3f4e0051

PR #722, opened 2018-01-07
tweak build.cmd

ssashir06:
hi, I have modified build.cmd file to improve usability while using VS2015 Native tools command.

1) ROOT_DIR is set as the path of the batch file, instead of %CD%. If current directory is wrong and this modification avoid making trouble, like creating files some wrong directory.
2) "exit" command of the batch file is now using /b option. If error was occured and this modification avoid force to close command prompt window.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -1,7 +1,7 @@
 @echo off
 REM //---------- set up variable ----------
 setlocal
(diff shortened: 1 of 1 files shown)


ssashir06:
Thanks!

PR #732, opened 2018-01-14
build on macOS High Sierra with Unreal Engine 4.16

ssashir06:
Changes for compiling with macOS High Sierra.

1) Adding ignored settings
2) undef nil to avoid compile error; nil is already defined by MacTypes.h
3) Add a GenerateProjectFiles.sh in Blocks path to create Xcode project
4) Update setup.sh to complete brew command correctly
4) Update setup.sh to avoid using unnecessary sudo command; without sudo works fine and it is confirmed in macOS environment

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -324,3 +324,25 @@ car_assets.zip
 /PythonClient/cloud.txt
 __pycache__/
 /PythonClient/metrics/
+
+### Xcode ###
+build/
+DerivedData/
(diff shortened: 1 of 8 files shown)


sytelus:
Thanks for this contribution. This had been asked by many folks in the community!

ssashir06:
I'm grad to hear that! Thank you.

PR #761, opened 2018-01-26
Bugs in the code in build.cmd

jayjagtap:
Even when cmake is installed properly and the path is rightly set, it shows the error cmake: directory not found.
Deleting/Commenting the part of build.cmd where it calls check_cmake.bat solves the problem.
Moreover, it should be explicitly specified that Visual Studio command prompt must be used.

Check my submission in the link below.


sytelus:
I'm unable to reproduce this and also I haven't heard this from any other users. The commenting out check_cmake is not a right approach in any case. We need to find out why check_cmake fails to do its job.

sytelus:
Ok... thanks. I look forward to your findings.

sytelus:
Closing this due to inactivity.

PR #785, opened 2018-02-02
Add nonlinear adaptive sliding controller and replace PID controller integration method

andrealaffly:
Added AdaptiveController.h and AdaptiveControllerParams.h to the firmware folder in simple_flight. These files implement an adaptive sliding control, which is more robust than the linear control system currently being used. A simple boolean variable is also added in firmware.h to switch between the adaptive and the linear controls. A code is added to feed state variables to the adaptive controller and read the results.

Additionally, the integration scheme in the PID controller was replaced with Runge-Kutta 4 for higher accuracy.

This code was written by Mr. Coleton Domann in Fall 2017 at the Advanced Control Systems Lab (ACSL) of the University of Oklahoma. For further information, visit http://lafflitto.com or contact us at a.lafflitto@ou.edu


sytelus:
Thanks for the submission. There were few problems with code which I've fixed now in this commit: https://github.com/Microsoft/AirSim/commit/a6aee1c7011aa60c5b51e18744a3e6dd9d421cc5

The changes that were required:
1. Code wasn't compilable as it generated lots of warnings (we have warnings as errors enabled).
2. There were many style and formatting issues.
3. The adaptive controller didn't implemented standard interface.
4. PID controller should use switch to use either usual or RK integrator.
5. Firmware should also use switch to decide between CascadedController vs AdapticeController.
6. Adaptive controller doesn't take in to account goal at all (i.e. desired state). So I'm concerned if this code actually works.

Please review above commit which has fixes for these issues. It would be great if this new code is properly tested because I am doubt that it actually works (see # 6)...

Issue #405, opened 2017-08-08
How to make ultra sonic sensor


winszenYoung:
Hi , I have a few question about sensors.

I used gazebo, it was suportted many sensor(ultra sonic, 3d lidar, etc, optical flow)

I recently use airsim. I find sensor document about cam( depth, rgb) 

but, it is not have a another sensor, 

I want to use ultrasonic senosr or etc.

How do i use this sensor?

If it is not suportted, How do i make a ultrasonic senosr or etc?

sytelus:
Currently its not available out of the box so you will need to roll your own. Its not too hard to do. there is a SensorBase class which you can use as your base class and then use ray tracing to shoot few rays in predefined cone to get distances and then add some noise. There is a function `UAirBlueprintLib::GetObstacle` that you can use for ray tracing.

realdealneil:
I am trying to do this.  It's easy to set up the base class and make a class for a laser altimeter (which is all I want for the time being).  However, I don't yet know enough to figure out how to access the pawn's attitude and shoot a ray out to find where the laser would hit the ground.  

How do I access the pawn from the sensor class?  Here's what I'm trying right now (below is the getOutputInternal function from my LaserAltimeterSimple.hpp implementation):

Output getOutputInternal()
    {
        Output output;
        const GroundTruth& ground_truth = getGroundTruth();

        auto altitude = ground_truth.environment->getState().geo_point.altitude;
	auto position = ground_truth.environment->getState().position;
	auto attitude = ground_truth.kinematics->pose;

	output.laser_alt_m = altitude;      // TODO: fill in with true height above ground
	output.reading_raw_m = output.laser_alt_m;   // TODO: fill in with ray-traced value from vehicle pawn laser rangefinder

        return output;
    }


***********************************
I can access the pose of the aircraft (from which I can extract the attitude quaternion, cool).  However, UAirBlueprintLib::GetObstacle needs more stuff.  (Pointer to the right actor, start and end point of the trace, and then I need to handle the FHitResult).  

I'm thinking that the ground_truth structure should probably contain "terrain_height" and then the laser altimeter would just use this terrain_height and add some noise, etc.  If this is the right way to implement this, can someone who knows Unreal/Airsim better than me offer some advice on where to locate the function that does the ray tracing and give me some hints on how to call ::GetObstacle and interpret the result?

RVBldr:
Internal VSTF 7412

msb336:
[...]

PR #791, opened 2018-02-05
Adding laser altimeter to PX4 sensors.  Originally written by Ben Sen…

realdealneil:
This pull request adds support for a laser altimeter sensor on the PX4 autopilot.  It uses ray tracing to find the position of the nearest obstacle below the multirotor vehicle.  

This is connected with the feature request referenced here:
https://github.com/Microsoft/AirSim/issues/405

Sensor implemented by Ben Sena
Branch-rebase-squash-merge done by Neil Johnson.


sytelus:
I've merged your changes in [this commit](https://github.com/Microsoft/AirSim/commit/c6fd9de5d75dd0d57383d2b198f35d03212c26ab). Thanks for great contribution!

While merging this, I realized, we have quite a bit of room to enhance the sensor architecture so sensors like this is easily pluggable. So I went ahead and did bunch of refactoring. I also refactored distance sensor so now it would be possible to add multiple of these sensors on same vehicle at arbitrary relative pose. The major change is that the distance sensor was mounted at 25-degree pitch in your example as default and in new change its mounted as front facing. I hope this is minor thing on your side and easily changeable.

You can see my changes in this commit:
https://github.com/Microsoft/AirSim/commit/2a77fd94a58f662600a35c012ad3767faf81e5a5

realdealneil:
Yes, this should be easy to refactor according to our needs.  The intention was to use this as a downward facing laser altimeter, but it could also be used as a distance sensor facing forward as you've coded it.  Eventually we should make the mounting angle configurable and allow for multiple distance sensors to be attached.  It would also be good to add sensor models that at least attempt to model the difference in sonar versus laser (versus other types of) distance sensors.  Maybe we'll open another feature request and work on that down the road.

Issue #776, opened 2018-01-31
Segmentation API is limited


aburgm:
The `simSetSegmentationObjectID` API as described [here](https://github.com/Microsoft/AirSim/blob/master/docs/image_apis.md#segmentation) is somewhat limited, in particular:

 * It is not possible to un-assign a segmentation ID so that it does not show up in the segmentation view (custom depth rendering disabled). In the neighbourhood environment, there are some leaves lying on the roads or the grass which I would like to just ignore for the segmentation so that they take on the segmentation ID of the underlying road or grass, respectively.

 * There are objects that do not have a reasonable (or easily to look-up) object ID, such as spline mesh components or mesh components instantiated by a foliage actor. Especially spline mesh components have the same name as the landscape, so there is no way to distinguish them from the landscape. This applies to the road in the mountains landscape environment, or the power lines, fences or the path in the park in the neighbourhood environment.

aburgm:
For the first point, an easy extension could be to turn off the custom depth rendering flag when the object ID passed to the API is negative.

For the second point, I often find myself in the situation where I want all instances of a particular mesh, such as "Path_01" or "Tree_01", to have the same object ID. I have played around a bit with changing the [UAirBlueprintLib::GetMeshName](https://github.com/Microsoft/AirSim/blob/master/Unreal/Plugins/AirSim/Source/AirBlueprintLib.cpp#L182) function to the following:

```c++
template<class T>
std::string UAirBlueprintLib::GetMeshName(T* mesh)
{
    UStaticMeshComponent* staticMesh = dynamic_cast<UStaticMeshComponent*>(mesh);
    if (mesh->GetOwner() && staticMesh && staticMesh->GetStaticMesh())
        return std::string(TCHAR_TO_UTF8(*(mesh->GetOwner()->GetName()))) + "/" + TCHAR_TO_UTF8(*staticMesh->GetStaticMesh()->GetName());
    else if (mesh->GetOwner())
        return std::string(TCHAR_TO_UTF8(*(mesh->GetOwner()->GetName()))) + "/null";
    else
        return ""; // std::string(TCHAR_TO_UTF8(*(UKismetSystemLibrary::GetDisplayName(mesh))));
}
```

This plays together nicely with the regex functionality of the `simSetSegmentationObjectID` API because it easily allows to match either the owner name or the static mesh name or both. Obviously this would break existing users of this API, so this is meant more as a starting point for discussion. Thoughts welcome!

sytelus:
[...]

PR #814, opened 2018-02-14
Allow to unset segmentation object ID

aburgm:
This allows to pass -1 with the simSetSegmentationObjectID API to turn off segmentation for the matched objects. See #776.


sytelus:
Good contribution! Thanks.

PR #831, opened 2018-02-19
Add projection mode for capture settings

aburgm:
This allows to switch to orthographic projection mode. It can be used to create a "map" of an environment by taking overhead orthographic pictures in segmentation view.


sytelus:
Good changes! Merged and added minor change in above commit.

PR #842, opened 2018-02-23
rosbridge first test

n1ckfg:


sytelus:
Check out this PR: https://github.com/Microsoft/AirSim/pull/1135

Also, we just completed massive upgrade that simplifies a lot of things. The new API layer should be much easier to extend. Please see: https://github.com/Microsoft/AirSim/blob/master/docs/upgrade_apis.md

PR #850, opened 2018-02-27
Add setting to disable rendering of main viewport

aburgm:
I spent some time optimizing the frame rate at which I am able to obtain images through the API, by just requesting images in a loop in Python. Unsurprisingly, it turned out that when getting uncompressed images the frame rate is higher than getting compressed images. Also unsurprisingly, it turned out that the frame rate is higher when not rendering the actual main viewport that is displayed on the screen, which can be disabled by running the command `show rendering` in the Unreal console.

So, this PR adds a setting that allows to disable the rendering of the main viewport in the same way, for cases where it is not needed and instead a high frame rate of images to be obtained through the API is desirable. In that case it also makes sure that the graphics pipeline is flushed after every frame has been drawn, because otherwise the API image request only gets scheduled to execute only after every other frame has actually been rendered. This effectively doubles the frame rate.

To test these tweaks I set up a particular scene in the neighbourhood environment at which I was getting about 18 FPS when rendered onscreen. When running my Python loop to get images through the API, images were coming in at about 3 FPS, then with main viewport rendering disabled around 7 FPS and with the pipeline flush tweak about 12 FPS. Given that the on-screen framerate I observed was 18 FPS there might be more improvements possible, but this seems like a good start and fairly low-hanging fruit.

I also realized that the loop that copies the image into the results after it was captured took about 10ms on my machine for a 512x512 image. I have optimized this so that it takes less than 1ms, but did this did not improve the total frame rate, probably because the GPU was already busy rendering the next frame at this point. I still thought it's a worthwhile contribution, though.

Would it make more sense to use a new ViewMode enumerat instead of a boolean "MainViewportVisible" setting for this? Although the view mode already has a "NoDisplay" setting that seems to do something different, so it might be confusing.

### DIFF
--- a/AirLib/include/common/AirSimSettings.hpp
+++ b/AirLib/include/common/AirSimSettings.hpp
@@ -167,7 +167,6 @@ public: //fields
     float clock_speed;
     bool engine_sound;
     bool log_messages_visible;
-    bool main_viewport_visible;
     HomeGeoPoint origin_geopoint;
 
 public: //methods
@@ -227,7 +226,6 @@ public: //methods
         clock_speed = 1.0f;
         engine_sound = true;     
         log_messages_visible = true;
-        main_viewport_visible = true;
         //0,0,0 in Unreal is mapped to this GPS coordinates
         origin_geopoint = HomeGeoPoint(GeoPoint(47.641468, -122.140165, 122)); 
     }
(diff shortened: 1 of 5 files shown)


sytelus:
The ViewMode = NoDisplay is exactly for the same purpose. I was playing around with various ways to disable rendering but without affecting SceneCapture2D renders. I think you have solved this issue. So it would be great to use ViewMode = NoDisplay for this purpose. Please feel free to remove/comment out old code for ViewMode = NoDisplay and replace with your new approach.

This is great PR! Thanks.

aburgm:
Ah, that makes sense, thanks! I moved the code over to `ACameraDirector`. It's not particularly elegant maybe because it modifies viewport settings instead of camera settings, but it seemed more beneficial to me to keep all the view mode logic in the same place.

sytelus:
Excellent! I've merged your changes and also did minor refactoring to clean up a bit.

PR #872, opened 2018-03-09
Development

SweetShot:
Fixed a bug where client calls will not work without any cameras


SweetShot:
Wrong place apologies

PR #890, opened 2018-03-19
Added moveByRotorSpeed command for QuadX

zbenic:
Enables to move the quadcopter in X configuration by setting each rotor speeds in rad/sec with the command moveByRotorSpeed.

### DIFF
--- a/AirLib/include/common/HelperFunctions.hpp
+++ b/AirLib/include/common/HelperFunctions.hpp
@@ -1,0 +1,56 @@
+#pragma once
+
+#include "vehicles/multirotor/RotorParams.hpp"
+
+namespace msr { namespace airlib { namespace helper_func {
+
+class Converter {
+public:
+    void getCommandArgumentsFromSpeed(const simple_flight::Axis4r& input_command_args, simple_flight::Axis4r& output_command_args)
+    {
+        simple_flight::Axis4r motor_outputs;
+        rotor_params_.calculateMaxThrust();
+        motor_outputs[0] = input_command_args.roll() / rotor_params_.max_speed;
+        motor_outputs[1] = input_command_args.pitch() / rotor_params_.max_speed;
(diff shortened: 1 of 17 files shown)


sytelus:
Hi @zbenic,

I was thinking about bit better way to implement this. One way I thought about was to rename existing mveByAngle to moveByAngleZ and then add another method moveByAngleThrottle. That's what I exactly did in this commit: https://github.com/Microsoft/AirSim/commit/b4a934b9dd49f7e0edddd364a3080bc5e243569f

Notice that moveByAngleThrottle ultimately is [implemented in simple_flight controller](https://github.com/Microsoft/AirSim/blob/master/AirLib/include/vehicles/multirotor/firmwares/simple_flight/SimpleFlightDroneController.hpp#L233).

Also, instead of yaw, I'm using yaw_rate because that's what RC does. However if you want yaw instead it's very simple change.

I haven't tested this thoroughly but hopefully this provides good starting point for your purpose.

zbenic:
Thanks for the info, but in my case your proposed solution is not suitable.
I really need to use the Passthrough controller for each control variable, as I do not wish to use any kind of controller (in this case AngleRate's PID).

zbenic:
Yeah, I don't need the same control as RC as the plan is to simulate autonomous multirotor.
I guess we could do that. I have named my command moveByRotorSpeed because at one point I convert command argument values (-o0 ... -o3) from rotor speeds to their corresponding roll, pitch, yaw, throttle values. Looks to me that, in this case, name moveByPassthrough could be a bit misleading.

DroneMesh (approved):

madratman:
Incorporating @sytelus suggestion in and closing in favor of https://github.com/microsoft/AirSim/pull/2297 
Planning to add `moveByMotorPWM` and `moveByMotorSpeed` in simpleflightpai

Issue #893, opened 2018-03-20
libc++ / libstdc++ compatibility issues


mescarra:
Hello everyone,

I'm trying to compile a simple example, like the HelloDrone, but also adding any other dynamic library you could find in linux (built with gcc and libstdc++). I just cannot get it done.

I'm using Ubuntu 16.04 on an ARM64 architecture. `g++ 5.4.1`, `clang++ 3.9.1`. I have installed AirSim's static libraries in `/usr/local/lib`.

So far, I've tried to use different combinations of g++, clang++-3.9, stdlib=libstdc++/libc++ and selective static/dynamic linking (-Wl,-Bstatic).

If I compile this way:
`g++ -I/opt/ros/kinetic/include  -I/usr/include/eigen3 -I/usr/local/include/airsim -I/usr/local/include/rpc -DRPCLIB_MSGPACK=clmdep_msgpack -isystem /usr/local/include/llvm -std=c++11 -DMSGPACK_PP_VARIADICS_MSVC=0 -fPIE -o Drone main.cpp -L/opt/ros/kinetic/lib -lrosbag_storage -lc++ -lc++abi -lpthread -lAirLib -lrpc -lMavLinkCom` (`lc++` and `lc++abi` are llvm compiled libraries.)

I get:

`undefined reference to 'rosbag::Bag::open(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, unsigned int)'`

which is reasonable because
`nm -C -D /opt/ros/kinetic/lib/librosbag_storage.so | grep rosbag::Bag::open`

outputs

`/opt/ros/kinetic/lib/librosbag_storage.so:0000000000023080 T rosbag::Bag::open(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned int)`

On the other hand, if I remove libc++ link (so defaults to libstdc++) I get undefined errors in rpc library like so:
`../build_debug/output/lib/librpc.a(format.cc.o): In function 'void clmdep_fmt::(anonymous namespace)::check_sign<wchar_t>(wchar_t const*&, clmdep_fmt::internal::Arg const&)':
.../rpclib-2.2.1/dependencies/src/format.cc:306: undefined reference to 'std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::~basic_string()'`

Any ideas?

Thanks in advance.

mescarra:
Well, I've made some progress.

[...]

PR #908, opened 2018-03-26
libc++ / libstdc++ compatibility issues

mescarra:
Fix related to this issue:
https://github.com/Microsoft/AirSim/issues/893

### DIFF
--- a/MavLinkCom/MavLinkTest/Commands.cpp
+++ b/MavLinkCom/MavLinkTest/Commands.cpp
@@ -1607,11 +1607,11 @@ void GotoCommand::HandleMessage(const MavLinkMessage& message)
             channel->moveToLocalPosition(tx, ty, tz, is_yaw, static_cast<float>(theading * M_PI / 180));
             
             if (this->hasLocalPosition) {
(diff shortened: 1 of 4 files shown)


sytelus:
Are you really sure removing std:: is the right approach. This should be supported by any compliant C++11 compiler. Also, I'm not sure about removing c++abi. Removing ABI can generate hidden hard to catch issues. Are there any other alternatives?

aburgm:
In C++ you can just use `std::abs` instead of fabsf if you include `<cmath>`. `std::abs` is overloaded for that purpose. See http://en.cppreference.com/w/cpp/numeric/math/fabs

mescarra:
Yes, using the overloaded `std::abs` is the right approach, I should modify that.

About the llvm's `libc++abi`, I believe clang's compiled programs are still being compiled against `c++abi` due to the `CommonSetup` inclusion.

On the other hand, I'm quite sure `libstdc++` does not separate the ABI in a different library, so it wouldn't be necessary when compiling with gcc defaults.

sytelus:
Sounds good. Thanks for looking into this.

sytelus:
I made above changes and have merged this.

PR #914, opened 2018-03-29
Added ability to interact with goal_

andrealaffly:
Adaptive sliding mode control law for quadrotors. This control law guarantees satisfactory trajectory tracking also in case one of the propellers experiences 50% efficiency loss.


sytelus:
Unfortunately this still doesn't work. Do you have Taranis Plus RC? This RC you can connect to PC via USB and manually fly the drone in AirSim. When I switch from CascadeController to AdaptiveController, drone doesn't fly and has strange behaviour. 

If you are sending new updates, could you please make sure that there no unnecessary forming changes (always use diff tools to make sure what changes you are doing)? Also, I got few compiler warnings for double -> float conversion. I fixed those. It would be great if you can build AirSim and test it successfully before we work on pull request.

PR #979, opened 2018-04-17
Thrustmaster support

kissslorinc:
Added support to Thrustmaster Steering Wheels including Force Feedback.
Working controls are steering, throttle and brake. 
Force Feedback effects include wheel autocenter and wheel rumble.
Tested on Thrustmaster T500 RS.


sytelus:
This looks great! I would merge this shortly after testing with our other devices.

sytelus:
Merged!

BrianN92:
Hi I bought [the thrustmaster steering wheel](https://www.walmart.com/ip/Thrustmaster-Ferrari-Red-Legend-Race-Wheel-PS3-PC/30824036?wmlspartner=wlpa&selectedSellerId=0&adid=22222222227021749969&wl0=&wl1=g&wl2=c&wl3=40342616792&wl4=pla-78302436992&wl5=9033315&wl6=&wl7=&wl8=&wl9=pla&wl10=8175035&wl11=online&wl12=30824036&wl13=&veh=sem) but it did not work. Every time I connect the steering wheel with Airsim, the vehicle keep turning left with the "steering value: -1.25". The  "Accel" value was 0.5 all the time. I cannot use the wheel to control car's direction, but I can do it by pushing the braking panels. Besides, there was no values of "steering', "throttle" or "break" inside the "rec.txt" file. It was so weird, so could you tell me how did you use this steering wheels to control the car? Thank you!

I did put the x360ce exe in the same directory as the AirSim binary.

kissslorinc:
Hi,
If you pulled the last codebase with this merged in, you won’t need x360ce, but you might need to configure the wheel in the config.json file by adding its ID in the „RemoteControlID” tag in the correct section.
When you start the game, you’ll immediately feel if it’s working, because the wheel rumble effect will kick in..
Also, i only had the chance to test it with the T500RS wheel,  i’m not sure it has the same layout as yours (but it probably does).
Let me know how it goes!

BrianN92:
"PhysXCar" : {
"RC": {
"RemoteControlID": 0
}
}
Like shown above, I have added its ID to the "settings.json" file, but that does not work. The car still behave abnormally. For some of the maps, like "neighborhood" and "downtown", the steering wheel works after I added the x360ce to the Binaries folders without changing the json file.

kissslorinc:
The problem might be that the precompiled binaries don't contain this code yet. If you compile the Blocks environment for yourself, it'll probably work. 
Until the devs make a new release, you might not be able to use your wheel with the other environments without x360ce.

PR #981, opened 2018-04-18
Make cmake the primary build system for all platforms

lovettchris:
This pull request promotes the "cmake" subfolder up to the main directory structure, deletes the hand built visual studio projects and switches to using "cmake" on windows (in build.cmd).  The Windows "build" output and VS project files are now generated in a "build" folder.  The Linux build.sh still puts the build output and cmake generated make files in a build_debug folder.


madratman:
I like the general direction of where this PR was going. For other projects, having cmake as the only build systems with VSCode as the iDE works great for dev experience in Windows and Linux both. Also would make maintenance less of a headache. @chrislovett do you have time to bump this up to current master? I can help.

madratman:
Closing due to lack of progress, and this will need to be done from scratch given how much it's lagging from current master. 
In addition clang7 + ue 422 / clang8 + ue 424 should be used to obviate the manual building of libcxx in setup.sh

PR #988, opened 2018-04-18
Modify the start PX4 command based on px4_build.md

yanno9:
The command to start start PX4 was not updated like the one in px4_build.md file.


lovettchris:
Ah, the real problem is the docs need to be updated to say "use the latest version", we no longer need the older version 1.5.5.  When you sync to the latest version the path in the docs is correct.  I will update the docs shortly.

PR #1082, opened 2018-05-09
Removed python 3 function annotation

TranBrian10:
One line of AirSimClient.py contains a Python 3 function annotation, which doesn't seem to change any functionality. I've removed the annotation to restore compatibility with Python 2.

Created issue #1083.

### DIFF
--- a/PythonClient/AirSimClient.py
+++ b/PythonClient/AirSimClient.py
@@ -528,7 +528,7 @@ class MultirotorClient(AirSimClientBase, object):
 
         
     # query vehicle state
-    def getMultirotorState(self) -> MultirotorState:
+    def getMultirotorState(self):
         return MultirotorState.from_msgpack(self.client.call('getMultirotorState'))
     def getPosition(self):
         return Vector3r.from_msgpack(self.client.call('getPosition'))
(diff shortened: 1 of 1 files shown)


sytelus:
Thanks for the PR. However internally we have been thinking about moving on from Python 2.7 as most of our users are using Python 3.5+. We have been actually planning to put lot more annotations so it will be helpful in IDE, linting and generating docs. Is there any scenario where Python 2.x is still critical and folks cannot upgrade?

kshitijgoel007:
@sytelus In my case, Python 2.x is most relevant when a ROS user wants to leverage AirSim for perception data for an Aerial Robot. ROS does not properly support Python 3 yet. I had to remove this line as well.

TranBrian10:
We're in a similar position at the WISE Lab in Waterloo. We use the AirSim Python Client with ROS to develop simulations for self-driving cars. As mentioned above, ROS requires Python 2.x, so we can't consider upgrading to Python 3. We may be able to switch to C++, but keeping Python 2 compatibility would be very useful to us.

sytelus:
Thanks for your feedback. I just looked in to way to make type annotations compatible with Python 2.7 and it looks like this simple change will do:

https://stackoverflow.com/a/42025515/207661

If you can update the PR, I'll merge this change.

TranBrian10:
Looks like it could get a little messy, especially if each function will be annotated. If they're only going to be used by tools like linting and docs, maybe we could group the annotations together instead of putting them under each function.

PR #1090, opened 2018-05-14
Adding custom port option to MultirotorClient base

kshitijgoel007:
Look at https://github.com/Microsoft/AirSim/commit/efe00f30d30cdf9695489e41bd6564eb9efcfe16 
Needed it to make python client to talk to multiple spawns at different ports. Very small change, but I think it might be useful to others.


sytelus:
Agreed with @jacobdeery - why is timeout being removed?

kshitijgoel007:
@jacobdeery @sytelus Sorry about that, restored the timeout.

sytelus:
Thanks. I've added these changes in following commit: https://github.com/Microsoft/AirSim/pull/1090

Please note that our APIs have had an upgrade nd there are some breaking changes. See https://github.com/Microsoft/AirSim/blob/master/docs/upgrade_apis.md.

Issue #1038, opened 2018-04-29
Reducing header file includes for Unreal Engine classes 


SvetoslavAngelov:
Hi Guys, 

I would like to contribute to the project, but I'm still making my way through the code base. While reading the code I noticed that there are few opportunities to reduce header includes for Unreal classes. For example in CarPawn.h:
```
#pragma once

// Use these:
#include "CoreMinimal.h"
#include "WheeledVehicle.h"
#include "SimJoyStick/SimJoyStick.h"
#include "CarPawnApi.h"
#include "CarPawn.generated.h"

// Instead of:
#include "CoreMinimal.h"
#include "WheeledVehicle.h"
#include "physics/Kinematics.hpp"
#include "CarPawnApi.h"
#include "SimJoyStick/SimJoyStick.h"
#include "UObject/ConstructorHelpers.h"
#include "Components/SkeletalMeshComponent.h"
#include "PhysicalMaterials/PhysicalMaterial.h"
#include "common/AirSimSettings.hpp"
#include "AirBlueprintLib.h"
#include "CarPawn.generated.h"

class UPhysicalMaterial;
class UCameraComponent;
class USpringArmComponent;
class UTextRenderComponent;
class UInputComponent;
class UAudioComponent;
```

The same can be done for the .cpp files, since there are a few includes, which aren't necessary. Would you be interested in a contribution of this type?

sytelus:
Yes, this would be a great contribution!

SvetoslavAngelov:
That's excellent! I'll go through a few more files and submit a pull request.

sytelus:
Thanks for this contribution!

PR #1117, opened 2018-05-28
Reduced header file includes in CarPawn

SvetoslavAngelov:
Reduced CarPawn.h header file includes, by moving the majority of includes to CarPawn.cpp. Made class forward declarations consistent, using format "class UClass* MyClass". Also removed redundant header includes in CarPawn.cpp.

This pull request is part of my initial #1038 Issue. I have identified other files which can benefit from header reduction, so I will continue work on this branch if pull request is accepted.


sytelus:
@SvetoslavAngelov - thanks for the contribution! BTW, heads up - we are doing some large scale refactoring in code base so I would suggest to hold of further changes until next week. That will help to minimize the conflicts.

PR #1135, opened 2018-06-05
ROS: add sample airsim ROS nodes for PythonClient and ROS integration…

TritonSailor:
This PR adds several example ROS nodes for publishing AirSim data using PythonClient, along with documentation describing setup and usage.

### DIFF
--- a/PythonClient/ROS/car_image_raw.py
+++ b/PythonClient/ROS/car_image_raw.py
@@ -1,0 +1,53 @@
+#!/usr/bin/env python
+
+# Example ROS node for publishing AirSim images.
+
+import rospy
+
+# ROS Image message
+from sensor_msgs.msg import Image
+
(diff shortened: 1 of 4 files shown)


weihangChen:
I passed by the ros-carla integration, they are using rosbridge instead of rosnode,and more sensor channels are implemented, 

https://github.com/carla-simulator/carla/tree/master/carla_ros_bridge

sytelus:
This is great contribution! Meanwhile we have re-worked API layer which meant there were few breaking changes. But no worries! I just updated your code to use APIs. Also, lowercased ROS folder to ros just to be consistent and updated doc to make sure folks copies airsim package folder. Please see my commit: https://github.com/Microsoft/AirSim/commit/c4bba469195d5a9d5b8541e227d175036eedd47f

PR #1139, opened 2018-06-07
Test with MultiCopter Hexa

askariz:
1. Adding DJIS900 model & blueprint
2. FlyingPawn max rotating_movements_ to 8


askariz:
and also some instructions modify [here](https://github.com/askariz/AirSim/wiki/hexacopter)

sytelus:
Hi @askariz , are DJI related assets in public domain? If we want to include them, they needs to be MIT licensed. Can you please confirm?

PR #1148, opened 2018-06-09
Updated docs with section on upgrading to VS2017

thias15:
Added one short section to FAQ that might help people encountering similar issues as I did.


sytelus:
This is now merged in commit https://github.com/Microsoft/AirSim/commit/ee41cc7c228ce916523527f135fe2efafb535f16

There have been lots of changes in API so upgrade was needed. Also CV mode is now activated using SimMode: ComputerVision instead of UsageScenario:ComputerVision.

Issue #1113, opened 2018-05-25
Enabling simulation mode (drone) after simSetPose


niknil702:
After calling simSetPose it seems like the simulation is set into CV mode, probably as expected. However, how to activate the drone mode again so that for example moveByVelocity and rotateByYawRate can be used again?

I want this for implementing random start pose (position/orientation) each episode.

ulrikborjesson:
I'm experiencing the same issue.

showaykerker:
Same here

PR #1158, opened 2018-06-12
Reset pending_pose_status after setting pose for Mulitrotor

niknil702:
After calling set pose the pending_pose_status is never set to NonePending so vehicle_.setPose keeps being called. I am not sure if this is the way to correct it, but it seems to solve this #1113 at least.

### DIFF
--- a/Unreal/Plugins/AirSim/Source/Multirotor/MultiRotorConnector.cpp
+++ b/Unreal/Plugins/AirSim/Source/Multirotor/MultiRotorConnector.cpp
@@ -157,8 +157,10 @@ void MultiRotorConnector::updateRenderedState(float dt)
         vehicle_.setPose(pose);
     }
 
-    if (pending_pose_status_ == PendingPoseStatus::RenderStatePending)
+    if (pending_pose_status_ == PendingPoseStatus::RenderStatePending) {
         vehicle_.setPose(pending_pose_);
+        pending_pose_status_ = PendingPoseStatus::NonePending;
+    }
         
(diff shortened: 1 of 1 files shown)


sytelus:
Hi @niknil702, This change doesn't look correct. Basically there are two methods that gets called in succession: `updateRenderedState` and then `updateRendering`. In first method we change the physics vehicle pose while in second method we change UE object pose. If you do above then second method would not update rendering.

Also, there huge amount of refactoring I did in last few weeks so lots of things have changed. Please take  a look: https://github.com/Microsoft/AirSim/blob/32a7a44cc9e84350a2299e2c3cb8579be765d97c/Unreal/Environments/Blocks/Plugins/AirSim/Source/Vehicles/Multirotor/MultirotorPawnSimApi.cpp#L112

niknil702:
@sytelus would it be correct to instead do
```
if (pending_pose_status_ == PendingPoseStatus::RenderStatePending) {
   vehicle_.setPose(pending_pose_);
   pending_pose_status_ = PendingPoseStatus::RenderPending;
}
```

As it is now, it seems after `pending_pose_status_ = PendingPoseStatus::RenderStatePending` (when setPose is called) it is never set to anything else and in turn` (pending_pose_status_ ==  PendingPoseStatus::RenderPending)` remains false in `updateRendering` forever.

sytelus:
@niknil702 You are correct. I've made the change: https://github.com/Microsoft/AirSim/commit/5f7893099be5ceae8fd76d1d74f88b0dab5690b2

PR #1161, opened 2018-06-13
Implement simSetObjectPose API

aburgm:
This is the counterpart to simGetObjectPose. It is useful when making recordings of an object in various poses, for instance.


sytelus:
Hi @aburgm, we just did release 1.2 and it has lots of changes in API layer. I've added new simSetObjectPose API in any case. See commit: https://github.com/Microsoft/AirSim/pull/1161.

Issue #1051, opened 2018-05-03
Can the speedometer be changed to mph instead of m/s?


NextSim:

sytelus:
Sure, it can be. The code is in CarPawn.cpp. There could be config in settings.json, perhaps.

PR #1173, opened 2018-06-19
#1051 - Add setting to display speed in mph

NextSim:
Add setting to to mph as speed instead of m/s.  If mph is selected, do the conversion and update the displayed units.


NextSim:
Working on fixing the conflicts.. first time doing a pull request and using git, so it's taking a bit to get familiar with it.

PR #1181, opened 2018-06-20
#1051 Add setting to display speed in mph

NextSim:


sytelus:
This seems like good change. However, can we make it slightly more cleaner and generic? I would suggest we add two settings:

1. speed_unit_factor (defaults to 1)
2. speed_unit_label (defaults to "m/s")

Then folks can change above to km/hr or miles/hr or miles/day whatever else they need for their specific vehicle type.

NextSim:
Sounds good

NextSim:
I think this is ready to be merged in.  Is there anything else I need to do?

Issue #1190, opened 2018-06-23
Windows build instructions missing a step


bertram1isu:
Hi,

I've gone through both the linux and windows build steps now, and I noticed on the Windows side, there was a missing step in the section linked below that cmake should be installed.  The way the instructions read, I assumed that the Visual Studio 2017 CMake support contained an embedded cmake binary, but I eventually figured out cmake should be downloaded and installed separately.

https://github.com/Microsoft/AirSim/blob/master/docs/build_windows.md#build-airsim

- J.

bertram1isu:
Fixed it myself.  :)

PR #1191, opened 2018-06-23
More helpful build instructions for those who are getting started

bertram1isu:
I ran into a number of issues figuring out how to build and run AirSim for the first time.  Captured what would have been helpful to me here. 

Related to issue #1190.


sytelus:
Thank you. I have incorporated most of your suggestions in this commit: https://github.com/Microsoft/AirSim/commit/f910350d35a6414290ae00679ea58e60d1096da3

However please note that you don't need to install cmake on Windows explicitly because build.cmd should do that for you.

bertram1isu:
Then there is potentially a bug in build.cmd because it failed multiple times when I ran it.  It finally resolved itself when I manually installed cmake

PR #1263, opened 2018-07-19
Added instruction to run binaries in low resolution mode

cseas:


sytelus:
This seems to go better in binaries download page. I've made this change: https://github.com/Microsoft/AirSim/commit/661d0c41b045d182221eebd17e571a4530eca3fb

PR #1282, opened 2018-07-24
Minor bugfixes in VectorMath.

thias15:
Fix x,y,z,w calls. Normalize to vector in lookAt. Handle division by zero in slerp.


thias15:
The value for FLT_MIN is define in float.h. Since code check was not passed it is probably a good idea to define it. 

#define FLT_MIN          1.175494351e-38F        // min normalized positive value

Issue #1279, opened 2018-07-24
Failure to build Cmake Modules using g++


wardzacharyj:
#### Enviroment
System: Ubuntu 16.0.4LTS
CXX Compiler: g++-6
CC Compiler: gcc 6
CMake: 3.12 (I have tried 3.9 also)

#### Error
From my executable on line 27
```

#include "common/common_utils/StrictMode.hpp"
STRICT_MODE_OFF
#ifndef RPCLIB_MSGPACK
#define RPCLIB_MSGPACK clmdep_msgpack
#endif // !RPCLIB_MSGPACK
#include "rpc/rpc_error.h"
STRICT_MODE_ON

Line 27 --> #include "vehicles/multirotor/api/MultirotorRpcLibClient.hpp"
#include "common/common_utils/FileSystem.hpp"
#include <iostream>
#include <chrono>
```
#### Console Output
```
from /home/zachw/.../src/my_project/src/airsim_interface.cpp:27:
/home/zachw/Test/src/my_project/AirSim/AirLib/include/common/VectorMath.hpp: In static member function ‘static QuaternionT msr::airlib::VectorMathT<Vector3T, QuaternionT, RealT>::lookAt(Vector3T, Vector3T)’:
/home/zachw/Test/src/my_project/AirSim/AirLib/include/common/VectorMath.hpp:505:21: error: ‘acosf’ is not a member of ‘std’
         RealT ang = std::acosf(dot);
                     ^~~
/home/zachw/Test/src/my_project/AirSim/AirLib/include/common/VectorMath.hpp:505:21: note: suggested alternative:
In file included from /usr/include/c++/6/cmath:45:0,
                 from /usr/include/c++/6/random:38,
                 from /home/zachw/Test/src/my_project/AirSim/AirLib/include/common/common_utils/Utils.hpp:19,
                 from /home/zachw/Test/src/my_project/AirSim/AirLib/include/common/Common.hpp:14,
                 from /home/zachw/Test/src/my_project/AirSim/AirLib/include/vehicles/multirotor/api/MultirotorRpcLibClient.hpp:7,
                 from /home/zachw/Test/src/my_project/src/airsim_interface.cpp:27:
/usr/include/x86_64-linux-gnu/bits/mathcalls.h:54:1: note:   ‘acosf’
 __MATHCALL (acos,, (_Mdouble_ __x));
 ^
[...]

PR #1284, opened 2018-07-25
Fixes issue #1279

wardzacharyj:
This fixes an issue I ran into when compiling the Airsim libraries with g++-6. It seems acosf was not being included in the std name-space unless I compiled with Clang and its headers. I was also able to resolve this by including `<cmath>` to `VectorMath.hpp`, but I opted for this solution as math.h's ascof is already available in the global name-space.


sytelus:
Looks like more better solution is to use std::acos instead of std::acosf. I've added this change.

PR #1298, opened 2018-07-27
Kinect-like data publisher

Parnia:
This is a new feature in which Kinect-like data is published in ROS. 
The published topic are:
/camera/depth_registered/image_raw    (RGB image)
/camera/rgb/image_rect_color    (depth image)
/camera/rgb/camera_info    (camera Information including camera parameters)
/tf    (the relationship between coordinate frames)

You can use this python script to run RTAB-Map SLAM, or any other SLAMs.
This is a screen-shot of 3D map of RTAB-Map in Landscape Mountains environment, using data published by kinect_publisher.py:

screenshot from 2018-07-28 00-11-18


BrunoXL:
@Parnia how you made to run RTMAP slam? 
From [their tutorial](http://wiki.ros.org/rtabmap_ros/Tutorials/SetupOnYourRobot#Kinect) I tried to run kiinect only version but I got the error below that causes `rgbd_odometry` to crash:

       terminate called after throwing an instance of 'cv_bridge::Exception'
       what():  Image is wrongly formed: height * step != size  or  480 * 1920 != 110592

Dou you have any clue?

PR #1357, opened 2018-08-28
Sgm

thias15:
Integration of SGM for depth navigation and data collection.


thias15:
Discovered one compilation warning. Will fix it and then open new pull request.

PR #1364, opened 2018-08-31
Support (including docs) for symlinking Unreal project to AirSim source

KerryMoffittRtn:
We have been using symlinks in our workflow, as opposed to copies of source.  The approach offers these advantages:

- No confusion about which copy of code is being edited
- Fewer steps in workflow (no copying)
- Less disk space and file structure complexity required

Hope others might find this useful!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -279,6 +279,7 @@ ModelManifest.xml
 /AirLib/cmake_install.cmake
 /MavLinkCom/lib/libMavLinkCom.a
 /Unreal/Plugins/AirSim/Source/AirLib/
+/Unreal/Plugins/AirSim/Source/AirLib
 /Unreal/Environments/Blocks/Plugins/AirSim/
 /enc_temp_folder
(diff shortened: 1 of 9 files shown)


sytelus:
Hi Kerry, I understand this change but I'm wondering about additional complexity we need for sym links. Currently build is as-is bit complex than we would like so adding additional modes adds more burden on us. I'm wondering if there is less intrusive way to support this mode.

KerryMoffittRtn:
Yes, I totally get that; even as I was finalizing the branch, I was feeling the concern myself.  If we were starting from scratch we might simply make this the default, but I understand that the existing legacy carries some weight.  Documenting it side-by-side as a secondary option, as I've done, feels like a bit much, but here's one other idea, off the top of my head: revert the docs to the way they were, and then have one standalone doc, symlinks.md or something, that includes all the bits necessary for this.  At least if they're in one isolated place, people who don't care never have to look at any of it...  

Other thoughts?  

Thanks for considering, in any case!

PR #1378, opened 2018-09-05
Allow for auto-generating more than one vehicle at a time

KerryMoffittRtn:
Added optional 'Count' and 'SpacingOffsetCm' parameters for objects in
the 'vehicles' object in settings.json


sytelus:
Hi Kerry, we had thought about auto-generation options. We think the best way to do this is via adding new APIs that allows to spawn new vehicles. This will avoid getting settings more convoluted and also add much more flexibility because most projects will go well beyond auto generating through some simple positions.

KerryMoffittRtn:
Good plan!  Makes a lot of sense to me.  Cheers.

Issue #621, opened 2017-11-19
Build LIDAR capabilities in AirSim


shrimalmadhur:
I'll be adding LIDAR capabilities. 
ETA: ~~12/4~~ TBD

mitchellspryn:
@shrimalmadhur Did you mean 12/4? It's currently 11/19...

shrimalmadhur:
Thanks. Updated!

sytelus:
Thanks. It would be great if you can summarize/outline your design.

shrimalmadhur:
I saw your comments in some other Issue that says to extend the SensorBase class the same way other Sensors have extended it. And then I can use the Raycasting function which is already there in UAirBluePrintLib. I am planning to do it in the same way.

tejaskhot:
Would this not be available for use with the drone?

sytelus:
@tejaskhot yes. All AirSim sensors are reusable across vehicle types.

nnayeer:
Any updates? Eagerly awaiting this.

shrimalmadhur:
Got caught up in other stuff. Will update the issue once I have some update.

realdealneil:
I am particularly interested in adding a laser altimeter to the simulator (or at least having ground truth for height-above-terrain).  I'm guessing that's a subset of what you're planning to do?

tejaskhot:
Is there any update on this?

benasena:
I have most of the pieces of this working. I did the ray trace in VehiclePawnWrapper, then passed it through MultiRotorConnector to a variable added to environment to get it in a distance sensor class based on the barometer implementation. I added the necessary reporters and MAVLink handlers, and can see the measurement in QGroundControl in the analyze widget and if I run "listener distance_sensor" in nsh it all looks good. I set EKF2_HGT_MODE=2 and it flies well in all modes.

For some reason it still doesn't seem to follow terrain, though. Am I missing parameters or is there something about the HIL firmware that ignores distance sensors? I believe I correctly implemented the members of the DISTANCE_SENSOR MAVLink message (http://mavlink.org/messages/common). My units are all centimeters there and I get meters when I run the nsh command. I did use Unix epoch time instead of boot time. Could that be it? Is there a good way to an appropriate value for that in AirSim?

nnayeer:
[...]

PR #1424, opened 2018-09-27
Add support for Lidar sensor

bmalhigh:
Add Lidar sensor in AirLib object model as well as Unreal
implementation using ray-casting. Add python client example.

Future work:
    - Enable configuration of lidar parameters
    - Enable visualization of point-cloud on client side

Resolves: #621


Unreal/Plugins/AirSim/Source/UnrealSensors/UnrealLidarSensor.cpp
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// returns a point-cloud for the tick
+msr::airlib::vector<msr::airlib::real_T> UnrealLidarSensor::getPointCloud(const msr::airlib::Pose& pose, const float deltaTime)
+{
+    points_.clear();
+
+    msr::airlib::LidarSimpleParams params = getParams();
+
+    const auto numberOfLasers = params.number_of_channels;
```
sytelus: would be good to use _ for local vars in this method for consistency.

Unreal/Plugins/AirSim/Source/UnrealSensors/UnrealLidarSensor.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+    FHitResult hit_result = FHitResult(ForceInit);
+    bool is_hit = UAirBlueprintLib::GetObstacle(actor_, start, end, hit_result);
+
+    if (is_hit)
+    {
+        if (false)
+        {
+            // Debug code.
+            // don't enable this for multirotors since drawing a debug point
```
sytelus: You can use `UAirBlueprintLib::IsInGameThread()` to check if you are allowed to do draw calls

Unreal/Plugins/AirSim/Source/UnrealSensors/UnrealLidarSensor.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+    // calculate final rotation for this laser point w.r.t. lidar pose
+    // there might be a better way of achieving this via AirSim VectorMath but this implementation should work as well
+    FQuat lidarOrientation(pose.orientation.x(), pose.orientation.y(), pose.orientation.z(), pose.orientation.w());
+    FRotator lidarRot = lidarOrientation.Rotator();
+    FRotator resultRot = UKismetMathLibrary::ComposeRotators(laserRot, lidarRot);
+
+    // calculate end position (params.range is expected to be in meters)
+    constexpr float TO_CENTIMETERS = 1e+2f;
+    FVector end = (params.range * TO_CENTIMETERS) * UKismetMathLibrary::GetForwardVector(resultRot) + start;
```
sytelus: I'm wondering if this computation can be bit cleaned up. We have 3 coordinate frames: world frame, vehicle frame, lidar frame. The usual way to do computation is to start with inner most frame and then ultimately get the vector in world frame. Also, it's often useful to do all computation in just one system (for example, either NED or UE). Usually we just keep everything in NED system and convert to UE system only when UE API call is needed. This keeps code portable to other engines (we have Unity port in works).

Below is just pseudo code. The suffix _w, _b and _l are used for world, body and lidar frames.

We have:
```
ray_h_l, ray_v_l = horizontal and vertical angles for ray in lidar frame
lidar_pose_b = lidar pose relative to vehicle body
vehicle_pose_w = vehicle pose relative to world frame

//get ray quaternion in lidar frame (angles must be in radians)
ray_q_l = VecorMath::toQuaternion(ray_v_l, 0, ray_h_l)
//get ray quaternion in body frame
ray_q_b = VectorMath::rotateQuaternion(ray_q_l, lidar_pose_b.orientation)
//get ray quaternion in world frame
ray_q_w = VectorMath::rotateQuaternion(ray_q_b, vehicle_pose_w.orientation)
//get ray vector
ray_w = VectorMath::rotateVector(VectorMath::front() * max_distance, ray_q_w) 
   + lidar_pose_b.position + vehicle_pose_w.position
```

Unreal/Plugins/AirSim/Source/Vehicles/Multirotor/MultirotorPawnSimApi.cpp
```diff
@@ -49,6 +49,8 @@ void MultirotorPawnSimApi::pawnTick(float dt)
 {
     unused(dt);
     //calls to update* are handled by physics engine and in SimModeWorldBase
+
+    drawLidarDebugPoints();
```
sytelus: If we move this out of PawnSimAPi to SimModeBase, this won't needed because visualization code will be agnostic to vehicle type.

AirLib/include/vehicles/car/api/CarApiBase.hpp
```diff
@@ -63,21 +66,105 @@ class CarApiBase : public VehicleApiBase  {
...
+    {
+        return sensors_;
+    }
+
+    SensorCollection& getSensors()
+    {
+        return sensors_;
+    }
+
+    void initialize(std::shared_ptr<SensorFactory> sensor_factory, const Kinematics::State& state, const Environment& environment)
```
sytelus: Just wondering if we should just add this to constructor? That way someone using this class won't forget to call initialize.

sytelus (commented):
Looks really good! Most of the comments are just minor nitpicks. I think we are very close to merging these changes.

KerryMoffittRtn:
Oh, this is great; thanks for doing it!  One thought: consider using the existing sim depth sensor, and resampling from the fixed-width / fixed-height sample deltas that one gets from a Kinect / flash lidar / existing sim-depth-sensor, instead of simulating each individual laser fiting.  That would presumably provide a giant speedup, since you can use the existing (GPU-based) render pipeline, vs. performing all the collision calculations on the CPU.  Cheers!

AirLib/include/api/RpcLibAdapatorsBase.hpp
```diff
@@ -461,6 +461,40 @@ class RpcLibAdapatorsBase {
             return response_adapter;
         }
     };
+
+    struct LidarData {
+
+        // header
```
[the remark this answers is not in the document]
bmalhigh: My intent here was that "header" and "data" were more like section headers... with the assumption that there will be multiple fields in header section along the way... but fixed for now.

AirLib/include/api/VehicleApiBase.hpp
```diff
@@ -94,6 +96,30 @@ class VehicleApiBase : public UpdatableObject {
...
+    virtual const SensorCollection& getSensors() const
+    {
+        throw VehicleCommandNotImplementedException("getSensors API is not supported for this vehicle");
+    }
+
+    // Lidar APIs
+    virtual LidarData getLidarData(const std::string& lidar_name) const
+    {
+        // TODO: find the appropriate Lidar based on the name
+        const LidarBase* lidar = static_cast<const LidarBase*>(getSensors().getByType(SensorBase::SensorType::Lidar));
```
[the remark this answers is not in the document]
bmalhigh: AFAIK, no. SensorCollection::getByType(...) returns only one SensorBase back. For now fixing it by using the combination of size() and using index in getByType().

AirLib/include/api/VehicleApiBase.hpp
```diff
@@ -94,6 +96,30 @@ class VehicleApiBase : public UpdatableObject {
...
+        if (lidar == nullptr)
+        {
+            throw VehicleControllerException("No matching lidar found on the vehicle");
+        }
+
+        LidarData lidarData;
```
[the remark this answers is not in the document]
bmalhigh: Fixed

AirLib/include/common/CommonStructs.hpp
```diff
@@ -295,5 +295,14 @@ struct RCData {
     }
 };
 
+struct LidarData {
+
+    TTimePoint time_stamp = 0;
+    vector<float> point_cloud;
```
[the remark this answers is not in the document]
bmalhigh: I guess we are not being consistent with real_T vs. float... I see other places in AirLib where we are using vector<float>... for example -- JoyStick::axes, ImageResponse::image_data_float... but I changed the LidarData::point_cloud to use real_T

AirLib/include/sensors/lidar/LidarBase.hpp
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#ifndef msr_airlib_LidarBase_hpp
+#define msr_airlib_LidarBase_hpp
+
+#include "sensors/SensorBase.hpp"
+
+namespace msr {
```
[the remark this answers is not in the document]
bmalhigh: Fixed

AirLib/include/vehicles/multirotor/MultiRotorParams.hpp
```diff
@@ -32,6 +32,7 @@ class MultiRotorParams {
         bool gps = true;
         bool barometer = true;
         bool distance = false; //this causes ray casts so disabled by default
+        bool lidar = true;     //this causes ray casts so disabled by default: TODO
```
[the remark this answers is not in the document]
bmalhigh: Yes, that was the intent. Fixed.

Unreal/Plugins/AirSim/Source/UnrealSensors/UnrealLidarSensor.cpp
bmalhigh: Yep. Hopefully fixed everywhere now.

Unreal/Plugins/AirSim/Source/UnrealSensors/UnrealLidarSensor.h
```diff
@@ -0,0 +1,37 @@
...
+
+    void CreateLasers();
+    bool ShootLaser(const msr::airlib::Pose& pose, uint32 channel, float horizontalAngle, 
+        msr::airlib::LidarSimpleParams params, Vector3r &point);
+
+private:
+    AActor* actor_;
+    const NedTransform* ned_transform_;
+
+    TArray<float> laser_angles_;
```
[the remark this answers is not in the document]
bmalhigh: Fixed.

Unreal/Plugins/AirSim/Source/PawnSimApi.cpp
```diff
@@ -10,6 +10,8 @@
 #include "NedTransform.h"
 #include "common/EarthUtils.hpp"
 
+#include "DrawDebugHelpers.h"
```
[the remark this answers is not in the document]
bmalhigh: Moved to SimModeBase.

Unreal/Plugins/AirSim/Source/Vehicles/Multirotor/MultirotorPawnSimApi.cpp
bmalhigh: Fixed.

Unreal/Plugins/AirSim/Source/UnrealSensors/UnrealLidarSensor.cpp
bmalhigh: Fixed

AirLib/include/vehicles/car/api/CarApiBase.hpp
bmalhigh: Done. It required to pass the params through the intermediate base class; so not very clean looking... but I made the change for now.

AirLib/include/sensors/lidar/LidarSimple.hpp
```diff
@@ -0,0 +1,99 @@
...
+                reporter.writeValue("Lidar-Range", params_.range);
+                reporter.writeValue("Lidar-FOV-Upper", params_.vertical_FOV_Upper);
+                reporter.writeValue("Lidar-FOV-Lower", params_.vertical_FOV_Lower);
+            }
+            //*** End: UpdatableState implementation ***//
+
+            virtual ~LidarSimple() = default;
+
+        protected:
+            virtual vector<real_T> getPointCloud(const Pose& pose, float deltaTime) = 0;
```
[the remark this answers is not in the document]
bmalhigh: Done.

AirLib/include/api/VehicleApiBase.hpp
```diff
@@ -94,6 +96,30 @@ class VehicleApiBase : public UpdatableObject {
...
+        LidarBase::Output output = lidar->getOutput();
+        
+        lidarData.time_stamp = output.time_stamp;
+        lidarData.point_cloud = std::vector<float>(output.point_cloud);
```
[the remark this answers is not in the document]
bmalhigh: Done

bmalhigh:
> Oh, this is great; thanks for doing it! One thought: instead of simulating each individual laser firing, consider using the existing sim depth sensor, and resampling from the fixed-width / fixed-height sample deltas that one gets from a Kinect / flash lidar / existing sim-depth-sensor to the fixed-angle deltas one gets with a spinning lidar like a traditional Velodyne. That would presumably provide a giant speedup, since you could use the existing (GPU-based) render pipeline, vs. performing all the collision calculations on the CPU. Cheers!

@KerryMoffittRtn -- I agree with the goal to reduce the number of lasers fired or the number of calcs done. However, I am unable to understand the solution that you are proposing above. Apologies for being dense. Can you clarify the above resampling approach in bit more detail?

Cheers.

KerryMoffittRtn:
Sure!  As I understand your approach, you cast a ray per laser firing, and test for collision in the world geometry.  This seems sound, and will presumably provide the functionality you desire, though it means asking Unreal to navigate the geometry of the scene to perform collision detection (presumably on the CPU) for every ray cast.  To understand the alternative approach I am describing, first consider the existing depth capture mechanism (see ImageType::DepthVis, USceneCaptureComponent2D, and the picture-in-picture that the '1' key toggles).  It works a lot like the standard scene renderer (heavily optimized, and taking full advantage of the parallelism offered by the GPU) except that it yields depth information instead of color information.  Now this is super fast, and outputs a 2D array of depth samples (lots of samples, i.e.) at once, _but_ they're not geometrically structured the way you want: the rays all pass through a focal plane, on which we have a fixed width and height between adjacent samples.  What you want for a spinning lidar is fixed _angle_ between adjacent samples.  So you would need to reorganize the data by resampling, and the cost of this resampling (though it may not be terrible) would need to be weighed against the fast performance of the initial depth sampling.  (You'd also need to collect samples on multiple planes, of course - at least three - to cover all 360 degrees of a circle.)  This is a little tricky, I realize.  Happy to chat more if I'm still not making sense.  :)

TritonSailor:
This is a very useful feature. If performance needs to be optimized, it would probably be helpful to ask in the unreal forums how to most efficiently implement this. Also, you might find some useful code in the Nvidia Waveworks project. It uses ray casting (on the gpu) to determine bouyancy of actors at various locations on the map.

AirLib/include/api/VehicleApiBase.hpp
```diff
@@ -11,6 +11,8 @@
 #include "safety/SafetyEval.hpp"
 #include "common/CommonStructs.hpp"
 #include "common/ImageCaptureBase.hpp"
+#include "sensors/SensorCollection.hpp"
+#include "sensors/Lidar/LidarBase.hpp"
```
anphunl: Should be #include "sensors/lidar/LidarBase.hpp", shouldn't it?

andre-nguyen:
@KerryMoffittRtn Is that how Gazebo does it? I think I saw something like that for the velodyne simulator in Gazebo and I was a bit confused as to why we would do it using cameras. 

Is there no way of manually raytracing on the gpu without going through the rendering pipeline?

AirLib/include/api/VehicleApiBase.hpp
SamHS-OCI: I think you're right, I was having build errors, replaced the suggested line and now it is working. Thanks.

bmalhigh: Ah, Linux issue... apologies... fixing now.

PR #1454, opened 2018-10-11
Fix the include error

chenhan1218:


chenhan1218:
Previously, the project will build fail on Ubuntu 18.04 which ship clang-5.0 version 5.0.1-4

With the fix, the project would be built successfully.

bmalhigh (approved):
Looks good... my bad with my previous checkin. Thanks for fixing.

PR #1463, opened 2018-10-13
Update README.md

daksh-sard:


sytelus:
Could you please provide the description of the change? For multilingual documentation we need a different approach?

rajnishc8 (commented):
Hindi Translation is very bad and lot of it makes no sense at all.

PR #1526, opened 2018-11-10
Add support for Unity on Linux 

msb336:
EditorUtility is not available when outside of Unity Editor. Depending on the editor causes project builds to fail


AirLib/include/vehicles/multirotor/firmwares/mavlink/MavLinkMultirotorApi.hpp
```diff
@@ -57,7 +57,7 @@ class MavLinkMultirotorApi : public MultirotorApiBase
         }
         catch (std::exception& ex) {
             is_ready_ = false;
-            is_ready_message_ = Utils::stringf("Failed to connect: %s", ex.what());
+            is_ready_message_ = msr::airlib::Utils::stringf("Failed to connect: %s", ex.what());
```
sytelus: Why is this prefix needed? These classes are in same namespace.

AirLib/include/vehicles/car/api/CarApiBase.hpp
```diff
@@ -67,9 +67,8 @@ class CarApiBase : public VehicleApiBase  {
     };
 
 public:
-
     // TODO: Temporary constructor for the Unity implementation which does not use the new Sensor Configuration Settings implementation.
-	//CarApiBase() {}
+	CarApiBase() {}
```
sytelus: Could you please rebase your branch so we don't get remerge of old commits?

Unity/AirLibWrapper/AirsimWrapper/Source/UnityToAirSimCalls.h
```diff
@@ -17,7 +17,9 @@ void StartServerThread(std::string vehicle_name, std::string sim_mode_name, int
 
 extern "C" EXPORT bool StartServer(char* vehicle_name, char* sim_mode_name, int port_number)
 {
-	LOGGER->WriteLog("Starting server for : " + std::string(sim_mode_name));
+	#if _WIN32
+		LOGGER->WriteLog("Starting server for : " + std::string(sim_mode_name));//**
```
sytelus: Should there be log call for Linux?

Unity/UnityDemo/Assets/AirSimAssets/Scripts/InitializeAirSim.cs
```diff
@@ -11,59 +11,52 @@ void Awake()
...
-            if (AirSimSettings.Initialize())
-            {
-                if (AirSimSettings.GetSettings().SimMode == "")
+			if(AirSimSettings.Initialize())
+			{
+
+                switch (AirSimSettings.GetSettings().SimMode)
                 {
-                    if (Application.isEditor)
+                    case "Car":
```
sytelus: This doesn't seem right. There is prompt statement which should happen if its editor mode.

Unity/UnityDemo/Assets/AirSimAssets/Scripts/InitializeAirSim.cs
```diff
@@ -95,11 +88,12 @@ public static bool CreateSettingsFileWithDefaultValues(string fileName)
...
-    private void LoadSceneAsPerSimMode()
+    public void LoadSceneAsPerSimMode(string load_name)
     {
-        if (AirSimSettings.GetSettings().SimMode == "Car")
+        Debug.Log("loading sim");
+        if (load_name == "Car")
             SceneManager.LoadSceneAsync("Scenes/CarDemo", LoadSceneMode.Single);
-        else if (AirSimSettings.GetSettings().SimMode == "Multirotor")
+        else if (load_name == "Multirotor")
             SceneManager.LoadSceneAsync("Scenes/DroneDemo", LoadSceneMode.Single);
```
sytelus: Just wondering if this works on Windows and Linux

Unity/UnityDemo/ProjectSettings/ProjectVersion.txt
```diff
@@ -1 +1 @@
-m_EditorVersion: 2018.2.14f1
+m_EditorVersion: 2017.2.0b11
```
sytelus: Is this version change required?

Unity/AirLibWrapper/AirsimWrapper/cmake/airlib-setup.cmake
```diff
@@ -0,0 +1,18 @@
+macro(BuildAirLib)
```
sytelus: Just wondering if we should be using same code as in https://github.com/Microsoft/AirSim/tree/master/cmake
In that case, can we just copy these files in build.sh from parent folder so we don't have two copies to maintain?

sytelus (changes requested):

AirLib/include/vehicles/multirotor/firmwares/mavlink/MavLinkMultirotorApi.hpp
msb336: When building one wrapper library using the source files for both AirLib and Airsim wrapper, the compiler complains about re-definitions of type names inside of a few files inside of AirSimWrapper/include, namely PawnSimApi.h. I needed to change one set of typedefs or the other.

The typedefs here were easier to trace and change than in AirSimWrapper.

Unity/UnityDemo/Assets/AirSimAssets/Scripts/InitializeAirSim.cs
msb336: I changed the GUI layout so that the prompt is built into the scene canvas. Editor popups can't build to Unity projects, and the prompt for "Choose sim mode" needs to be in the built project as well as the editor

Unity/AirLibWrapper/AirsimWrapper/Source/UnityToAirSimCalls.h
msb336: Yes there should. I had that off for debugging purposes, forgot to turn it back on.

Unity/UnityDemo/ProjectSettings/ProjectVersion.txt
msb336: No, that's a difference between my unity version and whatever version was on the original commit for unity.

PR #1527, opened 2018-11-11
Incorporating changes from the CarApiBase class

kenbier:
This fixes the latest build issue in the Unity projects. Changes to the CarApiBase class needed to be incorporated into the AirlibWrapper.


kenbier:
@ sytelus It would be great if we could add Windows building to the CI. Then we will know when any API changes break the Unity project, since we've only added Windows build support for now. 

Looks like they've added it here:
https://blog.travis-ci.com/2018-10-11-windows-early-release

PR #1534, opened 2018-11-14
Docs for Unity release

madratman:
Updates to documentation for Unity release


kenbier:
@madratman Thanks for this!

PR #1535, opened 2018-11-14
Lidar feature updates

bmalhigh:
1) Add support for configuring horizontal FOV
2) Add support for obtaining point-cloud in lidar frame
3) Expose lidar pose as part of lidar output

Validated on Windows 10 and Ubuntu 16.04

See lidar.md for the example configs.


Unreal/Plugins/AirSim/Source/UnrealSensors/UnrealLidarSensor.cpp
```diff
@@ -70,15 +70,24 @@ void UnrealLidarSensor::getPointCloud(const msr::airlib::Pose& lidar_pose, const
     const float angle_distance_of_tick = params.horizontal_rotation_frequency * 360.0f * delta_time;
     const float angle_distance_of_laser_measure = angle_distance_of_tick / points_to_scan_with_one_laser;
 
+    const float laser_start = std::fmod(360.0 + params.horizontal_FOV_start, 360.0f);
+    const float laser_end = std::fmod(360.0 + params.horizontal_FOV_end, 360.0f);
+
```
sytelus: You might want to use std::mod and attach f to 360.0. GCC doesn't seem to like fmod.

Unreal/Plugins/AirSim/Source/UnrealSensors/UnrealLidarSensor.cpp
```diff
@@ -135,8 +146,34 @@ bool UnrealLidarSensor::shootLaser(const msr::airlib::Pose& lidar_pose, const ms
             );
         }
 
-        point = ned_transform_->toLocalNed(hit_result.ImpactPoint);
-       
+        // decide the frame for the point-cloud
+        if (params.data_frame == "" || params.data_frame == "VehicleInertialFrame") {
+            // current detault behavior; though it is probably not very useful.
```
sytelus: It might be good idea to define const for VehicleInertialFrame and init params.data_frame to that value as default. That way if you change default in future, you will just need to make change at one place instead of checking if-then statements at many places.

Unreal/Plugins/AirSim/Source/UnrealSensors/UnrealLidarSensor.cpp
```diff
@@ -70,15 +70,24 @@ void UnrealLidarSensor::getPointCloud(const msr::airlib::Pose& lidar_pose, const
...
     // shoot lasers
     for (auto laser = 0u; laser < number_of_lasers; ++laser)
     {
+        const float vertical_angle = laser_angles_[laser];
+
         for (auto i = 0u; i < points_to_scan_with_one_laser; ++i)
         {
+            // check if the laser is outside the requested horizontal FOV
+            const float horizontal_angle = std::fmod(current_horizontal_angle_ + angle_distance_of_laser_measure * i, 360.0f);
+            if (horizontal_angle < laser_start && horizontal_angle > laser_end)
```
sytelus: std::fmod -> std::mod

sytelus: Should this be <= instead of < and >= instead of >

Unreal/Plugins/AirSim/Source/UnrealSensors/UnrealLidarSensor.cpp
```diff
@@ -70,15 +70,24 @@ void UnrealLidarSensor::getPointCloud(const msr::airlib::Pose& lidar_pose, const
...
+    const float laser_start = std::fmod(360.0 + params.horizontal_FOV_start, 360.0f);
+    const float laser_end = std::fmod(360.0 + params.horizontal_FOV_end, 360.0f);
+
     // shoot lasers
     for (auto laser = 0u; laser < number_of_lasers; ++laser)
     {
+        const float vertical_angle = laser_angles_[laser];
+
         for (auto i = 0u; i < points_to_scan_with_one_laser; ++i)
         {
```
sytelus: Just wondering how is time synchronized between frames when lidar is rotating slow, for example, 1 rotation per sec.

Unreal/Plugins/AirSim/Source/UnrealSensors/UnrealLidarSensor.cpp
```diff
@@ -94,30 +103,32 @@ void UnrealLidarSensor::getPointCloud(const msr::airlib::Pose& lidar_pose, const
...
     // get ray quaternion in lidar frame (angles must be in radians)
     msr::airlib::Quaternionr ray_q_l = msr::airlib::VectorMath::toQuaternion(
         msr::airlib::Utils::degreesToRadians(vertical_angle),   //pitch - rotation around Y axis
         0,                                                      //roll  - rotation around X axis
-        msr::airlib::Utils::degreesToRadians(horizontalAngle)); //yaw   - rotation around Z axis
+        msr::airlib::Utils::degreesToRadians(horizontal_angle));//yaw   - rotation around Z axis
 
     // get ray quaternion in body frame
-    msr::airlib::Quaternionr ray_q_b = VectorMath::rotateQuaternion(ray_q_l, lidar_pose.orientation, true);
+    msr::airlib::Quaternionr ray_q_b = VectorMath::coordOrientationAdd(ray_q_l, lidar_pose.orientation);
```
sytelus: Perf improvement opportunity: During a frame pose doesn't change so you have to compute many of the things only once instead of each iteration of for loop.

docs/lidar.md
```diff
@@ -23,19 +23,59 @@ NumberOfChannels          | Number of channels/lasers of the lidar
 Range                     | Range, in meters
 PointsPerSecond           | Number of points captured per second
```
sytelus: Can we add little blurb on lidar API in apis.md and link to here?

AirLib/include/api/RpcLibAdapatorsBase.hpp
```diff
@@ -469,8 +469,9 @@ class RpcLibAdapatorsBase {
```
sytelus: You might want to check if Unity build gets broken because of these changes. To do that, get latest and run Unity\build.cmd and see if compile is success. We need to add this in travis but it will need some investigation because travis only recently enabled Windows builds. So right now if some code change breaks Unity we won't know. In longer, we probably need to eliminate most of the Unreal and Unity related code.

sytelus (changes requested):

Unreal/Plugins/AirSim/Source/UnrealSensors/UnrealLidarSensor.cpp
bmalhigh: Are you sure this is an issue? We have been using std::fmod already in UnrealLidarSensor.cpp and also in other places (VectorMath.hpp, AngleLevelController.hpp),etc.

w.r.t. std::mod, did you mean std::modulus?

Unreal/Plugins/AirSim/Source/UnrealSensors/UnrealLidarSensor.cpp
bmalhigh: Not sure I understand the question in the context of this code. This particular line is just getting the specific laser to shoot. 

The number of points to shoot for each laser is calculated before this code based on the number of points per/sec, rotations per sec and time-delta in the tick.

Issue #1541, opened 2018-11-14
Build Error MavLinkCom C1001


Waarbubble:
Having just re Installed my windows 10 PC I proceeded to install 

- Visual Studio 2017 enterprise v. 15.9.0, with windows 8.1 SDK
- Unreal Engine 4.21.0
 - AirSim

but running the build.cmd I get the following error:

> r:\program files\airsim\mavlinkcom\common_utils\type_utils.hpp(50): fatal error C1001: An internal error has occurred i n the compiler. [R:\Program Files\AirSim\MavLinkCom\MavLinkCom.vcxproj] 
  (compiler file 'd:\agent\_work\3\s\src\vctools\compiler\cxxfe\sl\p1\c\types.c', line 4563)
   To work around this problem, try simplifying or changing the program near the locations listed above.
  Please choose the Technical Support command on the Visual C++
   Help menu, or open the Technical Support help file for more information
  r:\program files\airsim\mavlinkcom\common_utils\type_utils.hpp(59): note: see reference to class template instantiati
  on 'mavlink_utils::type_utils::detail::has_begin_end<T>' being compiled
  ThreadUtils.cpp
  AdHocConnection.cpp
  AdHocConnectionImpl.cpp

Any one know what is wrong?

micmarty:
I have the exact same issue. It looks like we're missing some Visual Studio component or something... I'll let you know when I find the solution :)

forestscorpion:
i have the exact same issue too -.-

nicksunyang:
me too

simon-achatz:
I'm having the same issue, already when running 'build.cmd' in root folder the first time.

arguskos:
Same error for me, but even from the root folder of the AirSim

nicksunyang:
I reinstall vs2017 , the problem can't be fixed

nicksunyang:
i also use Unreal Engine 4.18.3, uninstalled  4.19.2

Waarbubble:
A work around i found was to install visual studio 2017 v 15.6.8

madratman:
Hey everyone. Yeah, the issue is with VS 2017 15.9.0, which was released a couple days back on Nov 13. 
I tested with 15.8.9, and things do compile. Please uninstall 15.9.0 and install 15.8.9, as explained here (links for 15.8.9 therein)
https://docs.microsoft.com/en-us/visualstudio/productinfo/installing-an-earlier-release-of-vs2017

[...]

PR #1549, opened 2018-11-15
[readme] VS 15.9.0 is breaking airsim. use 15.8.9

madratman:
partial resolve for #1541

### DIFF
--- a/Unity/README.md
+++ b/Unity/README.md
@@ -15,10 +15,12 @@ This project is still in early development, expect some rough edges. We are work
 * Note: If you are using Unity for the first time, check out [the Getting started guide](https://docs.unity3d.com/Manual/GettingStarted.html). The [Unity User Manual](https://docs.unity3d.com/Manual/UnityManual.html) has additional tips, resources, and FAQs.
 
 #### Build Airsim
-* Install Visual Studio 2017, with **VC++** and **Windows SDK 8.x**.
-* Run `x64 Native Tools Command Prompt for VS 2017`. 
+* Install [Visual Studio 2017, ver <= 15.8.9](https://docs.microsoft.com/en-us/visualstudio/productinfo/installing-an-earlier-release-of-vs2017).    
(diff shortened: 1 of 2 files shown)


Unity/README.md
```diff
@@ -15,10 +15,14 @@ This project is still in early development, expect some rough edges. We are work
 * Note: If you are using Unity for the first time, check out [the Getting started guide](https://docs.unity3d.com/Manual/GettingStarted.html). The [Unity User Manual](https://docs.unity3d.com/Manual/UnityManual.html) has additional tips, resources, and FAQs.
 
 #### Build Airsim
-* Install Visual Studio 2017, with **VC++** and **Windows SDK 8.x**.
-* Run `x64 Native Tools Command Prompt for VS 2017`. 
+* Install [Visual Studio 2017, ver 15.8.9](https://docs.microsoft.com/en-us/visualstudio/productinfo/installing-an-earlier-release-of-vs2017)
+.    
+**Make sure** to select **VC++** and **Windows SDK 8.1** while installing VS 2017.   
+Note that **VS2017 `15.9.0`** would **not** work. Please choose `15.8.9` in the link above. If you have `15.9.0`, please downgrade as explained in the [link above](https://docs.microsoft.com/en-us/visualstudio/productinfo/installing-an-earlier-release-of-vs2017).   
+
```
sytelus: I would suggest we keep docs concise and as short as possible. Also, probably not good idea to give impression to users that they must have only one specific version. VS installs takes too long and folks won't be too happy about uninstalls of versions that they may already have and would work fine otherwise. The bleeding edge versions of VS will keep breaking and they often get bug fixed in few days. All that would make instructions too noisy. In past, the way we have handled this is by simply adding a line in FAQ section at the end of the page that would be something like this:

```
Q: I'm getting "xyz error" when running build.cmd.
A: We have noticed this happening with VS version 15.9.0. If you have this version, please uninstall it and use older and newer version.
```

The advantage here is that this will pop up in search results right away as most users search for error messages.

sytelus (changes requested):

PR #1587, opened 2018-11-29
fix problem with multirotor sim on PX4 drones.

lovettchris:
I think I found a bug.   The Environment object being used by Barometer is not the same Environment being updated by MultirotorPawnSimApi.cpp.  For some reason we have two separate Environments…
```cpp
PawnSimApi::PawnSimApi(const Params& params)
    : params_(params), ned_transform_(params.pawn, *params.global_transform)
{
    msr::airlib::Environment::State initial_environment;
    initial_environment.position = getPose().position;
    initial_environment.geo_point = params_.home_geopoint;
    environment_.reset(new msr::airlib::Environment(initial_environment));
```
and
```cpp
    MultiRotor(MultiRotorParams* params, VehicleApiBase* vehicle_api, 
        const Pose& initial_pose, const GeoPoint& home_geopoint)
        : params_(params), vehicle_api_(vehicle_api)
    {
        auto initial_kinematics = Kinematics::State::zero();
        initial_kinematics.pose = initial_pose;
        Environment::State initial_environment;
        initial_environment.position = initial_kinematics.pose.position;
        initial_environment.geo_point = home_geopoint;
        environment_.reset(new Environment(initial_environment));

        initialize(initial_kinematics, environment_.get());
    }
```
Changing MultiRotor constructor so that we can pass in the Environment* instead solves the problem.

This fixes takeoff on PX4 v1.8.1.

### DIFF
--- a/PythonClient/multirotor/land.py
+++ b/PythonClient/multirotor/land.py
@@ -11,7 +11,7 @@ if landed == airsim.LandedState.Landed:
     print("already landed...")
 else:
     print("landing...")
-    client.land()
+    client.landAsync()
 
 client.armDisarm(False)
 client.enableApiControl(False)
--- a/PythonClient/multirotor/land.py
+++ b/PythonClient/multirotor/land.py
@@ -11,7 +11,7 @@ if landed == airsim.LandedState.Landed:
     print("already landed...")
 else:
     print("landing...")
-    client.landAsync()
+    client.landAsync().join()
 
 client.armDisarm(False)
 client.enableApiControl(False)
--- a/Unreal/Plugins/AirSim/Source/PawnSimApi.cpp
+++ b/Unreal/Plugins/AirSim/Source/PawnSimApi.cpp
@@ -295,12 +295,6 @@ void PawnSimApi::update()
     environment_->update();
     //kinematics_->update();
 
-    // and in the case of subclasses that are overriding getGroundTruthEnvironment, make sure those 
-    // are also updated.
-    msr::airlib::Environment* vehicleEnvironment = const_cast<msr::airlib::Environment*>(getGroundTruthEnvironment());
-    vehicleEnvironment->setPosition(kinematics_.pose.position);
-    vehicleEnvironment->update();
-
     VehicleSimApiBase::update();
 }
 
@@ -515,6 +509,11 @@ const msr::airlib::Kinematics::State* PawnSimApi::getGroundTruthKinematics() con
     return &kinematics_;
 }
 const msr::airlib::Environment* PawnSimApi::getGroundTruthEnvironment() const
+{
+    return environment_.get();
+}
+
+msr::airlib::Environment* PawnSimApi::getEnvironment() const
 {
     return environment_.get();
 }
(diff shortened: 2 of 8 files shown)


PythonClient/multirotor/land.py
```diff
@@ -11,7 +11,7 @@
     print("already landed...")
 else:
     print("landing...")
-    client.land()
+    client.landAsync()
```
sytelus: This will continue code while landing is not complete. Does this needs to be done because some other issue?

Unreal/Plugins/AirSim/Source/PawnSimApi.cpp
```diff
@@ -418,6 +418,7 @@ void PawnSimApi::setPoseInternal(const Pose& pose, bool ignore_collision)
     //translate to new PawnSimApi position & orientation from NED to NEU
     FVector position = ned_transform_.fromLocalNed(pose.position);
     state_.current_position = position;
+    kinematics_.pose = pose;
```
sytelus: Do we still to set pose in kinematics?

Unreal/Plugins/AirSim/Source/PawnSimApi.cpp
```diff
@@ -512,6 +513,11 @@ const msr::airlib::Environment* PawnSimApi::getGroundTruthEnvironment() const
     return environment_.get();
 }
 
+msr::airlib::Environment* PawnSimApi::getEnvironment() const
+{
```
sytelus: Why not use getGroundTruthEnvironment?

sytelus (changes requested):

Unreal/Plugins/AirSim/Source/PawnSimApi.cpp
lovettchris: Yes, still need this because we also have two kinematics_ objects.  One in PawnSimApi and another in PhysicsBody, and I checked their addresses, they are not the same object.  Seems the connection is from MultirotorPawnSimApi who creates Multirotor which is the PhysicsBody, so here's another duplication going on...

Unreal/Plugins/AirSim/Source/PawnSimApi.cpp
lovettchris: has to be non-const.

Unreal/Plugins/AirSim/Source/PawnSimApi.cpp
lovettchris: Here's another fix that works, inside PawnSimApi.cpp, namely, use getGroundTruthKinematics instead of kinematics_.  But it would be nice to rationalize this duplication of kinematics state also...what do you recommend?

```cpp
void PawnSimApi::update()
{
    //update position from kinematics so we have latest position after physics update
    environment_->setPosition(getGroundTruthKinematics()->pose.position);
    environment_->update();
    //kinematics_->update();

    VehicleSimApiBase::update();
}
```

PythonClient/multirotor/land.py
lovettchris: added .join().

PR #1613, opened 2018-12-07
Make WindowsTargetPlatformVersion more easily configurable in AirSim

lovettchris:
This change adds a shared MSBuild file called AirSim.props so that it is easier to change the WindowsTargetPlatformVersion in your AirSim build.  This makes it easier for folks to build AirSim using Windows 10 SDK, since installing the Windows 8.1 SDK is a bit of a pain these days.


AirLib/AirLib.vcxproj
```diff
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), AirSim.props))\AirSim.props" />
```
sytelus: How does this affect experience from building and using IDE? For example, can I still override this option from IDE or do I must use new prop file?

AirSim.sln
```diff
@@ -46,131 +46,163 @@ EndProject
...
 		Debug|x64 = Debug|x64
 		Debug|x86 = Debug|x86
 		Release|Any CPU = Release|Any CPU
+		Release|ARM = Release|ARM
 		Release|x64 = Release|x64
 		Release|x86 = Release|x86
 	EndGlobalSection
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
 		{9FE9234B-373A-4D5A-AD6B-FB0B593312DD}.Debug|Any CPU.ActiveCfg = Debug|Win32
+		{9FE9234B-373A-4D5A-AD6B-FB0B593312DD}.Debug|ARM.ActiveCfg = Debug|Win32
```
sytelus: Curious if Win32 and Arm build is actually working...

DroneServer/log
```diff
@@ -0,0 +1,125 @@
+Microsoft (R) Build Engine version 15.9.20+g88f5fadfbe for .NET Framework
```
sytelus: Do we need this file?

build.cmd
```diff
@@ -180,6 +180,7 @@ robocopy /MIR MavLinkCom\lib %MAVLINK_TARGET_LIB%
 REM //---------- all our output goes to Unreal/Plugin folder ----------
 if NOT exist Unreal\Plugins\AirSim\Source\AirLib mkdir Unreal\Plugins\AirSim\Source\AirLib
 robocopy /MIR AirLib Unreal\Plugins\AirSim\Source\AirLib  /XD temp *. /njh /njs /ndl /np
+copy /y AirSim.props Unreal\Plugins\AirSim\Source\AirLib
```
sytelus: Would it be better to just mark .prop file as Copy if newer for build output?

sytelus (changes requested):

AirLib/AirLib.vcxproj
clovett: VS IDE build is the same, and it would be best to change the AirSim.props instead of manually changing every VS project to the new version you want.  Note that after changing AirSim.props you may need to restart VS in order to force this new version to be loaded.  If you edit the projects directly it would probably undo the change I'm making here by adding <WindowsTargetPlatformVersion>…</…> back to all the .vcxproj files, which kind of defeats the point of this change.

AirSim.sln
clovett: VS did this, not me.  Not sure why.

DroneServer/log
lovettchris: removed

build.cmd
lovettchris: that would copy build output, this is copying sources.  But I could add a special msbuild copy instruction to copy the file to .\AirLib so it gets picked up by the robocopy.  but then people would check it in.  So I think this is better for now.

AirLib/AirLib.vcxproj
lovettchris: github swallowed my take name, editing the project properties would add the WindowsTargetPlatformVersion tag back to the .vcxproj files, which is not what you want.

lovettchris: it might be possible to add a new MSBuild rule also to the AirSim.props to check for that and educate the user on the right way to modify this property.

PR #1635, opened 2018-12-16
fix deadlock and docs and improve some scripts

### LINKED ISSUE #719: Px4 unreal  freezes everytime when exiting the simulation
It doesn't seem to leave any trace in the terminal as where it fails.
I am not using the sitl but just remote control,it crashes every time I try to stop the simulation and freezes the entire  screen..
OS : Linux 14.04
Unreal version 4.17.2
Latest Airsim version

[PX4.log](https://github.com/Microsoft/AirSim/files/1602328/PX4.log)

lovettchris:
the deadlock was happening when in the middle of flying path.py you click "stop" in the Unreal editor.  The app would not stop because of deadlock in rpcserver.  The server stop() method hangs because there is another thread already being processed by rpc server, and that thread won't complete either because you clicked stop and other stuff has already shut down.  This fixes [bug 719](https://github.com/Microsoft/AirSim/issues/719).

### DIFF
--- a/AirLib/src/api/RpcLibServerBase.cpp
+++ b/AirLib/src/api/RpcLibServerBase.cpp
@@ -275,9 +275,6 @@ void RpcLibServerBase::start(bool block, std::size_t thread_count)
 
 void RpcLibServerBase::stop()
 {
-    for (VehicleApiBase* api : api_provider_->getVehicleApis()) {
-        api->reset(); // terminate any pending actions
-    }
     pimpl_->server.close_sessions();
     pimpl_->server.stop();
 }
--- a/AirLib/src/api/RpcLibServerBase.cpp
+++ b/AirLib/src/api/RpcLibServerBase.cpp
@@ -275,6 +275,10 @@ void RpcLibServerBase::start(bool block, std::size_t thread_count)
 
 void RpcLibServerBase::stop()
 {
+    for (VehicleApiBase* api : api_provider_->getVehicleApis()) {
+        api->reset(); // terminate any pending actions
+    }
+    pimpl_->server.close_sessions();
     pimpl_->server.stop();
 }
 
--- a/PythonClient/multirotor/box.py
+++ b/PythonClient/multirotor/box.py
@@ -8,7 +8,16 @@ client = airsim.MultirotorClient()
 client.confirmConnection()
 client.enableApiControl(True)
 client.armDisarm(True)
-client.takeoffAsync().join()
+
+landed = client.getMultirotorState().landed_state
+if landed == airsim.LandedState.Landed:
+    print("taking off...")
+    client.takeoffAsync().join()
+
+landed = client.getMultirotorState().landed_state
+if landed == airsim.LandedState.Landed:
+    print("takeoff failed - check Unreal message log for details")
+    sys.exit(1)
 
 print("Flying a small square box using moveByVelocityZ")
 print("Try pressing 't' in the AirSim view to see a pink trace of the flight")
@@ -46,3 +55,12 @@ print("moving by velocity vx=" + str(vx) + ", vy=" + str(vy) + ", yaw=0")
 client.moveByVelocityZAsync(vx, vy,z,duration, airsim.DrivetrainType.MaxDegreeOfFreedom, airsim.YawMode(False, 0)).join()
 time.sleep(delay)
 client.hoverAsync().join()
+
+print("ensure we are back home")
+client.moveToPositionAsync(0, 0, z, 1).join()
+
+print("landing...")
+client.landAsync().join()
+
+print("disarming.")
+client.armDisarm(False)
--- a/PythonClient/multirotor/path.py
+++ b/PythonClient/multirotor/path.py
@@ -1,14 +1,27 @@
+
+# see https://github.com/Microsoft/AirSim/wiki/moveOnPath-demo
+# this path is calculated to work nicely with the Neighborhood Map, assuming the 
+# start location is set to unreal map coordinates (310, 11200, 121)
 import setup_path 
 import airsim
 
 import sys
 import time
 
+def print_position(client):    
+    state = client.getMultirotorState()
+    gps = state.gps_location
+    print("gps location lat={}, lon={}, alt={}".format(gps.latitude, gps.longitude, gps.altitude))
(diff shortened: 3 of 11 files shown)


AirLib/src/api/RpcLibServerBase.cpp
```diff
@@ -275,6 +275,10 @@ void RpcLibServerBase::start(bool block, std::size_t thread_count)
 
 void RpcLibServerBase::stop()
 {
+    for (VehicleApiBase* api : api_provider_->getVehicleApis()) {
```
sytelus: I think destructor of class that implements VehicleApiBase should be cleaning up things. Calling reset() during stop is probably hacky workaround. Also, one should be able to start-stop RpcServer without stopping the game. So cleanup should be in response to EndPlay, not for RpcServer stop. In hypothetical case, imagine there are multiple API servers active (ROS, ZeroMQ etc) and user can turn off or turn on any one at any time.

PythonClient/multirotor/box.py
```diff
@@ -8,7 +8,16 @@
...
+
+landed = client.getMultirotorState().landed_state
+if landed == airsim.LandedState.Landed:
+    print("taking off...")
+    client.takeoffAsync().join()
+
+landed = client.getMultirotorState().landed_state
+if landed == airsim.LandedState.Landed:
+    print("takeoff failed - check Unreal message log for details")
+    sys.exit(1)
```
sytelus: I'm confused by this piece of code... would this work for simple_flight?

PythonClient/multirotor/path.py
```diff
@@ -1,33 +1,65 @@
...
+
 landed = client.getMultirotorState().landed_state
 if landed == airsim.LandedState.Landed:
     print("taking off...")
     client.takeoffAsync().join()
 else:
     client.hoverAsync().join()
 
+
+if client.getMultirotorState().landed_state == airsim.LandedState.Landed:
```
sytelus: Again confused by need to check landed state. One big issue is that landed state interpretation is hard and there would always be uncertainty. So my thought was that it is good to have as informational value but probably code shouldn't be dependent on this.

sytelus (changes requested):

AirLib/src/api/RpcLibServerBase.cpp
clovett: I found a bug that should be fixed by this pull request.  See [bug 719](https://github.com/Microsoft/AirSim/issues/719).

PythonClient/multirotor/path.py
sytelus: I think raising exception in MavLink code would be good idea. Landed state is probably not accurately  supported in either MavLink or simpleflight and most definitely not for DJI. So I think using it might cause reliability issue.

Unreal/Plugins/AirSim/Source/SimMode/SimModeBase.cpp
```diff
@@ -321,6 +321,11 @@ void ASimModeBase::reset()
             api->reset();
         }
     }, true);
+
+    // stop any pending RPC actions  (which avoids deadlocks on rpc thread)
```
sytelus: This shouldn’t be required because. VehicleApiBase is owned by VehicleSimApiBase and so reset on later calls reset on former.

Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.cpp
```diff
@@ -47,8 +47,10 @@ void ASimHUD::Tick(float DeltaSeconds)
 
 void ASimHUD::EndPlay(const EEndPlayReason::Type EndPlayReason)
 {
-    if (simmode_)
+    if (simmode_) {
```
sytelus: I am wary of using reset to free up acquired resources becUse reset can do lots of things depending on vehicle type. So the code in simmode must imply semantics required for all possible general simulations.  If we want to free up MavLink related resources or fix related issue, I think we should avoid altering the general code that would have much larger consequences. I’m thinking for your current issue, you can probably use destructors in MavLink implementor classes so you can eliminate changes needed at the simmode level.

Unreal/Plugins/AirSim/Source/SimMode/SimModeWorldBase.cpp
```diff
@@ -119,7 +119,7 @@ void ASimModeWorldBase::reset()
         physics_world_->reset();
     }, true);
     
-    //no need to call base reset because of our custom implementation
+    Super::reset();
```
sytelus: This is fairly hairy...

sytelus (changes requested):

PythonClient/multirotor/box.py
clovett: Why is checking LandedState vehicle specific?  It is in the drone API therefore it is API, and not drone specific by definition.  Now turns out the MavLink implementation DOES throw an exception, but the exception never makes it to the python script.  Not sure why, but that is another bug in the RPC stack somewhere.

PythonClient/multirotor/path.py
clovett: Mavlink code IS raising an exception, but it disappears and never makes it back to the python script.

Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.cpp
clovett: It is not a Mavlink specific issue.  It is an RPC issue.  One RPC call comes in asking to moveOnPath, and that call stays alive for a long time, then you click Stop which calls EndPlay.  EndPlay calls stop on the RPC server without stopping moveOnPath call, and the RPC server stop call hangs, presumable waiting for the moveOnPath to finish, but it cannot finish because EndPlay has been called, and the simulator has terminated, and so you have a deadlock.  The reset call I added causes the cancelToken to be set and that causes the moveOnPath call to terminate and then everything shuts down properly.

Unreal/Plugins/AirSim/Source/SimMode/SimModeWorldBase.cpp
clovett: It's the only way I could find a way to move my fix to EndPlay as you requested.

Issue #1516, opened 2018-11-04
Simulator crashes after arming (PX4 SITL)


petergerten:
AirSim: master
PX4: master
OS: Ubuntu 16.04

I set up my AirSim project with PX4 SITL and I succesfully get MAVLink data in QGroundControl.
As soon as I issue the arm command (and I still get and "armed" back in QGC) - AirSim crashes.

Most of the time I get:
```
terminating with uncaught exception of type std::runtime_error: UdpClientPort socket send failed with error: 111

Signal 6 caught.
Malloc Size=131076 LargeMemoryPoolOffset=131092 
CommonLinuxCrashHandler: Signal=6
Malloc Size=65535 LargeMemoryPoolOffset=196655 
Malloc Size=41839 LargeMemoryPoolOffset=238511 
Engine crash handling finished; re-raising signal 6 for the default handler. Good bye.
Aborted (core dumped)
```

Sometimes I get:
```
terminating with uncaught exception of type std::runtime_error: Message length 32 doesn't match expected length33

Signal 6 caught.
Malloc Size=131076 LargeMemoryPoolOffset=131092 
CommonLinuxCrashHandler: Signal=6
Malloc Size=65535 LargeMemoryPoolOffset=196655 
Malloc Size=41624 LargeMemoryPoolOffset=238296 
Engine crash handling finished; re-raising signal 6 for the default handler. Good bye.
Aborted (core dumped)
```

But it reproducibly crashes every single time. Before I switched to the PX4 SITL I used the simple flight controller setup which runs stable with my current setup. If I don't arm the system but just let the simulation run and produce mavlink output - everything works and does not crash.

petergerten:
This is probably a very dirty hack but I just tried what would happen if I comment out
```
		if (hr == SOCKET_ERROR)
		{
			hr = WSAGetLastError();
			// perhaps the client is gone, and may want to come back on a different port, in which case let's reset our remote port to allow that.
			remoteaddr.sin_port = 0;
			throw std::runtime_error(Utils::stringf("UdpClientPort socket send failed with error: %d\n", hr));
}
```

from UdpClientPort.cpp 

This seems to work for me. Of course it does not solve the underlying issue but it does not crash the application anymore.

petergerten:
@RVBldr can you clarify what that means?

RVBldr:
We're tracking this issue on our internal Visual Studio bug database.

petergerten:
[...]

PR #1654, opened 2018-12-22
Fix some bugs with mavlink2 support

### LINKED ISSUE #1388: Linux binary crashing with PX4 SITL
Trying to run the Linux Binary with PX4 SITL setup and it crashes before starting. It works fine without the SITL setup though. 
Here'e the error log. 
linuxerror

clovett:
This pull request fixes some bugs in mavlinkcom support for mavlink 2.  Specifically it was not handling the new MAVLINK_MSG_ID_AUTOPILOT_VERSION message which tells all MavLinkConnections what versionh we are using (including those that are joined).  This might also fix [GitHub issue 1388](https://github.com/Microsoft/AirSim/issues/1388) and [1516](https://github.com/Microsoft/AirSim/issues/1516).


MavLinkCom/src/impl/MavLinkNodeImpl.cpp
```diff
@@ -58,7 +58,15 @@ void MavLinkNodeImpl::sendHeartbeat()
...
         heartbeat.system_status = 0; // ignored by PX4
-        sendMessage(heartbeat);
+        try 
+        {
+            sendMessage(heartbeat);
+        }
+        catch (std::exception& e)
+        {
+            // ignore any failures here because we are running in our own thread here.
+            Utils::log(Utils::stringf("Caught and ignoring exception sending heartbeat: %s", e.what()));
```
sytelus: This is bit worry some. Why would exception occur and is there is no way to transfer this information somehow up the chain? What would be the worse case scenario here? May be exception occurs and user gets confused that nothing is happening?

MavLinkCom/src/impl/MavLinkVehicleImpl.cpp
```diff
@@ -148,7 +148,8 @@ AsyncResult<bool> MavLinkVehicleImpl::allowFlightControlOverUsb()
 
 void MavLinkVehicleImpl::handleMessage(std::shared_ptr<MavLinkConnection> connection, const MavLinkMessage& msg)
 {
-    unused(connection);
+    MavLinkNodeImpl::handleMessage(connection, msg);
```
sytelus: Any idea why this change?

MavLinkCom/src/MavLinkMessages.cpp
```diff
@@ -5101,6 +5101,36 @@ std::string MavLinkDebug::toJSon() {
  return ss.str();
 }
 
+int MavLinkProtocolVersion::pack(char* buffer) const {
```
sytelus: Would all these continue to work with MavLink 1.0 as well?

sytelus (commented):
Overall looks good and if you are happy we can merge it.

MavLinkCom/src/impl/MavLinkNodeImpl.cpp
clovett: I agree it is not ideal, but this function is running in an independent thread, and there is no other thread that we can transfer this to.

MavLinkCom/src/impl/MavLinkVehicleImpl.cpp
clovett: MavLinkNodeImpl needs the handleMessage also.

MavLinkCom/src/MavLinkMessages.cpp
clovett: yes.

Vinay545:
Are the linux binaries compiled with this patch. Because I am facing this issue with the existing binary releases. If not, when the new binaries that are compiled with this patch will be released.

Issue #1579, opened 2018-11-27
Unity - Confusion about where settings.json file should be located?


jasonhbartlett:
Win 10 Pro
Unity 2018.2.14f1 - Running in editor (hit Play)
VS Community 2017 15.8.8
SDK 8.1

ref: [https://github.com/Microsoft/AirSim/issues/1550](url)

Although it was pointed out in issue #1550 that the settings.json file needed to be placed in the C:\Program Files\Unity\Editor folder, I am confused as to why this is the case?  Having no settings.json file there causes the error "Problem in starting AirSim server!!! Please check logs for more information." and kills the Unity editor.  This is not the behavior AirSim typically has as typically the file is created automatically if it is missing, albiet it is typically created in the Documents/AirSim folder.

Also if not getting an error (cause if found the file in the editor folder) it WILL STILL create a settings.json file in the Documents/AirSim folder.  And I'm led to suspect that this is actually where the project settings are pulled from?

So, where exactly is the settings.json file needed?  And where are the settings pulled from?

msb336:
Unity calls for settings.json inside "InitializeAirSim.cs", which is located here:
Unity\UnityDemo\Assets\AirSimAssets\Scripts

```
    public static string GetAirSimSettingsFileName()
    {
        string fileName = Application.dataPath + "\\..\\settings.json";
        if (File.Exists(fileName))
        {
            return fileName;
        }

        fileName = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.MyDocuments), Path.Combine("AirSim", "settings.json"));
        if (File.Exists(fileName))
        {
            return fileName;
        }

        if (CreateSettingsFileWithDefaultValues(fileName))
            return fileName;
        else
            return string.Empty;
    }
```


Unity looks for settings.json inside of Documents\Airsim inside of the project directory. If it can't find either it will build a default  settings.json file inside of Documents\Airsim.

jasonhbartlett:
@msb336 Just poking around.  I believe the error message in the Unity console is coming from:
AirSim/Unity/AirLibWrapper/AirsimWrapper/Source/SimHUD/SimHUD.cpp Line 147.
This section is where it tries to read the file from the same location as the Unity.exe Editor executable.  Don't know why we are looking there for the settings file?  A.) We already are checking this in the Unity scripts and B.) Should only be looking in Documents\AirSim.  Correct?

msb336:
@jasonhbartlett 
[...]

PR #1684, opened 2019-01-09
bug fixes #1656

### LINKED ISSUE #1656: AirSim on Unity installation error
Hello, everyone:
When I tried to run ./build.sh in the AirSim/Unity/build.sh, the following error is raised:

./build.sh: line 14: cmake: command not found
make: *** No targets specified and no makefile found.  Stop.
cp: cannot stat 'libAirsimWrapper.so': No such file or directory

The previous steps works pretty fine. Does anyone know the solution?

[OS]
Ubuntu 18.04

msb336:
Stops AirSim from looking inside of the unity executable folder for settings.json file and removes the need to have cmake installed globally on your machine.

### DIFF
--- a/Unity/build.sh
+++ b/Unity/build.sh
@@ -7,11 +7,13 @@ if [ ! -d "linux-build" ]; then
     mkdir linux-build;
 fi
 
+
 cd linux-build;
 export CC="clang-5.0"
 export CXX="clang++-5.0"
+CMAKE="$(readlink -f ../../cmake_build/bin/cmake)"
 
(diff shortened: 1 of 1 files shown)


msb336:
All references in the documentation point to placing your settings.json file in the Documents/AirSim directory. There have been some issues brought up about the settings.json file location in unity, so it seems the most foolproof way to stop unexpected behavior for users is to force AirSim to only look in one location.

I'll remove the section of the commit related to #1579 , but we should update the docs to inform users that AirSim will look in the executable directory first.

sytelus (changes requested):

PR #1713, opened 2019-01-17
Gui for settings file selection

msb336:
Allows users to [choose from all json files inside documents/airsim directory from a dropdown menu before sim start](
https://www.youtube.com/watch?v=oFCB2wC9vjY).


jasonhbartlett:
This is pretty cool.  Should be pretty easy to implement in Unity as well.

Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.h
```diff
@@ -7,9 +7,13 @@
...
 #include "api/ApiServerBase.hpp"
 #include <memory>
+#include "SettingsWidget.h"
+
 #include "SimHUD.generated.h"
 
 
+
+
 UENUM(BlueprintType)
```
madratman: we should remove all extraneous empty lines added

Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.h
```diff
@@ -48,6 +52,15 @@ class AIRSIM_API ASimHUD : public AHUD
     virtual void EndPlay(const EEndPlayReason::Type EndPlayReason) override;
     virtual void Tick(float DeltaSeconds) override;
 
+    UPROPERTY(EditAnywhere, BlueprintReadWrite, Category = "Settings Widget")
```
madratman: rename uproperty to Settings Widget Loader to avoid conflict with the settings widget below

Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.cpp
```diff
@@ -43,6 +58,11 @@ void ASimHUD::Tick(float DeltaSeconds)
 {
     if (simmode_ && simmode_->EnableReport)
         widget_->updateDebugReport(simmode_->getDebugReport());
+    if ( settings_widget_->checkForInput() && not_built_ ) {
+        UE_LOG(LogTemp, Warning, TEXT("There was an input!"));
+        buildScene();
+        not_built_ = false;
```
madratman: instead of calling buildScene() in tick() and checking if not_built_ is false in every subsequent tick, why not call it  in void ASimHUD::BeginPlay()? 
Does that crash unreal?

madratman (commented):
It works OTB, added comments. most are just formatting comments. 
The main thing to check is if  buildScene() can be called in beginplay and wait to continue till user input is received?

Unreal/Plugins/AirSim/Source/SimHUD/SettingsWidget.h
```diff
@@ -0,0 +1,47 @@
+// Fill out your copyright notice in the Description page of Project Settings.
+
+#pragma once
+#ifdef _WIN32
```
madratman: we can remove this ifdef clause

Unreal/Plugins/AirSim/Source/SimHUD/SettingsWidget.h
```diff
@@ -0,0 +1,47 @@
+// Fill out your copyright notice in the Description page of Project Settings.
```
madratman: is this needed or from some boilerplate?

madratman (commented):
minor update in review

Unreal/Plugins/AirSim/Source/SimHUD/dirent.h
```diff
@@ -0,0 +1,1224 @@
...
+ 
+/*
+ * Include windows.h without Windows Sockets 1.1 to prevent conflicts with
+ * Windows Sockets 2.0.
+ */
+#ifndef WIN32_LEAN_AND_MEAN
+#   define WIN32_LEAN_AND_MEAN
+#endif
+#include <windows.h>
+
```
madratman: #include<windows.h> is problematic for linux ofc
Interestingly, I updated the PR branch some changes and now on compiling the second time, include<windows.h> is throwing an error. 
I am not sure why it compiled on the first try in ubuntu 16.04

madratman:
changes needed were pretty minor, and I overlayed a commit over Matt's awesome work. 
@sytelus this is ready for a merge.

madratman:
force pushed to fix commit authors

Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.cpp
sytelus: Agreed. We need to avoid more code in tick().

Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.cpp
```diff
@@ -23,18 +25,28 @@ void ASimHUD::BeginPlay()
...
-        initializeSettings();
+        createSettingsWidget();
+    }
+    catch (std::exception& ex) {
+        UAirBlueprintLib::LogMessageString("Error at startup: ", ex.what(), LogDebugLevel::Failure);
+        UAirBlueprintLib::ShowMessage(EAppMsgType::Ok, std::string("Error at startup: ") + ex.what(), "Error");
+    }
+}
+
+void ASimHUD::buildScene()
```
sytelus: HUD is not supposed to create scene so buildScene name is confusing.  May be createWidgets or initialize or something like that?

Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.cpp
```diff
@@ -243,13 +279,15 @@ void ASimHUD::setupInputBindings()
...
+{   
+    std::string settingsFile = std::string(TCHAR_TO_UTF8(*ue4_settings_string));
     std::string settingsText;
-    if (getSettingsText(settingsText))
+    
+    if (getSettingsText(settingsText, settingsFile))
         AirSimSettings::initializeSettings(settingsText);
     else
-        AirSimSettings::createDefaultSettingsFile();
+        AirSimSettings::createDefaultSettingsFile("settings_default.json"); // do not overwrite user's settings.json!
```
sytelus: What is going on here? We don't create settings.json? Please write comment if there is a major change from how things worked previously.

Unreal/Plugins/AirSim/Source/SimHUD/dirent.h
```diff
@@ -0,0 +1,1224 @@
+/*
```
sytelus: What is this file and why is it needed?

sytelus (changes requested):

Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.cpp
madratman: so the widget has an option to select the default settings.json, which calls AirSimSettings::createDefaultSettingsFile(), which (over)writes the user's settings.json in this line https://github.com/Microsoft/AirSim/blob/85fe8d4987339d5f32f1cf8ffc56b9f64636f789/AirLib/include/common/AirSimSettings.hpp#L406 in the current master branch. 

I think a lot of our users would have a custom file at their /home/$USER/Documents/AirSim/settings.json, and the widget should NOT overwrite that. 

Therefore, I added a default parameter for the local filename in this line https://github.com/Microsoft/AirSim/pull/1713/files#diff-4ee334619d0e207657a29f2139412385R396 to not overwrite settings.json, but instead create a settings_default.json. 

uh, while I am typing this, I do realize that perhaps I should just have defined another function / added a boolean parameter to createDefaultSettingsFile which would simply by pass the saveJsonFile() command https://github.com/Microsoft/AirSim/blob/85fe8d4987339d5f32f1cf8ffc56b9f64636f789/AirLib/include/common/AirSimSettings.hpp

lemme know what you think is better.

Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.cpp
msb336: If placed in BeginPlay the widget does not continue to check for user input after BeginPlay is called

Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.cpp
sytelus: Either way it would work. I'm just trying to understand workflow... I'm hoping user has option to modify json whichever they like. So they may modify using notepad or using UX. For example, I might modify in Notepad, save it and then reload it in UX and modify again. Would this work? Also, does default settings.json gets created on startup (if it doesn't exist)?

Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.cpp
sytelus: That doesn't sound right. Widget shouldn't need to be in game loop. That would be no-go scenario. Typical game might have dozens of complex widgets and if all of them need to be ticked all the time, it won't fly. This widget won't be shown 99.99% of the time but would get ticked all the time.

Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.cpp
```diff
@@ -172,6 +192,21 @@ void ASimHUD::inputEventToggleAll()
     updateWidgetSubwindowVisibility();
 }
```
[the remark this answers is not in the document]
sytelus: +1 this is important and often causes issue on Linux because git converts CRLF on fly.

sytelus (changes requested):

Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.cpp
msb336: The only json creation is when the default (empty) settings is selected. Otherwise, the widget only produces a dropdown menu of json files located in the Documents/AirSim directory.
[demo here](https://studio.youtube.com/video/oFCB2wC9vjY)

Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.cpp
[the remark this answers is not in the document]
madratman: Haha, I end up always using git in WSL when working in windows. Anyway, this should be resolved.. I think I fixed that in my commit.
In general, we do have a mix of tabs and spaces in airsim code https://youtu.be/SsoOG6ZeyUI

PR #1745, opened 2019-01-27
Add Unity Linux specific data recorder changes

arian-amador:
Adds check for running platform and adjusts the storage for data recording results.
It stores images and ground truth in the default settings.json location ~/Documents/AirSim/* with the proper forward slashes on Linux. 
Documentation was updated as well.


sytelus:
Hi, have you tested this change in Windows and Linux? Also wondering if this would work in OSX.

arian-amador:
@sytelus I've tested this on `Windows 10 Home Version 1803` and `Ubuntu 18.04.1 LTS` and it's working as expected. 
Unfortunately, I am unable to test in OSX.

PR #1748, opened 2019-01-29
SetSegmentationID assining wrong key

mustfkeskin:
Hello,

If I want to change segmentationId with regex there is logical problem.
First contains is not regex.

Second for example myRegex = Ground*
Contains command match some index but assingn wrong key.

If code change like below assign correct indices.

**Oldcode:** segmentationIds[objectName] = segmentationId;
**Newcode:** segmentationIds[s] = segmentationId;

### DIFF
--- a/Unity/UnityDemo/Assets/AirSimAssets/Scripts/HUD/CameraFiltersScript.cs
+++ b/Unity/UnityDemo/Assets/AirSimAssets/Scripts/HUD/CameraFiltersScript.cs
@@ -1,8 +1,10 @@
 ﻿using UnityEngine;
 using UnityEngine.Rendering;
 using System.Collections.Generic;
+using System.IO;
 
 namespace AirSimUnity {
+
     /*
      * MonoBehaviour class that is attached to cameras in the scene.
      * Used for applying image filters based on settings.json or Image request by the client to record the data.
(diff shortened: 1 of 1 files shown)


ykpgrr (changes requested):
I suggest you to be careful when submitting a Pull Request. You deleted one namespace (System.IO) and so many blank lines (the blank lines are necessary for clean code).
- You deleted `using System.IO` in line 4. You shouldn't.
- in line 63: there is extra `isNameRegex` --> `if (isNameRegex) {isNameRegex`
- in line 65: `keylist` should be `keyList` (upper case **L**) because the dictionary was initialized like `keyList`.

Thanks for your contribution, happy coding :)

ykpgrr (approved):
This changes seems to eliminate a bug which is caused for this situation:
let's assume:
`isNameRegex = True` , `objectName = "Ground"` and there is a key in "KeyList" Dictionary like "Ground_2" so `s="Ground_2"`.
The **bug** is that the code trying to change "Ground" in KeyList Dictionary. But there is no **key** like "Ground", it should be "Ground_2". This causes the code to fail. With changes 
`segmentationIds[objectName] = segmentationId;` as
`segmentationIds[s] = segmentationId;` **will be prevent this bug.**

On the other hand; I think there is still no regex in this algorithm although there is parameter for regex like `bool isNameRegex`. This can be another PR for this algorithm.

Happy coding :)

howard-mahe:
Hi, this PR can be closed since PR #1859 has been merged.

madratman:
thanks @howard-mahe !

Issue #1694, opened 2019-01-11
Deadlock is possible when recording is stopped


yanenkoa:
As of writing this text, it is currently possible to run into deadlock when stopping recording thread.

In PR #1505, a new method of capturing images from cameras and cameras' positions in sync with each other was introduced: the recording thread is adding a delegate function to call on a GameThread whenever drawing ends (`UGameViewportClient::OnEndDraw`). This code is in [`Unreal/Plugins/AirSim/Source/RenderRequest.cpp`](https://github.com/Microsoft/AirSim/blob/5246acd664c8cf4aeb1070c0e9ea64fdcf442ddd/Unreal/Plugins/AirSim/Source/RenderRequest.cpp#L63-L95).

Problems begin when we want to stop the recording. This is done by calling [`FRecordingThread::stopRecording`](https://github.com/Microsoft/AirSim/blob/5246acd664c8cf4aeb1070c0e9ea64fdcf442ddd/Unreal/Plugins/AirSim/Source/Recording/RecordingThread.cpp#L53-L60), which calls [`FRecordingThread::EnsureCompletion`](https://github.com/Microsoft/AirSim/blob/5246acd664c8cf4aeb1070c0e9ea64fdcf442ddd/Unreal/Plugins/AirSim/Source/Recording/RecordingThread.cpp#L114-L119), which sends a signal to recording thread to stop execution and then waits for it to actually finish.

`FRecordingThread::stopRecording` is **always** called in the GameThread. So whenever it is called, GameThread starts waiting for FRecordingThread to finish. The issue is, FRecordingThread can be waiting for drawing to end, and drawing is happening on a GameThread. So the situation is possible, when GameThread is waiting for FRecordingThread to finish, while FRecordingThread is waiting for a specific event on the GameThread to fire — a deadlock.

The result is, obviously enough, that UE4 app freezes completely and stops reacting to any sort of input and can only be closed by killing the process. In our project, we start and stop recording programmatically in an endless loop, so it's only a matter of time when this occurs and the simulation needs to be restarted manually. This can also occur when Play-In-Editor session is interrupted by pressing "Stop" button from inside UE4 editor or when `exit` command is executed in UE4 console.

[...]

PR #1750, opened 2019-01-29
Extend default recording of images/poses to all vehicles in a scene

saihv:
The currently default recording code (record button in-game, or key R) only records images and position/orientation data for the first vehicle in the game. This PR extends that functionality to all vehicles, by dealing with the full list of vehicle APIs (`VehicleSimApiBase` objects) rather than just one. For `n` vehicles in a scene, recording `n` images and `n` poses in `airsim_rec.txt` is treated as one step.

Note: This has only been tested on drones for up to five vehicles.


sytelus:
Thanks. This is good change but extensive as well. I'll go through this early next week.

saihv:
Just a quick note from my testing: it looks like having multiple vehicles recording simultaneously is causing AirSim to be more prone to the Game <-> Recording thread deadlock issue mentioned in https://github.com/Microsoft/AirSim/issues/1694. Utilizing the fix provided in the related PR: https://github.com/Microsoft/AirSim/pull/1695 seems to fix the problem for the multi vehicle recording case as well.

FYI, if it's helpful in any way: extending the recording to multiple vehicles does not cause the deadlock problem in v1.2.0 (Windows), only in the latest versions.

zhuboshun2015:
@saihv Could you give more details about how to achieve this ? I have modified the code(Extend default recording of images/poses to all vehicles) as you  upload,

 but next step is ?

zhuboshun2015:
> 谢谢。这是一个很好的改变，但也很广泛。下周初我会经历这个。

Hello ! Did you achieve that?

saihv:
@zhuboshun2015 If you've already adapted these changes into your code, all you'd have to do is press 'R' as you usually would, and images/poses from all drones will be recorded.

zhuboshun2015:
@saihv Yes, I did.But when I look at the.txt file, it seems that there is only the position and posture information of drone1.I don't know if I have to set 'settings.json'? This is my settings.json document.
[settings.txt](https://github.com/Microsoft/AirSim/files/3145059/settings.txt)

zhuboshun2015:
@saihv  thank you ! I have success in doing this . But it is a little difficult to recognize every drone's number.How do you recognize this ?

PR #1797, opened 2019-02-19
added stalebot

msb336:
Stalebot is an automoderator that will mark inactive issues as stale and close them accordingly. Currently it is set to mark issues as stale after 60 days, and close them 30 days after they are marked.


sytelus:
Please do not add this. We should not be closing issues because they are inactive. These issues comes up in search all the time and people comment all the time. The idea is not to artificially lower the issue count. People take lot of effort to write up issue and randomly closing them is very bad idea and signal to the community.

Issue #1657, opened 2018-12-23
IncrementFlushOnDraw is not a member of FSceneViewport.


smilefacehh:
Does anybody know how to solve this? I can't find the IncrementFlushOnDraw  method in class FSceneViewport. Thx.

MathieuCesar:
I realized I was using the latest version of the Unreal Engine (4.21) after installing using the default command:
git clone https://github.com/EpicGames/UnrealEngine.git
Instead, using the recommended version of the Unreal Engine (4.18) solved the problem:
git clone -b 4.18 https://github.com/EpicGames/UnrealEngine.git

madratman:
We had a few fixes for UE 4.21 last week in https://github.com/Microsoft/AirSim/commit/a179075a2a546537098c11cd1dcc1e8ff6589612. Might be worth trying again

hardsv:
I am facing the same issue ....using ue4.21.2 and Airsim latest version... downloaded it today.


1>------ Build started: Project: Blocks, Configuration: DebugGame_Editor x64 ------
1>Creating makefile for BlocksEditor (changes to module files)
1>While compiling D:\work\AirSim\AirSim\Unreal\Environments\Blocks\Intermediate\Build\BuildRules\BlocksModuleRules.dll:
1>d:\work\AirSim\AirSim\Unreal\Environments\Blocks\Plugins\AirSim\Source\AirSim.Build.cs(50,17): warning CS0618: 'UnrealBuildTool.ModuleRules.Definitions' is obsolete: 'The 'Definitions' property has been deprecated. Please use 'PublicDefinitions' instead.'
1>d:\work\AirSim\AirSim\Unreal\Environments\Blocks\Plugins\AirSim\Source\AirSim.Build.cs(51,17): warning CS0618: 'UnrealBuildTool.ModuleRules.Definitions' is obsolete: 'The 'Definitions' property has been deprecated. Please use 'PublicDefinitions' instead.'
1>d:\work\AirSim\AirSim\Unreal\Environments\Blocks\Plugins\AirSim\Source\AirSim.Build.cs(55,17): warning CS0618: 'UnrealBuildTool.ModuleRules.Definitions' is obsolete: 'The 'Definitions' property has been deprecated. Please use 'PublicDefinitions' instead.'
[...]

PR #1888, opened 2019-04-15
UE4.22: Fix deprecation warnings and compilation errors.

edowson:
Fixes the following warnings:
```
The 'Definitions' property has been deprecated. Use 'PublicDefinitions' instead.
```
Fixes the following compile errors:
```
error C2039: 'IncrementFlushOnDraw': is not a member of 'FSceneViewport'
```

### DIFF
--- a/Unreal/Plugins/AirSim/Source/AirSim.Build.cs
+++ b/Unreal/Plugins/AirSim/Source/AirSim.Build.cs
@@ -47,18 +47,18 @@ public class AirSim : ModuleRules
         switch (mode)
         {
             case CompileMode.HeaderOnlyNoRpc:
-                Definitions.Add("AIRLIB_HEADER_ONLY=1");
-                Definitions.Add("AIRLIB_NO_RPC=1");
+                PublicDefinitions.Add("AIRLIB_HEADER_ONLY=1");
+                PublicDefinitions.Add("AIRLIB_NO_RPC=1");
                 AddLibDependency("AirLib", Path.Combine(AirLibPath, "lib"), "AirLib", Target, false);
                 break;
             case CompileMode.HeaderOnlyWithRpc:
-                Definitions.Add("AIRLIB_HEADER_ONLY=1");
+                PublicDefinitions.Add("AIRLIB_HEADER_ONLY=1");
                 AddLibDependency("AirLib", Path.Combine(AirLibPath, "lib"), "AirLib", Target, false);
                 LoadAirSimDependency(Target, "rpclib", "rpc");
                 break;
             case CompileMode.CppCompileNoRpc:
                 LoadAirSimDependency(Target, "MavLinkCom", "MavLinkCom");
-                Definitions.Add("AIRLIB_NO_RPC=1");
+                PublicDefinitions.Add("AIRLIB_NO_RPC=1");
                 break;
             case CompileMode.CppCompileWithRpc:
                 LoadAirSimDependency(Target, "rpclib", "rpc");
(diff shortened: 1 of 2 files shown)


edowson:
I've submitted a fix for issue #1657.

Since the Flush On Draw function has been removed from viewports and frame buffers, I've removed the calls to those functions and the flush_on_draw_count_ member variable.

There is an impact in frame rate when the sensor data display is turned on. Here are the statistics on an 
Ubuntu-16.04.2 LTS host with nvidia-418 driver and a Titan-V GPU connected to a 34" display with the UE4Editor running in windowed mode at a resolution of 3037 × 1281:
- no display: 120 fps
- single display (either camera, depth or segmentation ): ~90 fps
- two displays: ~67 fps
- three displays: ~58 fps

I'll re-run these tests with UE-4.18.2 and report those results here, to see the performance drop with the hack enabled.

In any-case, that hack can no longer be applied to UE-4.22.

edowson:
Here is the procedure for setting up and configuring nvidia-418 driver support for vulkan libraries. This works for both the host and from within docker containers.

### Host system setup

Step 01: Install vulcan libraries and utilities
```bash
sudo apt install -y libvulkan1 vulkan-utils
```

Step 02: Edit the file `/usr/share/vulkan/icd.d/nvidia_icd.json` and rename `libGL.so.1` to `libGLX_nvidia.so.0`:
```json
{
    "file_format_version" : "1.0.0",
    "ICD": {
        "library_path": "libGLX_nvidia.so.0",
        "api_version" : "1.1.95"
    }
}
```

Step 03: Test Vulkan Rendering

```
vulkaninfo
vulkan-smoketest
```

### Docker container setup

Script the following into a `RUN` command in a `Dockerfile` so that you download the nvidia drivers, extract the *.run package and manually extract the libnvidia-glvkspirv.so.418.56 file to /usr/lib/x86_64-linux-gnu/, inside the docker image.

```bash
wget http://us.download.nvidia.com/XFree86/Linux-x86_64/418.56/NVIDIA-Linux-x86_64-418.56.run
NVIDIA-Linux-x86_64-418.56.run --extract-only
cp NVIDIA-Linux-x86_64-418.56/libnvidia-glvkspirv.so.418.56 /usr/lib/x86_64-linux-gnu/
# remove downloaded files
rm NVIDIA-Linux-x86_64-418.56.run
rm -Rf NVIDIA-Linux-x86_64-418.56
```

edowson:
Here are the performance figures using UE-4.18.3 with the render hack. 
no display: 120 fps
single display (either camera, depth or segmentation ): ~95 fps
two displays: ~70 fps
three displays: ~55 fps

I think we don't need the hack for UE4.22.

Unreal/Plugins/AirSim/Source/AirSim.Build.cs
```diff
@@ -47,18 +47,18 @@ private void SetupCompileMode(CompileMode mode, ReadOnlyTargetRules Target)
         switch (mode)
         {
             case CompileMode.HeaderOnlyNoRpc:
-                Definitions.Add("AIRLIB_HEADER_ONLY=1");
-                Definitions.Add("AIRLIB_NO_RPC=1");
+                PublicDefinitions.Add("AIRLIB_HEADER_ONLY=1");
```
madratman: @sytelus this is a breaking change for unreal 4.18 on windows. this PR should have been merged in a new 4.22 branch

madratman: in 4.19, this was changed to a warning. 
however, it throws an error in 4.18

edowson: I can submit a patch surrounding that code with a pre-processor check for UE_4_19_OR_LATER to handle the breakage on UE4 versions 18 and lower.

CI reported no build issue and since 4.18 is the current officially supported version, I assumed it was being tested against UE4.18.3. Unless CI is only setup to check and build on linux and not windows!

edowson: I've create a PR #1897 that should resolve the breakage for 4.18 on windows and handle versions 4.19 and higher.

https://github.com/Microsoft/AirSim/pull/1897/commits/b4464124d0ef288a151a5eb497db35853f1ce281

PR #1890, opened 2019-04-16
add a hexacopter model

lovettchris:
Here's some changes that make it easier to create a hexacopter model that works.

See demo video here: http://lovettsoftware.com/videos/px4hex.mp4


Unreal/Plugins/AirSim/Source/Vehicles/Multirotor/FlyingPawn.h
```diff
@@ -19,6 +19,9 @@ class AIRSIM_API AFlyingPawn : public APawn
     UPROPERTY(EditAnywhere, BlueprintReadWrite, Category = "Debugging")
     float RotatorFactor = 1.0f;
 
+    UPROPERTY(EditAnywhere, BlueprintReadWrite, Category = "Debugging")
+    int rotor_count = 4;
```
sytelus: Where does this gets set to 6?

Unreal/Plugins/AirSim/Source/Vehicles/Multirotor/FlyingPawn.cpp
```diff
@@ -3,8 +3,11 @@
 #include "AirBlueprintLib.h"
```
sytelus: We should copyright on DJI model. I think Jim had mentioned that we are allowed to use any DJI models in AirSim but would be good to confirm. Also are these uasset files saved in 4.18? Otherwise they might cause errors for other users.

sytelus (changes requested):
PR looks good. Just few questions...

Unreal/Plugins/AirSim/Source/Vehicles/Multirotor/FlyingPawn.h
lovettchris: It's a UPROPERTY so you can set it in the blueprint for hexacopter (BP_FlyingPawnHex).

Unreal/Plugins/AirSim/Source/Vehicles/Multirotor/FlyingPawn.cpp
lovettchris: The model came from [Udeshaya Wadhwa on grabcad.com](https://grabcad.com/library/dji-s900-hex-rotor-drone-1), and not from DJI.

MahdiulChowdhury:
> Here's some changes that make it easier to create a hexacopter model that works.
> 
> See demo video here: http://lovettsoftware.com/videos/px4hex.mp4

In the video, you showed the loaded model with the body and the propeller. Could you kindly show the steps for adding the propeller to body and rotating the propellers? perhaps a video of the steps would really help...
Thanks

Unreal/Plugins/AirSim/Source/Vehicles/Multirotor/FlyingPawn.h
mchowdhury-tft: UPROPERTY , are you referring to Blueprint editor? I did following steps: 
1. "Model": "Hexacopter"  //I added this in setting.json file 
2.  SimpleFlightQuadXParams.hpp I added  initializeRotorHexX(params.rotor_poses, params.rotor_count, arm_lengths.data(), rotor_z);
3. I add two more prop and rotation in the blueprint editor. 

Right now my problem is, sometimes 6 propellers, sometimes 4 and sometimes 5. I don't understand why it's doing this. 

Any idea what should I do ? please let me know...

latif11252:
the demo video is not working

Issue #1818, opened 2019-03-01
Improve image capture speed


tiax615:
Hi community，I think I may have found a way to improve the speed of image capture.
Modify the code in PIPCamera.cpp, updateCaptureComponentSetting()，
change
`render_target->InitAutoFormat(setting.width, setting.height);`
to
`render_target->InitCustomFormat(setting.width, setting.height, PF_B8G8R8A8, true);`

AirSim use SceneCaptureComponent2D SceneCapture mode is LDR, using RGBA8 is enough, may not need to use RGBA16f, which is supporting HDR mode. The Format of RenderTarget initialized by InitAutoFormat() is RGBA16f, which takes up resources. You can customize the Format by using InitCustomFormat(). So screenshots might become more efficient.

Thank you for reading！

tiax615:
By the way, I used the Victory plugin by Rama to take screenshots and the frame rate was up to 20fps!
I think it's also possible to write methods in C++ code without blueprints
![_20190301200638](https://user-images.githubusercontent.com/39846986/53637321-ca968d00-3c5d-11e9-94f4-a3d28128180c.png)

pengjili:
@tiax615 Really appreciate your sharing. I've tried your first way(didn't try 2nd yet), it did save a lot of time. But can I ask what's your image's resolution? Mine is 1280x800, and even using viewMode: "NoDisplay",  but it still costs about 250ms. I'd also like to put my frame rate up to 20fps, is there any other way to reduce time on saving images? Thanks so much again

tiax615:
> @tiax615 Really appreciate your sharing. I've tried your first way(didn't try 2nd yet), it did save a lot of time. But can I ask what's your image's resolution? Mine is 1280x800, and even using viewMode: "NoDisplay", but it still costs about 250ms each frame. I'd also like to put my frame rate up to 20fps, is there any other way to reduce time on saving images? Thanks so much again

Hi, My image resolution is 1280x720. And I'm looking for a faster way to do it, like video live. OBS?

PR #1910, opened 2019-04-24
Improve simgetimages speed 

madratman:
~(Don't merge until we find the right format for depth)~

This PR improves speed of simgetimages by :
 - changing pixel format to RGBA8 in PIPCamera.cpp following suggestion https://github.com/Microsoft/AirSim/issues/1818, and
- only copying RGB in RenderRequest.cpp in these lines https://github.com/Microsoft/AirSim/blob/master/Unreal/Plugins/AirSim/Source/RenderRequest.cpp#L109-L118
I've made changes from RGBA->RGB in the client side code I could find. 
An added benefit of RGBA -> RGB is that the image responses from the simgetimages() API match the unreal subwindow previews nicely. This has been a complaint by our users in the past, in that the simgetimage() responses tend to be much darker than the preview in the subwindow

Let's talk about RGB first:
- Behind the scenes, Unreal initializes the render target to RGBA16f as can be seen in the engine code [here](https://github.com/EpicGames/UnrealEngine/blob/7d9919ac7bfd80b7483012eab342cb427d60e8c9/Engine/Source/Runtime/Engine/Private/TextureRenderTarget2D.cpp#L27) 

- The suggestion in #1818 works well in that we can initialize it to `EPixelFormat::PF_B8G8R8A8`.   
There's also an `EPixelFormat::PF_R8G8B8A8_UINT`, but trying that makes the engine crash with assert failure in that the requested render target format is not supported.   
Interestingly, `PF_R8G8B8A8` leads to the same assert failure. 
There's no `PF_B8G8R8A8_UINT`, unfortunately.    
Side note: these names are misleading in that BGR is actually RGB - https://answers.unrealengine.com/questions/69615/are-epixelformats-pf-b8g8r8a8-and-pf-r8g8b8a8-flip.html   
Why engine crashes though, is something I am not sure about.    
- Q. We should try to get int8 images / textures? We can hopefully change the scene capture to get rgba int (not float) and then use `PF_R8G8B8A8_UINT`? Although I am not sure if this will help coz the GPU would have the images as floats only. 

Moving on to depth:
- Now, there should be a similar optimization possible for render targets for depth images. As far as I understand, it should be `EPixelFormat::PF_R16F` as alluded by this line https://github.com/Microsoft/AirSim/blob/master/Unreal/Plugins/AirSim/Source/RenderRequest.cpp#L127. However doing so, again leads to engine crashing with assert failure in that the requested render target format is not supported . 
Not completely sure why is that++  
My hunch is this happens coz the capture source is being initialized to [`ESceneCaptureSource::SCS_FinalColorLDR`](https://api.unrealengine.com/INT/API/Runtime/Engine/Engine/ESceneCaptureSource/index.html) for all image types at this line https://github.com/Microsoft/AirSim/blob/master/Unreal/Plugins/AirSim/Source/PIPCamera.cpp#L66. 

- `ESceneCaptureSource::SCS_FinalColorLDR` doesn't seem efficient for depth images. I tried changing it to `ESceneCaptureSource::SCS_SceneDepth`, similarly there's a `ESceneCaptureSource::SCS_Normal` which should be used for the normal image?
Unfortunately, engine crashes yet again with those two options. 

- Anyhow, for now, I have added a placeholder TMap for image_types to pixel_format types. The map is only used for all non depth image types (aka `image_type==0 || image_type==5 || image_type==6 || image_type==7`). 
for depth images, I still call `initautoformat` for now. 

P.S. *Technically*, we should use [`ETextureRenderTargetFormat`](https://api.unrealengine.com/INT/API/Runtime/Engine/Engine/ETextureRenderTargetFormat/index.html), which is the set of EPixelFormats exposed to rendertarget. 
However, the mapping is fairly intuitive from the variable names as can be seen in the engine code [in these lines](https://github.com/EpicGames/UnrealEngine/blob/7d9919ac7bfd80b7483012eab342cb427d60e8c9/Engine/Source/Runtime/Engine/Classes/Engine/TextureRenderTarget2D.h#L44-#L60), so, i am sticking with pixelformat

### DIFF
--- a/PythonClient/airsim/pfm.py
+++ b/PythonClient/airsim/pfm.py
@@ -49,7 +49,6 @@ def read_pfm(file):
 
     data = np.reshape(data, shape)
     # DEY: I don't know why this was there.
-    #data = np.flipud(data)
     file.close()
     
     return data, scale
@@ -63,8 +62,6 @@ def write_pfm(file, image, scale=1):
 
     if image.dtype.name != 'float32':
         raise Exception('Image dtype must be float32.')
-
-    image = np.flipud(image)
 
     if len(image.shape) == 3 and image.shape[2] == 3: # color image
         color = True
--- a/PythonClient/airsim/utils.py
+++ b/PythonClient/airsim/utils.py
@@ -157,7 +157,6 @@ def read_pfm(file):
 
     data = np.reshape(data, shape)
     # DEY: I don't know why this was there.
-    #data = np.flipud(data)
     file.close()
     
     return data, scale
@@ -172,11 +171,9 @@ def write_pfm(file, image, scale=1):
     if image.dtype.name != 'float32':
         raise Exception('Image dtype must be float32.')
 
-    image = np.flipud(image)
-
     if len(image.shape) == 3 and image.shape[2] == 3: # color image
         color = True
-    elif len(image.shape) == 2 or len(image.shape) == 3 and image.shape[2] == 1: # greyscale
+    elif len(image.shape) == 2 or len(image.shape) == 3 and image.shape[2] == 1: # grayscale
         color = False
     else:
         raise Exception('Image must have H x W x 3, H x W x 1 or H x W dimensions.')
@@ -206,9 +203,9 @@ def write_png(filename, image):
     height = image.shape[0]
 
     # reverse the vertical line order and add null bytes at the start
-    width_byte_4 = width * 4
-    raw_data = b''.join(b'\x00' + buf[span:span + width_byte_4]
-                        for span in range((height - 1) * width_byte_4, -1, - width_byte_4))
+    width_byte_3 = width * 3
+    raw_data = b''.join(b'\x00' + buf[span:span + width_byte_3]
+                        for span in range((height - 1) * width_byte_3, -1, - width_byte_3))
 
     def png_pack(png_tag, data):
         chunk_head = png_tag + data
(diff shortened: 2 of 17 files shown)


madratman:
update:
[This call to FTextureRenderTargetResource::IsSupportedFormat() was the assertion failure](https://github.com/EpicGames/UnrealEngine/blob/7d9919ac7bfd80b7483012eab342cb427d60e8c9/Engine/Source/Runtime/Engine/Private/TextureRenderTarget2D.cpp#L86), 

FTextureRenderTargetResource::IsSupportedFormat() can be seen in [these lines](https://github.com/EpicGames/UnrealEngine/blob/7d9919ac7bfd80b7483012eab342cb427d60e8c9/Engine/Source/Runtime/Engine/Private/TextureRenderTarget.cpp#L68-L80), which means only these pixel formats are supported:
```
PF_B8G8R8A8:
PF_A16B16G16R16:
PF_FloatRGB:
PF_FloatRGBA: // for exporting materials to .obj/.mtl
PF_A2B10G10R10: //Pixel inspector for normal buffer
PF_DepthStencil: //Pixel inspector for depth and stencil buffer
```

madratman:
update:
i tried `PF_DepthStencil` with `ESceneCaptureSource::SCS_SceneDepth` in vain. With that we have a weird memory writing error in the unreal stack. I guess as part of this in a shader function, which I haven't looked a lot into. 

Interestingly, `ESceneCaptureSource::SCS_SceneDepth` works fine with `PF_B8G8R8A8`, but the scaling of the depth image blows up a lot, so not messing with it anymore. 

Now, this PR is good to be merged from my end. For depth images (image_type = 2/3/4), we'll have placeholder formats in `APIPCamera::image_type_to_pixel_format_map_`, but they are not being used as can be seen in `APIPCamera::setupCameraFromSettings()`

fwiw, I found a similar discussion in carla here https://github.com/carla-simulator/carla/issues/750 with unfortunately similar conclusions

PR #1925, opened 2019-04-30
Image api update, bug fixes, computer vision scripts cleanup

ironclownfish:
- added simListSceneObjects to the api
- simGetCamInfo now returns the correct projection matrix
- Requesting CamInfo from a camera name that doesn't exist no longer crashes.

### DIFF
--- a/PythonClient/computer_vision/capture_ir_segmentation.py
+++ b/PythonClient/computer_vision/capture_ir_segmentation.py
@@ -7,6 +7,74 @@ import random
 import glob
 from airsim import *
 
+def rotation_matrix_from_angles(pry):
+    pitch = pry[0]
+    roll = pry[1]
+    yaw = pry[2]
+    sy = numpy.sin(yaw)
+    cy = numpy.cos(yaw)
+    sp = numpy.sin(pitch)
+    cp = numpy.cos(pitch)
+    sr = numpy.sin(roll)
+    cr = numpy.cos(roll)
+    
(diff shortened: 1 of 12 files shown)


PythonClient/computer_vision/capture_ir_segmentation.py
```diff
@@ -142,40 +206,52 @@ def main(client,
...
+    hippoList = client.simListSceneObjects('.*?Hippo.*?')
+    
+    objectList = elephantList
+    
     #Sample calls to main, varying camera angle and altitude.
     #straight down, 400ft
     main(client, 
          objectList, 
-         folder=r'auto\winter\400ft\down') 
+         irFolder=r'C:\\Users\\v-nigyde\\Documents\\SCREENSHOTS\\')#auto\winter\400ft\down')
```
madratman: can you make this a variable and initialize it on top of the file

ironclownfish: It seems like main() is intended to be called arbitrarily many times, with a different file path(s) each time. I could write something that generates the path, since it seems to follow the pattern: 'auto\season\elevation_ft\direction'

madratman:
@ironclownfish looks good to merge, but can you please squash the first 3 commits and the last 2 as well, for a cleaner history?
The first two have the same diff, and the 3rd is a merge commit. 
The last is a single line change.

PR #1933, opened 2019-05-06
Support for setting origin geopoint via command-line args

KerryMoffittRtn:
so we can have separate homes per map


KerryMoffittRtn:
This allows for a launcher per map, with independent home LLA

madratman:
Closing this as this will lead to inconsistency in terms of which settings are settable via command line args, and which aren't. I don't think a lot of users want / need this feature. 
Unless all settings are ported as command line options, we shouldn't be doing this.

PR #1935, opened 2019-05-06
Show main camera LLA in debug report text

KerryMoffittRtn:


madratman:
I've a hard time understanding the motivation behind, and the need for this PR in the repository. 
Why do we need to have this in the code? It's quite specific : we're adding the main camera's lat-long-alt in the debug report, but why? Was this feature requested by anyone?

PR #1937, opened 2019-05-06
Spawn vehicles via rpc

KerryMoffittRtn:


hos-b:
@KerryMoffittRtn Hi. Since apparently this PR is not getting merged, could you help me with my problem ? 
I'm trying to add some cars on the fly but I get the following error : 
`rpclib: function 'enableApiControl' (called with 2 arg(s)) threw an exception. The exception contained this information: Vehicle API for 'agent-0' is not available. This could either because this is simulation-only API or this vehicle does not exist.`
If I don't use `enableApiControl`, I won't get an error but it also won't spawn anything. I get the same error when I run HelloSpawnedDrones.

Thanks

jinkman:
The generated vehicle cannot receive commands

AirLib/include/api/RpcLibClientBase.hpp
```diff
@@ -80,6 +80,8 @@ class RpcLibClientBase {
     vector<ImageCaptureBase::ImageResponse> simGetImages(vector<ImageCaptureBase::ImageRequest> request, const std::string& vehicle_name = "");
     vector<uint8_t> simGetImage(const std::string& camera_name, ImageCaptureBase::ImageType type, const std::string& vehicle_name = "");
 
+	bool simAddVehicle(const std::string& vehicle_name, const std::string& vehicle_type, const std::string& pawn_path, float north, float east, float down);
```
jinkman: client.simAddVehicle("My", "simpleflight", "", 0, 0, 0);
	client.enableApiControl(true, "My");
	client.armDisarm(true, "My");
	client.takeoffAsync(1.0, "My")->waitOnLastTask();
	client.moveToPositionAsync(10, 10, -10, 10, 10, DrivetrainType::ForwardOnly, YawMode(false, 0.0), -1.0, 1.0, "My")->waitOnLastTask();
Don't know why these commands go wrong

jinkman: The error message looks like this：
LogTemp: Error: Exception occurred while updating world: PositionController does not support yaw axis i.e. 3
LogTemp: Error: Exception occurred while updating world: reset() must be called first before update()
LogTemp: Error: Exception occurred while updating world: reset() must be called first before update()
LogTemp: Error: Exception occurred while updating world: reset() must be called first before update()

rajat2004:
@KerryMoffittRtn Thanks for this! It's a very interesting PR, I've opened a new PR https://github.com/microsoft/AirSim/pull/2390 which is rebased on master, will be adding some more things on that

madratman:
Closing this in favor of https://github.com/microsoft/AirSim/pull/2390 
I am slightly wary of this feature, and I am not sure how many people actually want it. 
Spawning pure physics bodies at run time could be interesting in case of RL algorithms in the context of (dynamic) obstacle avoidance (a la papers throwing objects at robots learning to walk in sim.).

PR #1941, opened 2019-05-07
Fixed some Steering Wheel & Config Bug.

jasperxu:
1. Fixed Steering Wheel's Buttons Bug. 
When you press any key on the steering wheel, all keys are recognized as true.
Bit operations are using & instead of &&.

2. Fixed Steering Wheel is not Work.
In file CarPawnSimApi.cpp Line 90 we used the function getRCData() and checked the value of rc_data.is_initialized. However, the value of rc_data.is_initialized is not set in the function getRCData(). Causes rc_data.is_initialized to be in false all the time. So the value of rc_data.is_initialized is always false, and the steering wheel cannot be enabled.

3. Fixed RC failed to read in the configuration file.
At Line 546. The code is repeated. So we mast set RC like this.
```
"RC": {
    "RC": {
        "RemoteControlID": 0,
        "AllowAPIWhenDisconnected": false
    }
}
```

### DIFF
--- a/AirLib/include/common/AirSimSettings.hpp
+++ b/AirLib/include/common/AirSimSettings.hpp
@@ -708,10 +708,7 @@ private:
         vehicle_setting->is_fpv_vehicle = settings_json.getBool("IsFpvVehicle",
             vehicle_setting->is_fpv_vehicle);
 
-        Settings rc_json;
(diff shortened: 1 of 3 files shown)


madratman:
Thanks for the fixes @jasperxu

Issue #1760, opened 2019-02-02
CarClient Linear Velocity


ElliotHYLee:
From version 1.2.0, for the release versions, the car mode's linear velocity is not [0,0,0] at beginning when the car is still. It seems the velocities differ by the environment package. However, the MultirotorClient's linear velocity are all [0,0,0] at the beginning.

What's causing this? By simply subtracting the initial velocity, can this be usable?

![capture](https://user-images.githubusercontent.com/7537835/52158146-4a333b00-2663-11e9-88f9-609f7188f7d7.PNG)

The setting is:
{
  "SeeDocsAt": "https://github.com/Microsoft/AirSim/blob/master/docs/settings.md",
  "SettingsVersion": 1.2,
  
  "CameraDefaults": {
	"CaptureSettings": [
	  {
		"ImageType": 0,
		"Width": 720,
		"Height": 360,
		"TargetGamma": 3.0
	  }
	]
  }
}

dawgster:
I'm having the same issue. Is this a bug or a feature I don't understand yet? In the blocks environment, a value for linear_velocity.x_val of around -59 shows when in standstill.

ironclownfish:
After PR 2705 is merged, you can safely do this to achieve true pause.
![image](https://user-images.githubusercontent.com/6217407/82500575-749f4680-9aa8-11ea-8f81-cea30af669e3.png)

PR #1960, opened 2019-05-14
added offset option so velocity vectors don't get bungled with car sim (#1760)

msb336:
Car local_ned_offset_ is a non-zero vector, so it messes up the velocity vector when converting from Unreal `FVector` to airsim `Vector3r`. Commit adds an option to discard position offset when calculating velocity.
[see here](https://github.com/Microsoft/AirSim/issues/1760)

### DIFF
--- a/Unreal/Plugins/AirSim/Source/NedTransform.cpp
+++ b/Unreal/Plugins/AirSim/Source/NedTransform.cpp
@@ -26,14 +26,15 @@ NedTransform::NedTransform(const AActor* pivot, const FTransform& global_transfo
         local_ned_offset_ = FVector::ZeroVector;
 }
 
-NedTransform::Vector3r NedTransform::toLocalNed(const FVector& position) const
+NedTransform::Vector3r NedTransform::toLocalNed(const FVector& position, bool offset=true) const
 {
-    return NedTransform::toVector3r(position - local_ned_offset_,
+
+    return NedTransform::toVector3r((offset ? position - local_ned_offset_ : position),
         1 / world_to_meters_, true);
 }
-NedTransform::Vector3r NedTransform::toGlobalNed(const FVector& position) const
+NedTransform::Vector3r NedTransform::toGlobalNed(const FVector& position, bool offset=true) const
 {
-    return NedTransform::toVector3r(position - global_transform_.GetLocation(),
+    return NedTransform::toVector3r((offset ? position - global_transform_.GetLocation() : position),
         1 / world_to_meters_, true);
 }
 NedTransform::Quaternionr NedTransform::toNed(const FQuat& q) const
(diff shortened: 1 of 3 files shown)


madratman:
I think this will lead to unreadable code, as this boolean parameter can only be understood by us for now, and perhaps we will forget later as well why exactly it was added. 
In general, functions taking a bool to do something can be messy in the long term. 
I'd rather just have another transform function which doesn't add an offset 

I think it's better to rename the original function which adds an offset to `toLocalNedPosition(const FVector& position)`, and add a new one `toLocalNedVelocity(const FVector& velocity)` which doesn't add the offset. 

On second thought, I am guessing `toLocalNedVelocity` would be used for transforming accelerations as well, so maybe a more general name is needed, perhaps `toLocalNedWithOffset` `toLocalNedVelocityWithoutOffset`

rajat2004:
I have been working on adding support for Rover vehicle of Ardupilot using the Car vehicle and was facing the problem of sensor inconsistency, the PR fixed it!

@msb336 Thanks a lot for this!!
I'll be opening a PR for the things I did but hopefully, this gets merged soon

Unreal/Plugins/AirSim/Source/NedTransform.cpp
```diff
@@ -57,13 +58,13 @@ float NedTransform::fromNed(float length) const
...
+FVector NedTransform::fromLocalNed(const NedTransform::Vector3r& position, bool offset=true) const
 {
-    return NedTransform::toFVector(position, world_to_meters_, true) + local_ned_offset_;
+    return (offset ? NedTransform::toFVector(position, world_to_meters_, true) + local_ned_offset_ : NedTransform::toFVector(position, world_to_meters_, true));
 }
-FVector NedTransform::fromGlobalNed(const NedTransform::Vector3r& position) const
+FVector NedTransform::fromGlobalNed(const NedTransform::Vector3r& position, bool offset=true) const
 {
-    return NedTransform::toFVector(position, world_to_meters_, true) + global_transform_.GetLocation();
+    return (offset ? NedTransform::toFVector(position, world_to_meters_, true) + global_transform_.GetLocation(), NedTransform::toFVector(position, world_to_meters_, true));
```
rajat2004: Should be 
```
return (offset ? NedTransform::toFVector(position, world_to_meters_, true) + global_transform_.GetLocation() : NedTransform::toFVector(position, world_to_meters_, true));
```

colon instead of comma in the seperator

madratman: In any case that function is not used in the commit and is an extraneous addition. 
The only function used here was `toLocalNed` with the second param as false. 

Closing this PR in favor of https://github.com/microsoft/AirSim/pull/2481, reasoning is same as my old comment on this PR. adding a bool param is unintuitive and hurts readability.  Please rebase your ardurover one on master once 2481 is in.

PR #1970, opened 2019-05-17
Fixed low-repro reset bug.

ironclownfish:
Changed virtual void UpdatableObject::reset() to void UpdatableObject::reset(), and added a new function: UpdatableObject::resetImplementation(), which is called by UpdatableObject::reset().

reset() overrides always had to call super to get the base class functionality. Now the base class functionality is inherent in the reset() call. This allows the addition of a reset_in_progress flag in a single place, rather than in every reset() override. This new flag is used to make reset() reentrant, which fixes the bug.

Also removed all extraneous calls to super from reset() overrides (which are now resetImplementation() overrides). Moved other UpdatableObject reset/update failure cases into a virtual function: failResetUpdateOrdering(), so they can be disabled by overriding the function. This is a little cleaner, and the old way of disabling didn't work with the new reset stuff.

### DIFF
--- a/AirLib/include/physics/World.hpp
+++ b/AirLib/include/physics/World.hpp
@@ -27,9 +27,9 @@ public:
 
     //override updatable interface so we can synchronize physics engine
     //*** Start: UpdatableState implementation ***//
-    virtual void reset() override
+    virtual void resetImplementation() override
     {
-        UpdatableContainer::reset();
+        UpdatableContainer::resetImplementation();
         
         if (physics_engine_)
             physics_engine_->reset();
--- a/AirLib/src/api/RpcLibServerBase.cpp
+++ b/AirLib/src/api/RpcLibServerBase.cpp
@@ -162,6 +162,31 @@ RpcLibServerBase::RpcLibServerBase(ApiProvider* api_provider, const std::string&
         return RpcLibAdapatorsBase::LidarData(lidar_data);
     });
 
+    pimpl_->server.bind("getImuData", [&](const std::string& imu_name, const std::string& vehicle_name) -> RpcLibAdapatorsBase::ImuData {
+        const auto& imu_data = getVehicleApi(vehicle_name)->getImuData(imu_name);
+        return RpcLibAdapatorsBase::ImuData(imu_data);
+    });
+
+    pimpl_->server.bind("getBarometerData", [&](const std::string& barometer_name, const std::string& vehicle_name) -> RpcLibAdapatorsBase::BarometerData {
+        const auto& barometer_data = getVehicleApi(vehicle_name)->getBarometerData(barometer_name);
+        return RpcLibAdapatorsBase::BarometerData(barometer_data);
+    });
+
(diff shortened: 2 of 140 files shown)


madratman:
this has some commits from your previous PR https://github.com/microsoft/AirSim/pull/1959 / https://github.com/microsoft/AirSim/pull/1925
As this PR is on the same AfricaTrackingFeature branch.. Can you rebase your local branch onto current upstream master?

ironclownfish:
> this has some commits from your previous PR #1959 / #1925, probably as this PR is on the same AfricaTrackingFeature branch.. Can you rebase your local branch onto current upstream master?

Yeah I will sort that out. Gotta be less of a git n00b.

AirLib/include/physics/World.hpp
```diff
@@ -27,9 +27,9 @@ class World : public UpdatableContainer<UpdatableObject*> {
 
     //override updatable interface so we can synchronize physics engine
     //*** Start: UpdatableState implementation ***//
-    virtual void reset() override
+    virtual void resetImplementation() override
     {
-        UpdatableContainer::reset();
+        UpdatableContainer::resetImplementation();
```
madratman: this does not look right to me, or maybe I am misunderstanding it
this looks like the anti-pattern itself

AirLib/src/api/RpcLibServerBase.cpp
```diff
@@ -132,13 +132,20 @@ RpcLibServerBase::RpcLibServerBase(ApiProvider* api_provider, const std::string&
...
+			return;
+
+		//Reset
+		resetInProgress = true;
         auto* sim_world_api = getWorldSimApi();
         if (sim_world_api)
             sim_world_api->reset();
         else
             getVehicleApi("")->reset();
+		resetInProgress = false;
```
madratman: is this supposed to be scoped
don't quite understand why you need resetInProgress here

Unity/AirLibWrapper/AirsimWrapper/CMakeLists.txt
```diff
@@ -33,7 +33,7 @@ BuildAirsimWrapper()
 
 
 ###################### Link source files to library ######################################
-if (${MACOSX})
+if (${APPLE})
```
madratman: umm, no

Unity/AirLibWrapper/AirsimWrapper/Source/WorldSimApi.h
```diff
@@ -13,7 +13,7 @@ class WorldSimApi : public msr::airlib::WorldSimApiBase
 	WorldSimApi(SimModeBase* simmode, std::string vehicle_name);
 	virtual ~WorldSimApi();
 	virtual bool isPaused() const override;
-	virtual void reset() override;
```
madratman: worlsimapibase.hpp has reset defined as pure virtual still..

madratman:
@ironclownfish see my partial review above..
Due to the irrevocable git blunders on this PR, I am closing this, and making a new one from my fork.

AirLib/src/api/RpcLibServerBase.cpp
ironclownfish: This was a previous hack fix that I thought I removed. (Maybe did, but who can tell with this commit history :| )

PR #1976, opened 2019-05-21
Changed cmake/build.sh to point to correct rpclib directory

DTChuck:
The gcc build script in the cmake folder pointed to the incorrect location for rpclib. The rsync command would fail and the build script would error out. This fix points to the correct location for rpclib


madratman:
that script might get obviated with https://github.com/microsoft/AirSim/pull/2048, or already is not needed @lovettchris?
Also, seems like the intended goal of this PR was this file https://github.com/microsoft/AirSim/blob/master/cmake/gcc_build.sh, and not cmake/build.sh

all these cmake/*.sh seems like relics of the past to me which are not being used by anyone?
Can I just nuke them @lovettchris @sytelus ?

sytelus:
I agree. There shouldn't be two versions of build.sh. The one in cmake is horribly outdated anyway so better to nuke those files.

PR #2036, opened 2019-06-23
Update image_apis.md

rancheng:
correction on the color pattern channels and update the code to convert into the right format.


madratman:
The code is correct for uncompressed rgb image, aka `request = [airsim.ImageRequest("0", airsim.ImageType.Scene, False, False)]`. 
The comment above it is however misleading:  `# reshape array to 4 channel image array H X W X 4 ` should be  `# reshape array to 3 channel image array H X W X 3`
This change was made in PR https://github.com/microsoft/AirSim/pull/1910.

PR #2048, opened 2019-07-02
Merge PR 1934 and add some more fixes along the same lines

lovettchris:
I added "--gcc" option to setup.sh which bypasses install of LLVM 5 and "./build.sh --gcc" now works properly.  Fixed GCC compile issues involving std::sinf, std::cosf, etc.  And added a version check so cmake is only installed if user doesn't already have the right version.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -5,8 +5,9 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 pushd "$SCRIPT_DIR"  >/dev/null
 
 set -e
-# set -x
+set -x
 
+MIN_GCC_VERSION=6.0.0
 gccBuild=false
 # Parse command line arguments
 while [[ $# -gt 0 ]]
@@ -42,9 +43,21 @@ fi
 # set up paths of cc and cxx compiler
 if $gccBuild; then
     # variable for build output
-    build_dir=build_gcc_debug    
-    export CC="gcc-8"
-    export CXX="g++-8"
+    build_dir=build_gcc_debug
+    # gcc tools
+    gcc_ver=$(gcc --version 2>&1 | head -n1 | cut -d ' ' -f4 | awk '{print $NF}')
+    gcc_path=$(which cmake)
+    if [[ "$gcc_path" == "" ]] ; then
+        echo "ERROR: run setup.sh to install a good version of gcc."
+        exit 1
+    fi
+    if [[ $(python $SCRIPT_DIR/tools/version.py $gcc_ver $MIN_GCC_VERSION) ]]; then
+        export CC="gcc-6"
+        export CXX="g++-6"
+    else
+        export CC="gcc"
+        export CXX="g++"
+    fi
 else
     #check for correct verion of llvm
     if [[ ! -d "llvm-source-50" ]]; then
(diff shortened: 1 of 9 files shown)


build.sh
```diff
@@ -33,42 +29,73 @@ if [ ! -d "./external/rpclib/rpclib-2.2.1" ]; then
...
+    # variable for build output
+    build_dir=build_gcc_debug
+    # gcc tools
+    gcc_ver=$(gcc --version 2>&1 | head -n1 | cut -d ' ' -f4 | awk '{print $NF}')
+    gcc_path=$(which cmake)
+    if [[ "$gcc_path" == "" ]] ; then
+        echo "ERROR: run setup.sh to install a good version of gcc."
+        exit 1
+    fi
+    if [[ $(python $SCRIPT_DIR/tools/version.py $gcc_ver $MIN_GCC_VERSION) ]]; then
```
madratman: I don't think using python for just comparing versions is a good idea
why not just do it all in shell https://unix.stackexchange.com/a/285928
also, I don't agree with enforcing python >3.4 for build process, while our APIs still work in 2.7

setup.sh
```diff
@@ -26,57 +88,50 @@ if [ "$(uname)" == "Darwin" ]; then # osx
...
+    if [[ "$cmake_path" == "" ]] ; then
+        cmake_ver=0
+    fi
 
-if [ "$(uname)" == "Darwin" ]; then
-    CMAKE="$(greadlink -f cmake_build/bin/cmake)"
-else
-    CMAKE="$(readlink -f cmake_build/bin/cmake)"
+    #download cmake - v3.10.2 is not out of box in Ubuntu 16.04
+    if [[ $(python $SCRIPT_DIR/tools/version.py $cmake_ver $MIN_CMAKE_VERSION) ]]; then
```
madratman: same as above

madratman:
PR works beautifully for me in 18.04. thanks Chris. 
Let's get rid of the python script though

setup.sh
```diff
@@ -8,15 +8,77 @@ if [[ -d "llvm-source-39" ]]; then
...
+    --gcc)
+    gccBuild=true
+    shift # past argument
+    ;;
+esac
+done
+
+if $gccBuild; then
+    # gcc tools
+    gcc_ver=$(gcc --version 2>&1 | head -n1 | cut -d ' ' -f4 | awk '{print $NF}')
```
madratman: too complicated, me thinks
`gcc_ver=$(gcc -dumpfullversion)` gives you the same thing ;)

setup.sh
```diff
@@ -8,15 +8,77 @@ if [[ -d "llvm-source-39" ]]; then
...
+    gcc_path=$(which cmake)
+    if [[ "$gcc_path" == "" ]] ; then
+        gcc_ver=0
+    fi
+    if [[ $(python $SCRIPT_DIR/tools/version.py $gcc_ver $MIN_GCC_VERSION) ]]; then
```
madratman: replace with 
```   if [ "$(printf '%s\n' "$MIN_GCC_VERSION" "$gcc_ver" | sort -V | head -n1)" = "$MIN_GCC_VERSION" ]; then ```
and remove python script

madratman (changes requested):

madratman:
ok I just made those changes in my fork https://github.com/madratman/AirSim/commit/50f2e95599f260aa2a82ff2a5189a7bc43ee5054

doing some final checks

madratman:
Closing this and opening PR from my fork 
Your commits should remain, as I just build on them

build.sh
lovettchris: because I hate bash shell language :-) And python is more cross platform, but whatever, your call.

setup.sh
lovettchris: much nicer :-)

PR #2062, opened 2019-07-08
Prevent Environment from crashing when there is no sun UObject in the scene

msb336:
Indoor environments with no sun will crash when AirSim attempts to set the default sun rotation, as `sun_ = nullptr` in those situations.


Unreal/Plugins/AirSim/Source/SimMode/SimModeBase.cpp
```diff
@@ -160,7 +160,8 @@ void ASimModeBase::initializeTimeOfDay()
         UObjectProperty* sun_prop = Cast<UObjectProperty>(p);
         UObject* sun_obj = sun_prop->GetObjectPropertyValue_InContainer(sky_sphere_);
         sun_ = Cast<ADirectionalLight>(sun_obj);
-        default_sun_rotation_ = sun_->GetActorRotation();
+		if (sun_)
+			default_sun_rotation_ = sun_->GetActorRotation();
```
madratman: can we just follow the original indentation and then merge this?

madratman:
nm, fixed whitespace and merged in https://github.com/microsoft/AirSim/pull/2147 
thanks!

PR #2072, opened 2019-07-12
added getCarControls to client API

msb336:
getCarControls already existed on server side, I just exposed it via the client.


madratman:
lgtm

Issue #2142, opened 2019-08-19
Unreal Lidar incorrectly casting traces


WouterJansen:
**Steps to reproduce bug:**
 - Create a lidar sensor with an offset location from the vehicle origin
 - Turn on DrawDebugPoints for observing more easily the problem
 - Drive/fly the car/drone and make sure to rotate
 - Observe the incorrect lidar pointcloud being made

**Reason:**
`UnrealLidarSensor::shootLaser()`  in *UnrealLidarSensor.cpp* creates the starting location of the trace as 
`Vector3r start = lidar_pose.position + vehicle_pose.position;` (currently line 113)
 
If the lidar does not share the vehicle origin's X and Y position, this will start introducing an incorrect starting position for the trace as the vehicle starts rotating as the rotation of vehicle/lidar is completely ignored. 

**Fix:**
To fix one can change this line to
`Vector3r start = VectorMath::add(lidar_pose, vehicle_pose).position;`

ironclownfish:
Thanks, PR approved!

PR #2143, opened 2019-08-19
Fixed Unreal Lidar sensor incorrectly setting start position for the traces

WouterJansen:
Fixed Unreal Lidar sensor incorrectly setting start position for the traces by not accounting for the rotation of sensor/vehicle. Issue: #2142


WouterJansen:
Cleaned up my fork with reverts of commits I didn't think would end up in this pull request. Sorry, first time doing pull request from fork. Also re-updated to latest version of master.

PR #2168, opened 2019-08-29
fixed the mass param inconsistency

vinbo:
in the AdaptiveController.hpp, the mass is used in different value, from 0.916 to 0.9116. I think this might be a typo. I don't understand why the author use figures not variable name.


madratman:
Megrged manually in  #2298

PR #2173, opened 2019-09-02
Make the trace line persistent

songxiaocheng:
It is supposed to press "T" to toggle a persistent trace line but it is just shown as a little tail behind the drone. This commit fixes this problem.

### DIFF
--- a/Unreal/Plugins/AirSim/Source/PawnSimApi.cpp
+++ b/Unreal/Plugins/AirSim/Source/PawnSimApi.cpp
@@ -383,7 +383,7 @@ void PawnSimApi::plot(std::istream& s, FColor color, const Vector3r& offset)
         Vector3r current_point(x, y, z);
         current_point += offset;
         if (!VectorMath::hasNan(last_point)) {
-            UKismetSystemLibrary::DrawDebugLine(params_.pawn->GetWorld(), ned_transform_.fromLocalNed(last_point), ned_transform_.fromLocalNed(current_point), color, 0, 3.0F);
+            DrawDebugLine(params_.pawn->GetWorld(), ned_transform_.fromLocalNed(last_point), ned_transform_.fromLocalNed(current_point), color, true, -1.0F, 0, 3.0F);
         }
         last_point = current_point;
     }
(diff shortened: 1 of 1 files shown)


madratman:
thanks for this. 
I think it would be nice to have a call to `FlushPersistentDebugLines` as well, perhaps with another key, to clear the debug lines on demand https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Engine/Public/DrawDebugHelpers.h#L18

songxiaocheng:
> thanks for this.
> I think it would be nice to have a call to `FlushPersistentDebugLines` as well, perhaps with another key, to clear the debug lines on demand https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Engine/Public/DrawDebugHelpers.h#L18

songxiaocheng:
> thanks for this.
> I think it would be nice to have a call to `FlushPersistentDebugLines` as well, perhaps with another key, to clear the debug lines on demand https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Engine/Public/DrawDebugHelpers.h#L18

I believe this is already implemented in the previous version [here](https://github.com/microsoft/AirSim/blob/39fb3b0b051050f4bc9d770838808756b64ff5ab/Unreal/Plugins/AirSim/Source/PawnSimApi.cpp#L360) where `UKismetSystemLibrary::FlushPersistentDebugLines` works well which did the same thing with `FlushPersistentDebugLines`. Just press "T" to toggle the trace line on and off.

The [`DrawDebugLine`](https://docs.unrealengine.com/en-US/API/Runtime/Engine/DrawDebugLine/index.html) was replaced by [`UKismetSystemLibrary::DrawDebugLine`](https://docs.unrealengine.com/en-US/API/Runtime/Engine/Kismet/UKismetSystemLibrary/DrawDebugLine/index.html) in [this commit](https://github.com/microsoft/AirSim/commit/b5f9a0502d5d5f677247efa67aff09398db3b895) but it seems the latter function lacks a `bPersistentLines` option which causes the problem I mentioned above.

PR #2011, opened 2019-06-10
Lidarimprovements

ironclownfish:
Added an api function: simGetLidarSegmentation().
Returns the segmentation of each lidar point's collided object in the last lidar update. i.e. the segmentation IDs align with getLidarData().point_cloud

PR #2224, opened 2019-10-01
fixed type naming error created by PR #2011

msb336:
Changed type definition in UnrealLidarSensor::getPointCloud from vector to msr::airlib::vector


ironclownfish (approved):


PR #2243, opened 2019-10-10
PID derivate action update & fixed Python/C++ confusion

48cfu:


AirLib/include/vehicles/multirotor/firmwares/simple_flight/firmware/PidController.hpp
```diff
@@ -99,11 +99,9 @@ class PidController : public IUpdatable {
         if (dt > min_dt_) {
             integrator->update(dt, error, last_time_);
 
-            //To eliminate "derivative kick", we assume goal was approximately
-            //constant between successive calls. dE = dGoal - dInput = -dInput
-            float error_der = - (goal_ - last_goal_) / dt;
+            float error_der = - (error - last_error_) / dt;
```
saihv: Shouldn't the sign be reversed here?

saihv:
Thanks for submitting this!

@sytelus I was hoping you could shed some light on the idea and original implementation: dError seems to be referencing goal instead of input (per the [comment here](https://github.com/microsoft/AirSim/blob/master/AirLib/include/vehicles/multirotor/firmwares/simple_flight/firmware/PidController.hpp#L102)). Do we benefit from derivative kick assumption? 

For more info, I tested three variants with a sample case of moveByVelocity(): the current implementation, using dError = -dInput (according to the comment mentioned above), and this PR. The command was to move by a velocity of -20 m/s for 5 seconds at an altitude of 20 m. The results are seen below

Kd_0_2

Kd_1_0

For low values of Kd, all three approaches behave similarly, but with aggressive values, things change. The current implementation becomes very unstable at settling time, the derivative kick assumption (-dInput) results in large oscillations both at setpoint change and settling combined with a loss of altitude, and leaving it unchanged behaves better with some loss of altitude.

sytelus:
Hi @saihv, excellent analysis! Few comments...

1. The current implementation is inspired from this article series (which is implemented in Arduino PID library).
2. I realize that [comment ](https://github.com/microsoft/AirSim/blob/master/AirLib/include/vehicles/multirotor/firmwares/simple_flight/firmware/PidController.hpp#L102) actually doesn't match the current code! According to the commane, the code should be doing `dInput = measured_ - last_measured_` but instead it does `dGoal`. I remember that I'd lot of trouble making implementation as specified in that article work so I'm guessing I probably ended up changing the code diverting from the recommendation for derivative kick but didn't synchronized the comment.
3. For small kd, yes all 3 approach looks similar. This may be because kp takes over and for the experiment you mentioned, just using kp might work pretty well. However, typically, we cannot increase kp too much otherwise system becomes unstable and oscillate as error keeps jumping between -ve and positive around the goal. So typically kp is kept bit smaller and we rely on kd to increase the signal as error gap keeps increasing. In nutshell, kp and kd are tighly connected and its an art form to tune these two. If we increase kd enormously, kp must be reduced proportionally.
4. I'm assuming, left graph is current system, i.e., dE = -dGoal for kd=1.0. To me this looks like most stable if the goal is to maintain the altitude constant. The middle graph is for the comment in the code which probably shows the issue I was facing and may be why I decided to change code (while forgetting to change the comment). The PR is right one which looks less stable and is the regular approach without derivative kick compensation. If my this understanding is correct than correct code seems to be the best performer, no?
5. There is one more alternative which is to set dE = dGoal-dInput as per the above article. The article assumes goal is more or less constant so dGoal=0 which translates to dE = -dInput. This however isn't going to change the current graph where goal is indeed constant but for low value of kd, may be it works better?

One overall comment is that tuning this PID controller was a lot of effort and it was done manually in naive way. I think if you use more algorithmic approach to tune it then you might get much better values for kd, kp, ki but more importantly it might help you see issues more clearly about derivative kich when goal suddenly changes. Fortunately this is fairly easy to do because PID controller as well as quadrotor dynamics are implemented as library that can be used in simple console program to create PID test suit. Another direction is also to try out learned controllers as PIDs are fairly fragile when quadrotor parameters changes or during collisions or aggressive maneuverings like event.

saihv:
@sytelus

> I'm assuming, left graph is current system, i.e., dE = -dGoal for kd=1.0. To me this looks like most stable if the goal is to maintain the altitude constant. 

Actually, no, the true goal is to move left with a velocity of 20 m/s for 5 seconds and stop. I was just using the altitude as another marker of the stability of dynamics. I forgot to mention this part earlier: the graph was plotted over 30 seconds, so 5 seconds is 10000 time steps. After this, you can notice that the current implementation results in heavy oscillations, and takes a very long time to converge. 

Agree with point 3, and as to point 5: isn't dError = dGoal - dInput the same as what this PR proposes? I think the interesting part is that between dError = -dInput, and dError = dGoal - dInput, the only difference is when the goal setpoint changes. It seems like that difference is enough to affect the performance after the setpoint change significantly.

sytelus:
I'm still unclear about the setup. Do you expect to keep altitude stay stable? This is the most visible aspect in terms of stability. The roll may oscillate which is fine if altitude is still keeping same. For this PR implementation, I see quadrotor falling down by significant height after 10K steps. In real world, this behavior would be unacceptable (if the goal is to move without losing altitude).

PR #2255, opened 2019-10-15
Fix home position stability check

laurelkeys:
By doing `pos = self.home` both are pointing to the same address (which we can see by printing their [`id`](https://docs.python.org/3/library/functions.html#id) or seeing that checking if `pos is self.home` always returns True).

Thus, the `if` statement on line 57 is always going to the `else` branch:

https://github.com/microsoft/AirSim/blob/c6f0729006ecd6d792a750edc7e27021469c5649/PythonClient/multirotor/orbit.py#L56-L57

### DIFF
--- a/PythonClient/multirotor/orbit.py
+++ b/PythonClient/multirotor/orbit.py
@@ -53,10 +53,10 @@ class OrbitNavigator:
         start = time.time()
         count = 0
         while count < 100:
-            pos = self.home
+            pos = self.client.getMultirotorState().kinematics_estimated.position
             if abs(pos.z_val - self.home.z_val) > 1:                                
                 count = 0
-                self.home = pos
+                self.home = self.client.getMultirotorState().kinematics_estimated.position
                 if time.time() - start > 10:
                     print("Drone position is drifting, we are waiting for it to settle down...")
                     start = time


laurelkeys:
Also, I think that on [line 66](https://github.com/microsoft/AirSim/blob/master/PythonClient/multirotor/orbit.py#L66) it'd then be better to do `self.center = pos`, since it would avoid making another API call to the client, and once we've reached this line we already know that the values of `pos` and `self.home` are stable.

May I commit this change to the PR aswell?

PythonClient/multirotor/orbit.py
```diff
@@ -53,10 +53,10 @@ def __init__(self, radius = 2, altitude = 10, speed = 2, iterations = 1, center
         start = time.time()
         count = 0
         while count < 100:
-            pos = self.home
+            pos = self.client.getMultirotorState().kinematics_estimated.position
             if abs(pos.z_val - self.home.z_val) > 1:                                
                 count = 0
-                self.home = pos
+                self.home = self.client.getMultirotorState().kinematics_estimated.position
```
madratman: Hmm, wouldn't this reset the home position to current position. 
I think the `self.home` was set to save the original position (altitude) in the first place? 

I do agree with your other change on L56 (`pos = self.client.getMultirotorState().kinematics_estimated.position`)

laurelkeys: Indeed, but so does `self.home = pos`, doesn't it? From what I understand the aim is to keep resetting its value while the drone isn't stable (so we know the "actual" starting position once it is).

However, I think that keeping only the change on L56 should also do the trick. Should I update the PR to remove L59? Also, can I add `self.center = pos` to L66?

madratman: Yeah, that's what I meant. The L56 change should be enough.
Yes, let's use pos in L66 to save a server call

laurelkeys:
@madratman I'm closing this PR since I had deleted my original branch.
I have updated it in https://github.com/microsoft/AirSim/pull/2488.

PR #2256, opened 2019-10-15
added time dilation to pause entire world on simPause()

msb336:
This will close a lot of issues requesting simPause to pause all animations in the world along with the player controlled character.

### DIFF
--- a/Unreal/Plugins/AirSim/Source/WorldSimApi.cpp
+++ b/Unreal/Plugins/AirSim/Source/WorldSimApi.cpp
@@ -27,6 +27,8 @@ void WorldSimApi::reset()
 void WorldSimApi::pause(bool is_paused)
 {
     simmode_->pause(is_paused);
+	auto time_dilation = is_paused ? 0.0 : 1.0 ;
+	simmode_->GetWorld()->SetGlobalTimeDilation(time_dilation);
 }
 
 void WorldSimApi::continueForTime(double seconds)


madratman:
Closing in favor of #2256. unreal's sim pause game / set global dilation should be called in simmodeworldbase.cpp instead of simmodebase.cpp. 

with this PR, on hitting sim Pause, I see a warning in the unreal console:
```
[2019.11.04-18.52.25:101][434]LogBlueprintUserMessages: Warning: Time Dilation must be between 0.000100 and 20.000000.  Clamped value to that range.
```
this hints that the game is not actually paused, but it's going on really slow at 1e-4 rate. 

Googling this a bit yields https://forums.unrealengine.com/development-discussion/blueprint-visual-scripting/7217-slowmo-and-game-pause-via-blueprints-tutorial
quoting a relevant snippet from the above article:
```

Prepare the same set up as you did for the slow motion but replace the event with the one associated with your pause button. Change the time dilation in the node A to 0, compile and run the map. Unfortunately, when you hit the pause button, you can still see the ball moving. So, what's happened? Well, "Set Global Time Dilation" node cannot set Time Dilation to 0 as it will break some internal logic and will cause many other issues. When you set it to 0, it will automatically defaulted to a very small but non-zero number. In order to truly pause the game, we will need a "Set Game Paused" node (Call Function->Game->Set Game Paused). Replace our "Set Global Time Dilation" nodes with the "Set Game Paused" nodes and put the tick in the box of the one connected to the output "A" of the flip flop connected to the "Pause" event.
```

which finally yields the right API we should use `SetGamePaused/`: https://docs.unrealengine.com/en-US/API/Runtime/Engine/Kismet/UGameplayStatics/SetGamePaused/index.html

PR #2292, opened 2019-11-04
[simPause fix]pause unreal engine with SetGamePaused

madratman:


madratman:
see discussion in #2256 and #2282

sytelus:
This looks good to me.

PR #2300, opened 2019-11-06
[simPause/ContinueForTime] 

madratman:
move all unreal API calls to ASimModeWorldBase::Tick()


madratman:
closing as this is not working as I expected it would. the previous impl is good enough.

PR #2303, opened 2019-11-08
Fix unity settings file

wolf843:
test pull


rajat2004:
Hey @wolf843, does this also fix the problem of an extra `settings.json` file getting created in the main directory if the `settings.json` in `Documents/AirSim` is not present?
The diff is pretty big so it's a bit difficult to follow

Unity/UnityDemo/Assets/AirSimAssets/Scripts/Utilities/AirSimSettings.cs
```diff
@@ -37,287 +38,3753 @@ public struct CamerasSettings {
...
+                Debug.LogError("Error: 'settings.json' file SettingsVersion missing!  SettingsVersion attribute requried.");
+#if UNITY_EDITOR
+                EditorUtility.DisplayDialog("'settings.json' file SettingsVersion Error!!!",
+                    "'settings.json' file is missing or SettingsVersion", "Exit");
+#endif
+                Application.Quit();
+            }
+
+            // Verify SettingsVersion is greater than 1.2
+            if (SettingsVersion != 1.2)
```
rajat2004: Should be `<` here

rajat2004:
> Hey @wolf843, does this also fix the problem of an extra `settings.json` file getting created in the main directory if the `settings.json` in `Documents/AirSim` is not present?

I think the problem still persists, since the [GetAirSimSettingsFileName()](https://github.com/microsoft/AirSim/blob/50b9979a5ed44664f2b6032999014954359d5988/Unity/UnityDemo/Assets/AirSimAssets/Scripts/Utilities/AirSimSettings.cs#L2329-L2357) is the same as in the [older InitializeAirsim.cs](https://github.com/microsoft/AirSim/blob/master/Unity/UnityDemo/Assets/AirSimAssets/Scripts/InitializeAirSim.cs#L42-L67)

Could be mistaken, please correct me if that's the case

rajat2004:
@madratman I think the best way to get the settings fixed would be to do piecewise PRs or atleast smaller commits, will make it much easier to review and test

zimmy87:
closed due to license agreement not being signed and unanswered old CR feedback

PR #2310, opened 2019-11-11
add support for SITL PX4 Lockstep

lovettchris:
This makes it possible to use PX4 SITL version 1.9.* and above with the lockstep feature.  See [Lockstep Simulation](https://dev.px4.io/master/en/simulation/).   Flying without lockstep on v1.8.* is still supported.  Lockstep mode recommends a TCP connection which is new for AirSim.  The new settings.json for this will be:
```json
{
  "SettingsVersion": 1.2,
  "SimMode": "Multirotor",
  "Vehicles": {
    "PX4": {
        "VehicleType": "PX4Multirotor",
        "UseSerial": false,
        "UseTcp": true,
        "TcpPort": 4560,
        "ControlPort": 14580,
        "Parameters": {
            "NAV_RCL_ACT": 0,
            "NAV_DLL_ACT": 0,
            "LPE_LAT": 47.641468,
            "LPE_LON": -122.140165,
            "COM_OBL_ACT": 1
        }
    }
  }
}
```


sytelus:
Could you provide high-level overview of the changes? It seems like huge changeset and its changing many public signatures and classes. Questions such as...

1. What is the deal with UDP and ethernet? 
2. Would this work with old SITL? Should we keep things compatible and only add new features? 
3. Do we worry about people with old config json instantly breaking even if they are not using PX4/SITL but have some settings lying around?
4. What backward compatibility is broken from users perspective? I think it would be good to add little doc on this or add appropriate informative error message.

madratman:
In addition to Shital's comments - I've some minor things to say initially - can you also please cleanup the git history - aka squash the merge commits and rebase on master. there are quite some whitespace in the diff as well. If it is standardizing old files, it's good. If not, let's remove that. 
Have a look at https://github.com/mockito/mockito/wiki/Using-git-to-prepare-your-PR-to-have-a-clean-history / https://stackoverflow.com/questions/17577409/git-remove-merge-commit-from-history

lovettchris:
Thanks for the review comments.
1. PX4 SITL added TCP to support lockstep.  UDP does not guarantee delivery which is critical for lockstep to function properly.  
2. I added backward compat code to read old SitlIp and SitlPort settings.
3. Old non-lockstep mode on UDP port still works (I've tested on v1.8.2)
4. Fixed
5. Nothing is broken.

Ratnesh: Whitespace was a fix, one of my files had lots of TAB characters, so I untabbified the file.  Can't we just squash merge when PR is accepted using this:

image

AirLib/include/common/AirSimSettings.hpp
```diff
@@ -244,28 +245,35 @@ struct AirSimSettings {
...
+        bool use_tcp = false;
+        int tcp_port = 4560;
+
+        // The PX4 SITL app requires receiving drone commands over a different mavlink channel called
+        // the "ground control station" channel.  
         // So set this to empty string to disable this separate command channel.
-        std::string sitl_ip_address = "127.0.0.1";
-        int sitl_ip_port = 14556;
+        std::string gcs_address = "127.0.0.1";
+        int gcs_port = 14540;
```
julianoes: `14540` is actually the port usually used by MAVSDK or MAVROS while `14550` is used for QGC or other ground stations.

lovettchris: Right in this case "gcs_port" is taking the place of MAVSDK inside AirSim.  I'm calling it "ground_control_station_port" or "gcs_port" for short, because it is the port that AirSim can use to control the vehicle (so it is not the sitl port as it was named before).  The sitl port is the one that is used to send HIL_SENSOR messages.  Perhaps "gcs_port" is too close to "qgc_port" in naming though... but I couldn't think of a better name....

julianoes: Ok makes sense!

PythonClient/multirotor/path.py
```diff
@@ -8,26 +8,43 @@
 client.confirmConnection()
 client.enableApiControl(True)
 
+print("arming the drone...")
 client.armDisarm(True)
 
-landed = client.getMultirotorState().landed_state
-if landed == airsim.LandedState.Landed:
+state = client.getMultirotorState()
+if state.landed_state == airsim.LandedState.Landed:
```
sytelus: Is the API docs changed for this?

PythonClient/multirotor/path.py
```diff
@@ -8,26 +8,43 @@
...
     print("taking off...")
     client.takeoffAsync().join()
 else:
     client.hoverAsync().join()
 
+time.sleep(1)
+
+state = client.getMultirotorState()
```
sytelus: I am assuming this would work as before for other flight controllers?

PythonClient/multirotor/path.py
```diff
@@ -8,26 +8,43 @@
...
 # z of -7 is 7 meters above the original launch point.
 z = -7
+print("make sure we are hovering at 7 meters...")
+client.moveToZAsync(z, 1).join()
 
 # see https://github.com/Microsoft/AirSim/wiki/moveOnPath-demo
 
 # this method is async and we are not waiting for the result since we are passing timeout_sec=0.
-result = client.moveOnPathAsync([airsim.Vector3r(0,-253,z),airsim.Vector3r(125,-253,z),airsim.Vector3r(125,0,z),airsim.Vector3r(0,0,z),airsim.Vector3r(0,0,-20)], 
-                        12, 120,
```
sytelus: Are these parameters different just because of better visualization or because something is different from dynamics wise (i.e., would previous params work without issues)?

PythonClient/requirements.txt
```diff
@@ -0,0 +1,2 @@
+opencv-contrib-python
```
sytelus: We should delete requirements.txt. All dependencies should be in setup.py and require package redeployment (many users might be using this through pip). Are these new deps?

sytelus:
Overall looks pretty good. It would be great if you could add few comments as mentioned in the review. That would help me and anyone else in future to understand code better.

sytelus (changes requested):
Overall looks pretty good. It would be great if you could add few comments as mentioned in the review. That would help me and anyone else in the future to understand code better.

AirLib/include/vehicles/multirotor/firmwares/arducopter/ArduCopterApi.hpp
```diff
@@ -34,7 +34,7 @@ class ArduCopterApi : public MultirotorApiBase {
 
 public:
     ArduCopterApi(const MultiRotorParams* vehicle_params, const AirSimSettings::MavLinkConnectionInfo& connection_info)
-        : ip(connection_info.ip_address), vehicle_params_(vehicle_params)
+        : ip(connection_info.udp_address), vehicle_params_(vehicle_params)
```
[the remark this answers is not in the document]
rajat2004: ArduCopter also uses UDP only

rajat2004:
The naming of the settings fields seems strange, will comment if I have some suggestions.
Since the settings will be changed significantly, we could do something like bump up the `SettingsVersion` field, so if any person if having that as 1.3, then the new fields would be looked for, otherwise, the old settings would apply. 
I think that should be good enough approach for keeping backwards compatibility, and would also cause fewer support problems. Just a suggestion, what does everyone think of this?

PythonClient/multirotor/path.py
lovettchris: there is no api change here.  It is just capturing the state in a variable, then using that.

AirLib/include/common/AirSimSettings.hpp
```diff
@@ -8,6 +8,7 @@
 #include <vector>
 #include <exception>
 #include <functional>
+#include <map>
```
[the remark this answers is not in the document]
lovettchris: because it is used in the same file "std::map<int, CaptureSetting> capture_settings".  I must have included AirSimSettings in a new place which created a compile error because #include <map> was missing.

AirLib/include/common/common_utils/MedianFilter.hpp
```diff
@@ -13,7 +13,7 @@ namespace common_utils {
 template <typename T>
 class MedianFilter {
     private:
-        std::vector<T> buffer_, buffer_copy_;
```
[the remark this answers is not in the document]
lovettchris: I restored this file to it's original and added a new filter called SmoothingFilter and added comments.

MavLinkCom/common_utils/ThreadUtils.cpp
```diff
@@ -34,3 +36,34 @@ bool CurrentThread::setMaximumPriority()
...
+
+#ifdef _WIN32
+typedef HRESULT (WINAPI *SetThreadDescriptionFunction)( _In_ HANDLE hThread, _In_ PCWSTR lpThreadDescription);
+static SetThreadDescriptionFunction setThreadDescriptionFunction = nullptr;
+#endif
+
+bool CurrentThread::setThreadName(const std::string& name)
+{
+#ifdef _WIN32
+    // unfortunately this is only available on Windows 10, and AirSim is not limited to that.
```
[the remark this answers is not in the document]
lovettchris: Added comments, and there is a #else block for linux/mac machines to use.

AirLib/include/vehicles/multirotor/api/MultirotorCommon.hpp
```diff
@@ -74,15 +74,18 @@ struct MultirotorState {
     uint64_t timestamp;
     LandedState landed_state;
     RCData rc_data;
+    bool ready;
```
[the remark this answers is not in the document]
lovettchris: done

PythonClient/multirotor/path.py
lovettchris: there was a bad bug before that caused drone position to be wrong in python client, after fixing that, these changes were needed to make the drone not hit telephone poles, but it is still a good question, we should test this still works on v1.8.2 but I would say that is a pretty low priority at this point.

PythonClient/requirements.txt
lovettchris: done.

PythonClient/multirotor/path.py
lovettchris: Yes, still need to test that simpleflight works on this map (but path.py requires neighborhood map to be setup with the drone in the correct takeoff location), I suppose we should add a teleport step at the beginning of this script to ensure drone is in that location.

sytelus:
@madratman - changes looks good to me but I haven't tested it myself. If you are happy, please feel free to merge.

rajat2004:
I've tested this branch with ArduPilot, no functional changes as expected, backwards compatibility of settings is working fine. Haven't tested the PX4 side till now since did a new system installation recently and PX4 setup is left, will get to it soon. The renaming of the `sitl_ip_port` to `gcs_port` is still very strange, especially in the context of ArduPilot, which uses that port to send motor control signals to. But it is appropriate for its intended use with the PX4 SITL App, so not much choice there. 

I was thinking of adding another field to the Settings, something like `control_port`, to replace the earlier `sitl_ip_port`, will be used just for ArduPilot, and others who might want to use AirSim as a Physics Backend as ArduPilot does. After this PR goes in.

There were some problems with the altitude reported by AirSim which I noticed while testing, but it's not due to any changes in this PR, will try to look into it and maybe open an issue regarding that

PythonClient/multirotor/path.py
lovettchris: I tested SimpleFlight and it works.

lovettchris:
Rajat, thanks for testing ArduPilot.  Regarding your question on renaming of sitl_ip_port to gcs_port, my reasoning is as follows.   With PX4 in SITL mode one needs 2 sockets, a "hil" socket that sends HIL_SENSOR and other HIL_* messages, but this "hil" socket cannot be used to send flight commands.  This socket is connected via the "udp_port" or "tcp_port".  So as you can see naming the second socket "sitl_ip_port" the one that can sent flight controls, is a very confusing name.  So I renamed it "gcs_port" which stands for "ground control station port", since it represents the port that can be used to "control" the drone.  Renaming that "control_port" would be fine too.  So anyway, I don't know if ArduPilot needs 2 sockets, if not, then it should use udp_port and ignore gcs_port.

rajat2004:
@lovettchris ArduPilot also needs 2 ports, there's actually only 1 socket being used, but there are 2 port numbers, one for the socket to bind to, earlier named `sitl_ip_port` and now changed to `gcs_port`. This port number is used to receive the motor control messages, i.e ArduPilot sends the motor signals to this port number.
The other port number, `udp_port` is used to tell AirSim which port number to send the sensor signals to (ArduPilot will bind to that port number). So it needs 2 ports.

The idea behind the renaming to `gcs_port` is clearer now, but `gcs` could be confusing since users might try to use that port to connect to a GCS such as QGC or something else which should actually be the `qgc_port` if I understand correctly. 
Since the usage of the `gcs_port` is to actually control the drone, I feel `control_port` could be a better choice. Your thoughts on this?

lovettchris:
@rajat2004, I see thanks.  So ArduPilot should use udp_port to bootstrap and ignore gcs_port.  I agree control_port is fine and gcs_port could be confused with qgc_port.  So I'll rename it "control_port".

rajat2004:
@lovettchris Great, thanks a lot!

lovettchris:
Ok, I pushed that fix, and here's a new build of neighborhood environment with the rename:
[Neighborhood.zip](https://drive.google.com/file/d/1YOSlA8Iy50oSih42otatt_aAgsVN54dN/view?usp=sharing).

rajat2004:
Did the PX4 setup today, it installed `clang-6.0` which doesn't work with Unreal Engine 4.18, need to remove it first in order to use UE Editor. Will need to add a note for that, though not sure if that's just with the master branch or with the 1.9.2 release also. This was with Ubuntu 18.04 though, might not be the case for 16.04.

Also, one more thing which I would like to ask, any thoughts on having a new release of AirSim? This PR would be a big reason to have one, and there are several other fixes and features, such as the ArduCopter support, already gone in and more in the current open PRs. It's been quite some time since the last release, especially for Linux, has been about a year since then. 
Your opinions on this matter @lovettchris @madratman @sytelus ?

lovettchris:
@rajat2004, thanks for the info about conflicting clang-6 required by PX4.  I agree a new release is in order after this PR goes in. I will bring it up with the team.

saifullah3396:
I'm getting the following errors on connecting with PX4 only when running the blocks environment exported as package from UE4Editor. I don't get this error when I run the environment inside the UE4Editor using the play button. Any clue why this might be happening? 
```
Connected to SITL over TCP.
Connecting to PX4 Control UDP port 14580, local IP 127.0.0.1, remote IP...
received first heartbeat
[2019.12.13-10.35.44:989][460]LogTemp: Connected to SITL over TCP.
[2019.12.13-10.35.44:989][460]LogTemp: Connecting to PX4 Control UDP port 14580, local IP 127.0.0.1, remote IP...
[2019.12.13-10.35.44:989][460]LogTemp: received first heartbeat
Ground control connected over UDP.
terminating with uncaught exception of type std::runtime_error: UdpClientPort socket send failed with error: 9
Signal 6 caught.
Malloc Size=131076 LargeMemoryPoolOffset=131092 
CommonLinuxCrashHandler: Signal=6
Malloc Size=65535 LargeMemoryPoolOffset=196655 
Malloc Size=44263 LargeMemoryPoolOffset=240935 
Engine crash handling finished; re-raising signal 6 for the default handler. Good bye.
Aborted (core dumped)
```
Can anyone else please check if this error is reproducable?

lovettchris:
@saifullah3396, hmmm, would be great if you could help debug this some more, I can't repo that bug.   The only thing that happens on the UdpClientPort immediately after "Ground control connected over UDP" is sending of heartbeats, which is a sendMessage call, but that happens deep down in this code which catches any exceptions from the port and rethrows with a different message, so it is strange that the UdpClientPort exception bubbled up as an unhandled exception:

```cpp
        try {
            port->write(message_buf, len);
        }
        catch (std::exception& e) {
            throw std::runtime_error(Utils::stringf("MavLinkConnectionImpl: Error sending message on connection '%s', details: %s", name.c_str(), e.what()));
        }
```

saifullah3396:
@lovettchris I'll try to see if I can pinpoint the exception since it seems to be raised from a part of the code different from what you've shown. I'll also update the code with your latest [commit](https://github.com/microsoft/AirSim/pull/2310/commits/b3304c2ae7a026782fbf73030998ae1e5c04f03b) and see if that resolves the issue.

saifullah3396:
Hi @lovettchris, it is defined in PX4MultirotorApithat the tcp connectoin should be remade if broken and if px4 is restarted, however, I am not able to reconnect right afterwards. I have seen the configuration of the tcp sockets used in this case and I don't see the SO_REUSEADDR and SO_REUSEPORT options in there. Maybe adding those might fix this? I get this error on reconnection with px4.
```
Waiting for TCP connection on port 4560, local IP 127.0.0.1
Accepting TCP socket failed, is another instance running?
TcpClientPort socket bind failed with error: 98
```

saifullah3396:
> @saifullah3396, hmmm, would be great if you could help debug this some more, I can't repo that bug. The only thing that happens on the UdpClientPort immediately after "Ground control connected over UDP" is sending of heartbeats, which is a sendMessage call, but that happens deep down in this code which catches any exceptions from the port and rethrows with a different message, so it is strange that the UdpClientPort exception bubbled up as an unhandled exception:
> 
> ```c++
>         try {
>             port->write(message_buf, len);
>         }
>         catch (std::exception& e) {
>             throw std::runtime_error(Utils::stringf("MavLinkConnectionImpl: Error sending message on connection '%s', details: %s", name.c_str(), e.what()));
>         }
> ```
> 
> And this is caught higher up and logged. It would be helpful to see the PX4 console output also. I do see this exception getting thrown occasionally on Windows, but it is always caught.

The error is begin raised when making a connection to logviewer in `MavLinkMultiRotorApi::connectToLogViewer()`. The error is raised at join() which is being called inside `MavLinkMultiRotorApi::createProxy()`. I resolved the issue for now by setting "LogViewerHostIp": "" in settings.json. It's confusing for me why this error is raised only when the blocks environment is launched as a separate package? I'm not running log viewer here and haven't checked if it works otherwise. Here is the stack trace for the error:
```
(gdb) backtrace
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff6841801 in __GI_abort () at abort.c:79
#2  0x0000000006636ffd in abort_message ()
#3  0x00000000066370f1 in default_terminate_handler() ()
#4  0x0000000006636ea3 in std::__terminate(void (*)()) ()
#5  0x0000000006636926 in __cxa_throw ()
#6  0x000000000605102c in UdpClientPort::UdpSocketImpl::write (this=0x7fffd823c040, ptr=0x7fffd5f3ba08 "\376\020\002\001\001o", count=24)
    at /home/sai/air_sim/AirSim/MavLinkCom/src/serial_com/UdpClientPort.cpp:195
#7  0x00000000060506f2 in UdpClientPort::write (this=0x7fffd823e998, ptr=0x7fffd5f3ba08 "\376\020\002\001\001o", count=24) at /home/sai/air_sim/AirSim/MavLinkCom/src/serial_com/UdpClientPort.cpp:318
#8  0x0000000006007847 in mavlinkcom_impl::MavLinkConnectionImpl::sendMessage (this=0x7fffd5f3b920, m=...) at /home/sai/air_sim/AirSim/MavLinkCom/src/impl/MavLinkConnectionImpl.cpp:226
#9  0x0000000005f8102c in mavlinkcom::MavLinkConnection::sendMessage (this=0x7fffd823c0e8, msg=...) at /home/sai/air_sim/AirSim/MavLinkCom/src/MavLinkConnection.cpp:98
#10 0x0000000006009242 in mavlinkcom_impl::MavLinkConnectionImpl::joinLeftSubscriber (this=0x7fffd5f3b530, remote=..., connection=..., msg=...)
    at /home/sai/air_sim/AirSim/MavLinkCom/src/impl/MavLinkConnectionImpl.cpp:373
#11 0x0000000006012e00 in std::__1::__invoke<void (mavlinkcom_impl::MavLinkConnectionImpl::*&)(std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, mavlinkcom::MavLinkMessage const&), mavlinkcom_impl::MavLinkConnectionImpl*&, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>&, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, mavlinkcom::MavLinkMessage const&, void> (__f=
    @0x7fffd823c0a8: (void (mavlinkcom_impl::MavLinkConnectionImpl::*)(mavlinkcom_impl::MavLinkConnectionImpl * const, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, const mavlinkcom::MavLinkMessage &)) 0x60091c0 <mavlinkcom_impl::MavLinkConnectionImpl::joinLeftSubscriber(std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, mavlinkcom::MavLinkMessage const&)>, __a0=@0x7fffd823c0b8: 0x7fffd5f3b530, __args=..., __args=..., __args=...)
    at /home/sai/air_sim/AirSim/llvm-build/output/include/c++/v1/type_traits:4264
#12 std::__1::__apply_functor<void (mavlinkcom_impl::MavLinkConnectionImpl::*)(std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, mavlinkcom::MavLinkMessage const&), std::__1::tuple<mavlinkcom_impl::MavLinkConnectionImpl*, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, std::__1::placeholders::__ph<1>, std::__1::placeholders::__ph<2> >, 0ul, 1ul, 2ul, 3ul, std::__1::tuple<std::__1::shared_ptr<mavlinkcom::MavLinkConnection>&&, mavlinkcom::MavLinkMessage const&> > (__f=
    @0x7fffd823c0a8: (void (mavlinkcom_impl::MavLinkConnectionImpl::*)(mavlinkcom_impl::MavLinkConnectionImpl * const, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, const mavlinkcom::MavLinkMessage &)) 0x60091c0 <mavlinkcom_impl::MavLinkConnectionImpl::joinLeftSubscriber(std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, mavlinkcom::MavLinkMessage const&)>, __bound_args=..., __args=...) at /home/sai/air_sim/AirSim/llvm-build/output/include/c++/v1/functional:2219
#13 std::__1::__bind<void (mavlinkcom_impl::MavLinkConnectionImpl::*)(std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, mavlinkcom::MavLinkMessage const&), mavlinkcom_impl::MavLinkConnectionImpl*, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>&, std::__1::placeholders::__ph<1> const&, std::__1::placeholders::__ph<2> const&>::operator()<std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, mavlinkcom::MavLinkMessage const&> (this=0x7fffd823c0a8, __args=..., __args=...) at /home/sai/air_sim/AirSim/llvm-build/output/include/c++/v1/functional:2252
#14 std::__1::__invoke<std::__1::__bind<void (mavlinkcom_impl::MavLinkConnectionImpl::*)(std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, mavlinkcom::MavLinkMessage const&), mavlinkcom_impl::MavLinkConnectionImpl*, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>&, std::__1::placeholders::__ph<1> const&, std::__1::placeholders::__ph<2> const&>&, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, mavlinkcom::MavLinkMessage const&> (__f=..., __args=..., __args=...) at /home/sai/air_sim/AirSim/llvm-build/output/include/c++/v1/type_traits:4323
#15 std::__1::__invoke_void_return_wrapper<void>::__call<std::__1::__bind<void (mavlinkcom_impl::MavLinkConnectionImpl::*)(std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, mavlinkcom::MavLinkMessage const&), mavlinkcom_impl::MavLinkConnectionImpl*, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>&, std::__1::placeholders::__ph<1> const&, std::__1::placeholders::__ph<2> const&>&, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, mavlinkcom::MavLinkMessage const&> (__args=..., __args=..., __args=...)
    at /home/sai/air_sim/AirSim/llvm-build/output/include/c++/v1/__functional_base:349
#16 0x000000000601251c in std::__1::__function::__func<std::__1::__bind<void (mavlinkcom_impl::MavLinkConnectionImpl::*)(std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, mavlinkcom::MavLinkMessage const&), mavlinkcom_impl::MavLinkConnectionImpl*, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>&, std::__1::placeholders::__ph<1> const&, std::__1::placeholders::__ph<2> const&>, std::__1::allocator<std::__1::__bind<void (mavlinkcom_impl::MavLinkConnectionImpl::*)(std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, mavlinkcom::MavLinkMessage const&), mavlinkcom_impl::MavLinkConnectionImpl*, std::__1::shared_ptr<mavlinkcom::MavLinkConnection>&, std::__1::placeholders::__ph<1> const&, std::__1::placeholders::__ph<2> const&> >, void (std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, mavlinkcom::MavLinkMessage const&)>::operator()(std::__1::shared_ptr<mavlinkcom::MavLinkConnection>&&, mavlinkcom::MavLinkMessage const&) (this=0x7fffd823c0a0, __arg=..., __arg=...) at /home/sai/air_sim/AirSim/llvm-build/output/include/c++/v1/functional:1562
#17 0x000000000600d502 in std::__1::function<void (std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, mavlinkcom::MavLinkMessage const&)>::operator()(std::__1::shared_ptr<mavlinkcom::MavLinkConnection>, mavlinkcom::MavLinkMessage const&) const (this=0x7fffc0b6e290, __arg=..., __arg=...) at /home/sai/air_sim/AirSim/llvm-build/output/include/c++/v1/functional:1916
#18 0x000000000600ac59 in mavlinkcom_impl::MavLinkConnectionImpl::drainQueue (this=0x7fffd5f3b530) at /home/sai/air_sim/AirSim/MavLinkCom/src/impl/MavLinkConnectionImpl.cpp:537
#19 0x0000000006005cc3 in mavlinkcom_impl::MavLinkConnectionImpl::publishPackets (this=0x7fffd5f3b530) at /home/sai/air_sim/AirSim/MavLinkCom/src/impl/MavLinkConnectionImpl.cpp:563
#20 0x000000000600e574 in std::__1::__invoke<void (mavlinkcom_impl::MavLinkConnectionImpl::*)(), mavlinkcom_impl::MavLinkConnectionImpl*, , void> (
    __f=@0x7fffe42615a8: (void (mavlinkcom_impl::MavLinkConnectionImpl::*)(mavlinkcom_impl::MavLinkConnectionImpl * const)) 0x6005bd0 <mavlinkcom_impl::MavLinkConnectionImpl::publishPackets()>, 
    __a0=@0x7fffe42615b8: 0x7fffd5f3b530) at /home/sai/air_sim/AirSim/llvm-build/output/include/c++/v1/type_traits:4264
#21 std::__1::__thread_execute<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, void (mavlinkcom_impl::MavLinkConnectionImpl::*)(), mavlinkcom_impl::MavLinkConnectionImpl*, 2ul> (__t=...) at /home/sai/air_sim/AirSim/llvm-build/output/include/c++/v1/thread:342
#22 std::__1::__thread_proxy<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, void (mavlinkcom_impl::MavLinkConnectionImpl::*)(), mavl---Type <return> to continue, or q <return> to quit---
inkcom_impl::MavLinkConnectionImpl*> > (__vp=0x7fffe42615a0) at /home/sai/air_sim/AirSim/llvm-build/output/include/c++/v1/thread:352
#23 0x00007ffff7bbd6db in start_thread (arg=0x7fff827fc700) at pthread_create.c:463
#24 0x00007ffff692288f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

rajat2004:
Mentioning it here since this is more active, could a repo owner look into restarting the Travis build? It can't test with UE, but still is useful for an initial test. There's also an open PR https://github.com/microsoft/AirSim/pull/2356 for reworking travis.yml and adding Windows build which is where things break more frequently
Thanks!

pedrogasg:
Hi everyone, I start testing this solution and I can confirm as @saifullah3396 mentioned we need to add `"LogViewerHostIp": ""` to make this work without logviewer. Everything else seems to work without another issue

lovettchris:
@rajat2004 , thanks for the stack trace.  I can see why that is throwing an exception if log viewer is not there, but I can't see why that crashes the app, there should be a try/catch higher up the stack in MavLinkConnectionImpl::drainQueue.  But for now I will change the default setting to: `"LogViewerHostIp": ""`.

PythonClient/multirotor/path.py
lovettchris: Have now tested both v1.8.2 and v1.9 work correctly with this new path.py.

lovettchris:
New build of [Neighborhood map ](https://drive.google.com/open?id=1g01e8den4wmWu59ix97c6eJ9k65GEM4-) with the latest bits.

saifullah3396:
@lovettchris It is important to note that this error is only raised in packaged project. In the editor on pressing play and connecting with px4 I do get the upper level exception. 
```
[2020.01.08-06.20.16:431][ 84]LogTemp: Ground control connected over UDP.
Caught and ignoring exception sending heartbeat: MavLinkConnectionImpl: Error sending message on connection 'Proxy to: LogViewerOut at 127.0.0.1:14389', details: UdpClientPort socket send failed with error: 111
MavLinkConnectionImpl: Error handling message 74 on connection 'gcs', details: MavLinkConnectionImpl: Error sending message on connection 'Proxy to: LogViewer at 127.0.0.1:14388', details: UdpClientPort socket send failed with error: 9
Got GPS lock
[2020.01.08-06.20.18:460][ 85]LogTemp: Got GPS lock
Enabling lockstep mode
```
Could there be some build flags in unreal engine that turns off exception handling while packaging the project? I use the following script for packaging:
```
$UnrealDir/Engine/Build/BatchFiles/RunUAT.sh BuildCookRun -project="$SCRIPT_DIR/Unreal/Environments/Blocks/Blocks.uproject" -clientconfig=Debug -cook -allmaps -build -stage -pak -archive -archivedirectory="$SCRIPT_DIR/Unreal/Environments/Blocks/Saved/Package"
```

lovettchris:
@saifullah3396, I see, well that is interesting, there must be some sort of build difference then in how exceptions work in the packaged version...

Issue #2305, opened 2019-11-09
"Edit on GitHub" links are broken


lovettchris:
At all the pages have an "Edit on GitHub" link, but clicking that link leads to 404 from github.

For example, this page:
https://microsoft.github.io/AirSim/docs/px4_sitl/

Gives this broken link:
https://github.com/microsoft/airsim/edit/master/docs/docs/px4_sitl.md

rajat2004:
I've opened #2313 for fixing this

jonyMarino:
Being treated in #2313

rajat2004:
This can be closed actually, `Edit on Github` button's working

jonyMarino:
I wasn't sure because the links on the first post are not working. But may be they are not on the documents any more

PR #2313, opened 2019-11-14
Fix Edit on Github button

rajat2004:
#2305

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -5,7 +5,7 @@ site_dir: build
 docs_dir: doc_root
 repo_url: https://github.com/microsoft/airsim
 #site_url: http://microsoft.github.io/airsim
-#edit_uri: edit/master/docs
+edit_uri: edit/master/
 #repo_name: GitHub
 site_description: 'Open source simulator based on Unreal Engine for autonomous vehicles from Microsoft AI & Research'
 


madratman:
Hmm, so these htmls are autogenerated by [mkdocs](https://www.mkdocs.org/). Pretty sure there's a one liner fix for this somewhere on the gh-pages branch.. 
cc @sytelus?

madratman:
Hmm, so these htmls are autogenerated by [mkdocs](https://www.mkdocs.org/). Pretty sure there's a one liner fix for this somewhere on the gh-pages branch.. 
cc @sytelus?

rajat2004:
@madratman Yeah, I thought so too, but this is my first time working with Github pages so wasn't sure exactly what to do and thought to give it a shot
Thanks for pointing in the right direction, I think the problem is here -
https://github.com/microsoft/AirSim/blob/4d6e0ad12196764d1e480cb2af4b075ea84409da/docs/mkdocs.yml#L8

It has been commented out, but that's the default value set by [mkdocs edit-uri](https://www.mkdocs.org/user-guide/configuration/#edit_uri)
Below, in the `nav` part, the files are given relative paths, which are appended to the `edit-uri`
https://github.com/microsoft/AirSim/blob/4d6e0ad12196764d1e480cb2af4b075ea84409da/docs/mkdocs.yml#L112

So I think there 2 fixes here, first 1-liner, specify the `edit-uri` to `edit/master/`
Else, remove the `docs/` from all the file names specified below

I think the first one seems correct, I'll go ahead and make the change, if the second one is better, let me know and I'll do that

rajat2004:
Changes made

madratman:
thanks!

rajat2004:
@madratman I'm not sure why but the issue hasn't been fixed till now, the Edit button is still pointing to the incorrect location. 
Either it's due to some problem with the pages deployment (pages haven't been deployed yet)
Else, this is the incorrect fix. The problem here is that I'm not able to understand how the `mkdocs` layout is working in Airsim, in the Mkdocs documentation, everywhere, the `mkdocs.yml` is in the root folder, with the `docs` folder being one of the folders in the root, rather than being in the `docs` directory itself. This is causing problems while trying to build the pages locally also

madratman:
It's this - ' Either it's due to some problem with the pages deployment (pages haven't been deployed yet)'. I'll look into it

madratman:
This wasn't the fix. 
I have manually built the docs and updated the gh pages. 
the edit button (and all the python->c++) fixes should be there. 
mkdocs is a bit weird as far as relative links are concerned. 
Now, the edit on gh button fails for the markdowns on the root dir of the repo (aka readme, changelong, support, issue_template, contributing.mds). However, it works for everything else. 
IMO, this is the lesser of the two devils. 
There are a few issues on mkdocs repo wrt relative links. Will look into automating this.

rajat2004:
Sorry for the wrong fix, and thanks for updating the docs!

Issue #1677, opened 2019-01-04
Incorrect behavior of MoveToPositionAsync #1643，how to solve it?


love-mj:
The arrival of the target will vary considerably with speed.
The UE4 coordinate is converted to the ned coordinate using tolocalned, but the arrival of the target point will vary greatly with the speed.I don't know why?

[](url)
![image](https://user-images.githubusercontent.com/37679671/50684303-f75c6880-104f-11e9-892c-41b2be6d388b.png)

majouda7:
I have experienced exactly the same issue (I posted it in Incorrect behavior of MoveToPositionAsync #1643). It seems that AirSim is not aware about this offset in unreal because when I call client.simGetVehiclePose().position or client.simGetGroundTruthKinematics().position, it returns exactly the same requested position.
My setting.json file looks like this:

{
  "SeeDocsAt": "https://github.com/Microsoft/AirSim/blob/master/docs/settings.md",
  "SettingsVersion": 1.2,
  "SimMode": "Multirotor",
  "Vehicles": {
    "SimpleFlight": {
      "VehicleType": "SimpleFlight",
      "AutoCreate": true,


      "Sensors": {
        "LidarSensor1": {
          "SensorType": 6,
          "Enabled": true,
          "Range": 100,
          "NumberOfChannels": 2,
          "RotationsPerSecond": 10,
          "PointsPerSecond": 100000,
          "X": 0,
          "Y": 0,
          "Z": -0.8,
          "Roll": 0,
          "Pitch": 0,
          "Yaw": 0,
          "VerticalFOVUpper": 0,
          "VerticalFOVLower": 0,
          "HorizontalFOVStart": 0,
          "HorizontalFOVEnd": 359,
          "DrawDebugPoints": true,
          "DataFrame": "SensorLocalFrame"
        }
      }
    }
  }
}

love-mj:
@lovettchris My setting.json is the default mode,
{
  "SettingsVersion": 1.2,
  "SimMode": "Multirotor",
  "Vehicles": {
		"PX4": {
		   "VehicleType": "PX4Multirotor",
		   "SerialPort": "*"
		}
	}
}

saihv:
I think I have the same problem. I am trying to control multiple SimpleFlight drones with the Python API, and I am encountering errors in how MoveToPositionAsync actually moves the drones. For instance, this is a set of commands I gave the drone:

```
[...]

PR #2324, opened 2019-11-22
Fix losing physics after setting pose

### LINKED ISSUE #1246: data mismatch between simGetVehiclePose() and simSetVehiclePose()
>>> stepsize= [0,0,10]
>>> airsim_client.simSetVehiclePose(airsim.Pose(airsim.Vector3r(currentposition.x_val+ stepsize[0], currentposition.y_val+stepsize[1], currentposition.z_val+stepsize[2]), airsim.to_quaternion(0, 0, 20)), True)
>>> currentposition.x_val+ stepsize[0]
3.0517577442878974e-07
>>> currentposition.y_val+stepsize[1]
0.0
>>> currentposition.z_val+stepsize[2]
**21.820711135864258**
>>> currentposition = airsim_client.simGetVehiclePose().position
>>> print(currentposition)
<Vector3r> {   'x_val': 3.0517577442878974e-07,
    'y_val': 0.0,
    'z_val': **11.820711135864258**}

### LINKED ISSUE #1272: Can't control drone after using simSetVehiclePose
When I use simSetVehiclePose, the drone is set at the right position (as long as there is no collision) and I can still see the rotors moving depending on my control inputs.
However, the drone is stuck at the same position.

I don't see anything missing from the description of that function here https://github.com/Microsoft/AirSim/blob/master/docs/apis.md.

What am I doing wrong ?

Also, what is the "teleport" parameter that is mentioned ? The only parameters of that function are "pose", "ignore_collision" and "vehicle_name"

### LINKED ISSUE #1856: "simSetVehiclePose" Simulation Mode Change
Hello,

I want to use the "simSetVehiclePose" to respawn the drone randomly in the environment. However, whenever I use this function "simSetVehiclePose", the computer vision mode is activated and all physics of the drone are gone. Is it possible to switch back to "Multirotor" Mode using python API?

Thank you

-------------------------------
Ubuntu 18.04

AIRSIM settings.json
  "SettingsVersion": 1.2,
  "SimMode": "Multirotor"

### LINKED ISSUE #3049: client.simSetVehiclePose() and then client.simGetVehiclePose() mismatch
This is probably the same problem like in https://github.com/microsoft/AirSim/issues/1246
I managed to solve/hack it with:

client.simSetVehiclePose(some pose) 
sleep(1.0)
client.simGetVehiclePose()

But, it didn't work with sleep(0.1) ... I guess simSetVehiclePose() doesn't happen right away, but added to some "TODO list"....
Is there a way to "flush" this TODO list ?

songxiaocheng:
Fixes #1272
Fixes #1856
Fixes #1246
Fixes #3049 : After using `simSetVehiclePose` in multirotor mode, the drone can't be controlled by most of the vehicle commands.
It is because when starting playing, the physical body is set to be grounded. Then we set pose directly, but the `body.isGrounded()` is still true, which causes `FastPhysicsEngine` not working as expected.

### DIFF
--- a/AirLib/src/vehicles/multirotor/api/MultirotorApiBase.cpp
+++ b/AirLib/src/vehicles/multirotor/api/MultirotorApiBase.cpp
@@ -269,7 +269,7 @@ bool MultirotorApiBase::moveOnPath(const vector<Vector3r>& path, float velocity,
     float goal_dist = 0;
 
     //until we are at the end of the path (last seg is always zero size)
-    while (!waiter.isTimeout() && (next_path_loc.seg_index < path_segs.size()-1 || goal_dist > 0)
+    while (!waiter.isTimeout() && (next_path_loc.seg_index < path_segs.size()-1 || goal_dist >= 0)
         ) { //current position is approximately at the last end point
 
         float seg_velocity = path_segs.at(next_path_loc.seg_index).seg_velocity;
--- a/AirLib/include/physics/FastPhysicsEngine.hpp
+++ b/AirLib/include/physics/FastPhysicsEngine.hpp
@@ -69,6 +69,7 @@ private:
     {
         TTimeDelta dt = clock()->updateSince(body.last_kinematics_time);
 
+        body.lock();
         //get current kinematics state of the body - this state existed since last dt seconds
         const Kinematics::State& current = body.getKinematics();
         Kinematics::State next;
@@ -95,6 +96,7 @@ private:
 
         body.setWrench(next_wrench);
         body.updateKinematics(next);
+        body.unlock();
 
 
 		//TODO: this is now being done in PawnSimApi::update. We need to re-think this sequence
--- a/AirLib/include/physics/PhysicsBody.hpp
+++ b/AirLib/include/physics/PhysicsBody.hpp
@@ -225,6 +225,16 @@ public: //methods
 		grounded_ = grounded;
 	}
 
+	void lock()
+	{
+		mutex_.lock();
+	}
+
+	void unlock()
+	{
+		mutex_.unlock();
+	}
+
 public:
     //for use in physics engine: //TODO: use getter/setter or friend method?
     TTimePoint last_kinematics_time;
@@ -243,6 +253,7 @@ private:
     CollisionResponse collision_response_;
 
 	bool grounded_ = false;
+	std::mutex mutex_;
 };
 
 }} //namespace
--- a/Unreal/Plugins/AirSim/Source/Vehicles/Multirotor/MultirotorPawnSimApi.cpp
+++ b/Unreal/Plugins/AirSim/Source/Vehicles/Multirotor/MultirotorPawnSimApi.cpp
@@ -10,12 +10,6 @@ MultirotorPawnSimApi::MultirotorPawnSimApi(const Params& params)
     : PawnSimApi(params),
       pawn_events_(static_cast<MultirotorPawnEvents*>(params.pawn_events))
 {
-    //reset roll & pitch of vehicle as multirotors required to be on plain surface at start
-    Pose pose = getPose();
-    float pitch, roll, yaw;
-    VectorMath::toEulerianAngle(pose.orientation, pitch, roll, yaw);
-    pose.orientation = VectorMath::toQuaternion(0, 0, yaw);
-    setPose(pose, false);
 }
 
 void MultirotorPawnSimApi::initialize()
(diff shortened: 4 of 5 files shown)


laurelkeys:
Hey @songxiaocheng I've tried your solution but there's still some unexpect behavior when calling `simSetVehiclePose` consecutive times, the drone seems to always start flying back to the first position it teleported to.

Could you test this short example on `Blocks` to check if the same problem occurs for you?
```python
ORIENTATION = airsim.Quaternionr(0.012, -0.156, 0.076, 0.985)

POSITIONS = [(15, 18, -40),  (55, 18, -40), (55, -20, -40), (15, -20, -40), (15, 18, -40)]

client.confirmConnection()
client.enableApiControl(True)
client.armDisarm(True)

try:
    if client.getMultirotorState().landed_state == airsim.LandedState.Landed:
        client.takeoffAsync(timeout_sec=5).join()
    client.hoverAsync().join()  

    for position in POSITIONS:
        pose = airsim.Pose(airsim.Vector3r(*position), ORIENTATION)
        client.simSetVehiclePose(pose, ignore_collison=True)
        client.hoverAsync().join()
        time.sleep(2)

    client.landAsync().join()
except KeyboardInterrupt:
    client.reset()
```

Any clues of why this happens/how to fix it would be much appreciated 🙂

p.s. I've also tried things like `client.moveByManualAsync(0, 0, 0, duration=2).join()` instead of `time.sleep(2)` without success, so that doesn't seem to be the cause of the problem.

songxiaocheng:
Hi @laurelkeys, thanks for your feedback, and I just figured out where the problem is.

By the way, as the author said, [the `simSetVehiclePose` is meant for CV mode](https://github.com/microsoft/AirSim/issues/1272#issuecomment-407269765). So as not to mislead others, the above short example is only for debugging, rather than the conventional way to use Airsim. However, I think `simSetVehiclePose` is also pretty useful for non-CV mode if it works well.

At first, there are two facts in the current implementation:

- The `hover` command eventually uses `moveOnPath(path,...)` function, where the input `path` contains only one point, that is the current point. However, since the drone is already at the current point, the function `moveOnPath` invokes NO control command. See `MultirotorApiBase` for more detail.
https://github.com/microsoft/AirSim/blob/7d398add700ce4ba9ebe150dffabd9267f77a0f2/AirLib/src/vehicles/multirotor/api/MultirotorApiBase.cpp#L269-L274
- If there is enough long time where no offboard API calls received, the drone will enter "hover mode" (do not confuse it with the hover API command) for safety, that is, the `goal_mode_` was set to "PositionMode" and the `goal_` was set to the current position. See `OffboardApi::update()` for more detail.
https://github.com/microsoft/AirSim/blob/99208a26e908a156624a8ccfc8aac72029aa75ea/AirLib/include/vehicles/multirotor/firmwares/simple_flight/firmware/OffboardApi.hpp#L43-L59

Now, let me try to explain what happened in your example. 

1. When the first time you set the pose and then hover, everything is OK (although `hover` command did nothing). Suppose the drone was at (x0, y0, z0) at that time. 
2. Then you slept for some time. The drone was entering "hover mode" automatically (in `OffboardApi::update()`) and the `goal_` was set to (x0, y0, z0) and the `goal_mode_` was set to "PositionMode".
3. Then, you set the drone to another position (x1,y1,z1) and then hover, the drone was immediately moved to the point you set, but the `hover` command still did nothing. However, the `goal_` and `goal_mode_` in `OffboardApi` had NEVER been changed! That is why the drone always tries to go back to (x0, y0, z0).
4. The rest is the same as 3.

In short, in your case, the `simSetVehiclePose` and `hover` command NEVER commit any offboard API, the `goal_` was set to the "current position" at the time entering hover mode (that is, when you call `sleep` first time).

To solve this problem, I think we should invoke AT LEAST ONE control command in `moveOnPath` function even the drone was already at the destination (@seheevic also mentioned this problem [here](https://github.com/microsoft/AirSim/issues/1677#issuecomment-553697308)). 

My fixes are committed to this PR, and you can test it with your example. The above reason may be the real cause of the losing-physics problem, so I completely changed my previous submit and rebased all changes. This patch may influence the behavior of `hover`, `moveToZ`, and `moveToPosition` commands.

laurelkeys:
Thanks for the detailed explanation @songxiaocheng !

AirLib/src/vehicles/multirotor/api/MultirotorApiBase.cpp
```diff
@@ -269,7 +269,7 @@ bool MultirotorApiBase::moveOnPath(const vector<Vector3r>& path, float velocity,
     float goal_dist = 0;
 
     //until we are at the end of the path (last seg is always zero size)
-    while (!waiter.isTimeout() && (next_path_loc.seg_index < path_segs.size()-1 || goal_dist > 0)
+    while (!waiter.isTimeout() && (next_path_loc.seg_index < path_segs.size()-1 || goal_dist >= 0)
```
zimmy87, line 272: can you explain the purpose of changing the last conditional to "goal_dist >= 0"?

zimmy87:
Thank you @songxiaocheng for submitting this pull request. I have one open comment I'd like to be addressed before merging.

jonyMarino:
Fixes #1246
Fixes #3049

zimmy87:
The bulk of the PR was merged in #3363, with the exception of the one item with an open question

AirLib/src/vehicles/multirotor/api/MultirotorApiBase.cpp
songxiaocheng: Sorry for my late reply. I made this change in an attempt to resolve the unexpected behavior mentioned above. The detail is in [my comment](https://github.com/microsoft/AirSim/pull/2324#issuecomment-606686250) above. When calling `hover` during the drone is motionless, it invokes `moveOnPath` ultimately, where I think both `next_path_loc.seg_index` and `path_segs.size()-1` are 1. Thus it will never jump in the `while` block with `goal_dist > 0`.

zimmy87: I tried running @laurelkeys's script with `goal_dist > 0` and `goal_dist >= 0` and I didn't see any change in behavior(the drone still moves toward (0,0,0) after each simSetVehiclePose() call), so I'm not sure this is fully addressing that issue

PR #2326, opened 2019-11-27
Fix print statement for Python 3 

ceroytres:
Fix print statement for Python 3 in Docker build script.


madratman:
python 2 is deprecated, and therefore we don't need to support it with the future import line.

madratman:
merged manually in #2471

PR #2333, opened 2019-12-06
Fix comparing cmake version in setup.sh

rajat2004:
`which` if the command is not present doesn't give empty string, instead says "not found"
Fixes the detection of required tools, plus sets the correct path of CMake on OSX

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -47,12 +47,13 @@ if $gccBuild; then
     # variable for build output
     build_dir=build_gcc_debug
     # gcc tools
-    gcc_ver=$(gcc -dumpfullversion)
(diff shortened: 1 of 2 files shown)


rajat2004:
Noticed while doing a fresh setup on Ubuntu 18.04 where default CMake version is 3.10.2

rajat2004:
The last commit just moves the block of code for finding the CMake binary since earlier, it was being executed only if the platform was Linux, thus disabling finding the binary on OSX

madratman:
really good catch. `setup.sh` is wasting a lot of unnecessary time in any machine which already has the up to date versions of these deps.

PR #2347, opened 2019-12-16
ArduCopterApi: Improvements & Fixes

rajat2004:


AirLib/include/vehicles/multirotor/firmwares/arducopter/ArduCopterApi.hpp
```diff
@@ -346,21 +346,21 @@ class ArduCopterApi : public MultirotorApiBase {
...
-                           "\"angular_velocity\": [%lf, %lf, %lf],"
-                           "\"linear_acceleration\": [%lf, %lf, %lf]"
+                           "\"angular_velocity\": [%.12f, %.12f, %.12f],"
+                           "\"linear_acceleration\": [%.12f, %.12f, %.12f]"
                            "},"
                            "\"gps\": {"
-                           "\"lat\": %lf,"
-                           "\"lon\": %lf,"
-                           "\"alt\": %lf"
+                           "\"lat\": %.7f,"
```
tridge: i'd suggest a few more digits for the GPS pos

tridge: actually, 7 is fine, it ends up with cm level, which is typical precision for a GPS, so it's fine

tridge (approved):

AirLib/include/vehicles/multirotor/firmwares/arducopter/ArduCopterApi.hpp
rajat2004: I had kept it at 7 decimal digits since it's being converted into int after being multiplied by 10^7

sytelus:
@madratman - this looks good. Do you want to merge it?

madratman:
LGTM. Breaking into a setupGenericQuad is fine as it is only effecting the arducopter api header file. 
In general, that piece of code is duplicated in AirSim

PR #2356, opened 2019-12-20
Rework travis.yml

rajat2004:
Remove docker build, add Ubuntu 18.04, 16.04, and Windows, macOS builds using `build.sh` as well as `Unity/build.sh`

This should help in catching breaking changes on different OSes, as well as on Unreal vs Unity
Unfortunately adding Unreal inside Travis doesn't seem possible

Should GCC build be added also?
Unity compilation throws about 340 warnings, which exceeds the Travis output log size (4 MB), seems like incorrect flags are being applied to the Unity build (Created issue https://github.com/microsoft/AirSim/issues/2478 for discussion)

Needs #2333 and #2328 (if only #2333 is there, CMake won't be downloaded when it's not required and without the fix of build.sh in #2328, cmake won't be found)

### DIFF
--- a/Unity/build.sh
+++ b/Unity/build.sh
@@ -9,8 +9,13 @@ fi
 
 
 cd linux-build;
-export CC="clang-5.0"
-export CXX="clang++-5.0"
+if [[ -n $CIINSTALL ]]; then # use downloaded binaries on Travis
+    export CC=${LLVM_DIR}/bin/clang
+    export CXX=${LLVM_DIR}/bin/clang++
+else
+    export CC="clang-5.0"
+    export CXX="clang++-5.0"
+fi
 
 if [ "$(uname)" == "Darwin" ]; then
     CMAKE="$(greadlink -f ../../cmake_build/bin/cmake)"
@@ -25,7 +30,7 @@ if [ ! -d "../UnityDemo/Assets/Plugins" ]; then
 fi
 
 if [ "$(uname)" == "Darwin" ]; then
-	cp AirsimWrapper.bundle ../UnityDemo/Assets/Plugins;
+	cp -r AirsimWrapper.bundle ../UnityDemo/Assets/Plugins;
 else
 	cp libAirsimWrapper.so ../UnityDemo/Assets/Plugins;
 fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,73 +1,56 @@
 language: cpp
-#git:
-# shorten this if we can nail down submodule depth
-#  depth: 3
 
-fast_finish: false
-
-matrix:
+jobs:
   allow_failures:
     - env: ALLOW_FAILURES=true
 
   include:
-    # linux version have unique dependencies, so we set them up individually
-    - os: linux
-      dist: trusty
-      sudo: required
-      services:
-        - docker
-    #- os: windows
-    # dotnet: 2.1.5
-    # services:
-    #   - docker
-        
+    - name: Ubuntu 18.04
+      os: linux
+      dist: bionic
 
-    # OS X CMake
-    #- os: osx
-    #  sudo: required
-    #  osx_image: xcode8.2
-    #  compiler:
-    #    - clang
-    #  env:
-    #    - TOOL="cmake"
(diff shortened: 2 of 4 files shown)


rajat2004:
Windows build is working now, build with this commit on my Travis page: https://travis-ci.com/rajat2004/AirSim/builds/142728157

rajat2004:
I've been trying to get MacOS build working, but having problems with that, branch is here if anyone's interested - https://github.com/rajat2004/AirSim/tree/travis-rework-macos
Travis build for branches can be seen here - https://travis-ci.com/rajat2004/AirSim/branches

rajat2004:
Any repo owner needs to look into starting the Travis builds though

rajat2004:
Build on Ubuntu fails with maximum log length error due to lots and lots of compilation warnings
Needs https://github.com/microsoft/AirSim/pull/2325
The current build of branch - https://travis-ci.com/rajat2004/AirSim/builds/143132453

rajat2004:
OSX build will also need https://github.com/microsoft/AirSim/pull/2333

rajat2004:
Travis builds are running for the PRs, but are not showing up in the checks

madratman:
lgtm on the first part, quite some unnecessary if else's in the original file.

madratman:
I think you shared links wrt a couple of PRs running this in your fork somewhere? Can you share them again?

madratman:
I think you shared links wrt a couple of PRs running this in your fork somewhere? Can you share them again?

rajat2004:
https://travis-ci.com/github/rajat2004/AirSim/builds/154715324
This is a build with my 2 test branches, should be quite up-to-date since this was run 3 days ago, with the Unity fixes included (since this is a PR build against the Travis branch)
Will put links of other builds as well, and rebase this one so that the latest master can be tested

madratman:
yeah, I somehow missed the links on the top of the page - hence deleted my comment. thanks though

rajat2004:
Don't know why Travis build isn't appearing on this PR, current build of master with the branch - https://travis-ci.com/github/rajat2004/AirSim/builds/155782696
Also deleted my old comments since were quite out of date

Unity/build.sh
```diff
@@ -9,8 +9,13 @@ fi
...
 cd linux-build;
-export CC="clang-5.0"
-export CXX="clang++-5.0"
+if [[ -n $CIINSTALL ]]; then # use downloaded binaries on Travis
+    export CC=${LLVM_DIR}/bin/clang
+    export CXX=${LLVM_DIR}/bin/clang++
+else
+    export CC="clang-5.0"
+    export CXX="clang++-5.0"
+fi
```
madratman: do we know which version of clang these pre-built binaries are? 
does travis allow for choosing a version of pre-built binaries?

In general, I am planning to remove the custom libcxx in airsim's setup.sh scripts. 
https://docs.unrealengine.com/en-US/Platforms/Linux/GettingStarted/index.html if one uses the clang version with the right unreal version as specified in that page, a custom libcxx build can be obviated. 

hence, curious to know which version are these pre built binaries and is there an option to select them

rajat2004: I've specified the binary version here - https://github.com/microsoft/AirSim/pull/2356/files#diff-354f30a63fb0907d4ad57269548329e3R27

Removing the need for custom libcxx would be good

madratman: ah cool, that's exactly the kind of line I was trying to find. 
PR is looking good to me. 
I am not sure if travis has a weird edge case when one updates the yml itself? However, I would be surprised if that isn't handled by them

rajat2004: It works even when `travis.yml` is updated, might be a problem with AirSim itself, have noticed that the build doesn't appear on the other PRs also

rajat2004:
Wow, great, thanks for merging!

Unity/build.sh
madratman: really weird. accio @sytelus, if I may.

rajat2004: @madratman @sytelus Were you able to figure out why the Travis build isn't working? It's still not triggering on the PRs

Issue #2345, opened 2019-12-12
Creating a new vehicle with mass bigger than 3kg


sebbom:
Hello AirSim community,

I am in the process of integrating new vehicles in AirSim but I am wondering if it is actually even possible to have heavier platform.
As stated in: [https://github.com/microsoft/AirSim/blob/0c8dc72ed293ac3fc9445bbc41cec3044fdb523b/AirLib/include/vehicles/multirotor/firmwares/simple_flight/SimpleFlightQuadXParams.hpp#L42 ](https://github.com/microsoft/AirSim/blob/0c8dc72ed293ac3fc9445bbc41cec3044fdb523b/AirLib/include/vehicles/multirotor/firmwares/simple_flight/SimpleFlightQuadXParams.hpp#L42)
it seems that there is a limit in a potential mass, is it only due to the specific simpleflight flight control or because of the physics engine?

I have been trying without success to integrate other flight controller tuned for heavier platform without any success and start to consider that the physics engine is specifically tuned for lighter ones.

Thanks in advance,

marcociara379:
I can confirm the bug, and I think it relates to [my issue #2341](https://github.com/microsoft/AirSim/issues/2341).

ibrahim0v:
@sebbom Hey,
Can you give more details about what you are trying to do? I have integrated my 8kg 35+Nm torque drone with great physical reactions (well, except the collisions and elasticity and stuff like that) but it is working "okay" now

jonyMarino:
Closed due to age and incompleteness.
Please feel free to open a new issue if you still have this problem, and include the following (as requested when opening a bug issue)

*Make sure to write all reproduction steps
*Include full error message in text form
*Search issues for error message before filing issue
*Attach screenshot if applicable
*Include code to run if applicable

Thanks!

Issue #2341, opened 2019-12-11
normalizedForce is compared to normalizedGravity, is not a physical nonsense?


marcociara379:
Hello everyone.
Looking at [FastPhysicsEngine.hpp](https://github.com/microsoft/AirSim/blob/master/AirLib/include/physics/FastPhysicsEngine.hpp#L330) it is possible to see that the aircraft is set to not grounded (`body.setGrounded(false)`) whenever the normalizedForce (squared norm of body_wrench.force) is greater or equal the normalizedGravity. Why is that? They are different physical quantities. Should not it be something like 
```
if (normalizedForce/body.getMass()/body.getMass() >= normalizedGravity)
[...]
```

Thanks,
Marco

madratman:
https://github.com/microsoft/AirSim/pull/2473

PR #2357, opened 2019-12-24
comparison force vs gravity fixed in FastPhysicsEngine.hpp

marcociara379:
This pull request fixes issue #2345 and #2341 
In `FastPhysicsEngine.hpp` the squared norm of the gravitational acceleration was compared to body wrench squared norm, leading to improper physical behavior for thrusts and masses different from the original quadcopter in AirSim, since it is not much different from 1 kg, thus hiding the issue. 
With this pull request, the gravitational acceleration is compared to the specific bodywrench squared, leading to proper physical behavior.


madratman:
good catch! closing in favor of https://github.com/microsoft/AirSim/pull/2473. (I don't the link the double division by mass. "weight" is more readable)

PR #2368, opened 2020-01-08
Bug fix; Connect to remote server;

xuhao1:
Fix bugs on angular velocity units of ros package;
Use _hard timestamp_ from simulation instead of soft timestamp
Add host IP options  make this can use ros package on remote device like a Linux machine while the simulation is on Windows


madratman:
LGTM, thanks! The IMU msg for certainly a todo I didn't finish. The next goal was to add have some VIO demos in AirSim over ROS.

PR #2172, opened 2019-09-01
Restructure of Car Vehicle for external firmware control

rajat2004:
This PR restructures the Car vehicle code to make it similar to Multirotor in terms of adding support for external firmware control of the car vehicle
An example of how to add external firmware control can be seen here - https://github.com/microsoft/AirSim/commit/a5dce95d82506741412f26072e3d4b7e6f20cdd9 (ArduPilot's Rover vehicle, the entire branch is https://github.com/microsoft/AirSim/compare/master...rajat2004:pr-ardurover2)
(Didn't open PR for the entire thing since would be more difficult to review)

This PR has been tested to be working with manual control from the keyboard on Ubuntu 16.04 and Windows 10, with the below Python APIs on Ubuntu *(Unreal Engine)*

Things tested to be working - 
* Movement APIs - `hello_car.py`
* getCarState - `car_monitor.py`
* Pause - `pause_continue_car.py`
* Multi-vehicle - `multi_agent_car.py`
* ClockSpeed

Unity - Compilation only till now, real testing soon to follow

There are a few more things which can be cleaned up more, such as removing the `CarState` and `CarControls` from `CarApiBase`
Feedback on the approach by the other devs is very much needed!

PR #2383, opened 2020-01-15
ArduRover Vehicle support

rajat2004:
Add support for ArduPilot's Rover vehicle in SITL (https://ardupilot.org/rover/)
Based on https://github.com/microsoft/AirSim/pull/2172 which restructures Car vehicle in AirSim (First 2 commits)

Usage instructions here - https://ardupilot.org/dev/docs/sitl-with-airsim.html#launch-rover-sitl

### DIFF
--- a/AirLib/include/vehicles/car/firmwares/ardurover/ArduRoverApi.hpp
+++ b/AirLib/include/vehicles/car/firmwares/ardurover/ArduRoverApi.hpp
@@ -1,0 +1,310 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#ifndef msr_airlib_ArduRoverCarController_hpp
+#define msr_airlib_ArduRoverCarController_hpp
+
+#include "vehicles/car/api/CarApiBase.hpp"
+#include "sensors/SensorCollection.hpp"
+
+#include "common/Common.hpp"
+#include "common/AirSimSettings.hpp"
+
+// Sensors
+#include "sensors/imu/ImuBase.hpp"
+#include "sensors/gps/GpsBase.hpp"
+#include "sensors/magnetometer/MagnetometerBase.hpp"
+#include "sensors/barometer/BarometerBase.hpp"
+#include "sensors/lidar/LidarBase.hpp"
+
+#include "UdpSocket.hpp"
+
+#include <sstream>
+#include <iostream>
+
+#define __STDC_FORMAT_MACROS
+#include <inttypes.h>
+
+namespace msr { namespace airlib {
+
+class ArduRoverApi : public CarApiBase {
+
+public:
+    ArduRoverApi(const AirSimSettings::VehicleSetting* vehicle_setting, std::shared_ptr<SensorFactory> sensor_factory,
+                 const Kinematics::State& state, const Environment& environment, const msr::airlib::GeoPoint& home_geopoint)
+    : CarApiBase(vehicle_setting, sensor_factory, state, environment),
+      state_(state), home_geopoint_(home_geopoint)
+    {
+        connection_info_ = static_cast<const AirSimSettings::MavLinkVehicleSetting*>(vehicle_setting)->connection_info;
+        sensors_ = &getSensors();
+
+        connect();
+    }
+
+    ~ArduRoverApi()
+    {
+        closeConnections();
+    }
+
+protected:
+    virtual void resetImplementation() override
+    {
+        CarApiBase::resetImplementation();
+    }
+
+public:
+    // Update sensor data & send to Ardupilot
+    virtual void update() override
+    {
+        CarApiBase::update();
+
+        sendSensors();
+        recvRoverControl();
+    }
+
+    virtual const SensorCollection& getSensors() const override
+    {
+        return CarApiBase::getSensors();
+    }
+
+    // TODO: VehicleApiBase implementation
+    virtual bool isApiControlEnabled() const override
+    {
+        // Return true so that CarPawnSim gets control message from external firmware and not keyboard
+        return true;
+    }
+
+    virtual void enableApiControl(bool is_enabled) override
+    {
+        Utils::log("enableApiControl() - Not Implemented", Utils::kLogLevelInfo);
+        unused(is_enabled);
+    }
+
+    virtual bool armDisarm(bool arm) override
+    {
+        Utils::log("armDisarm() - Not Implemented", Utils::kLogLevelInfo);
+        unused(arm);
+        return false;
+    }
+
+    virtual GeoPoint getHomeGeoPoint() const override
+    {
+        return home_geopoint_;
+    }
+
+    virtual void getStatusMessages(std::vector<std::string>& messages) override
+    {
+        unused(messages);
+    }
+
+public:
+    virtual void setCarControls(const CarControls& controls) override
+    {
+        // Currently ArduPilot vehicles don't implement AirSim movement APIs
+        unused(controls);
+    }
+
+    virtual void updateCarState(const CarState& car_state) override
+    {
+        last_car_state_ = car_state;
+    }
+
+    virtual const CarState& getCarState() const override
+    {
+        return last_car_state_;
+    }
+
+    virtual const CarControls& getCarControls() const override
+    {
+        return last_controls_;
+    }
+
+protected:
+    void closeConnections()
+    {
+        if (udpSocket_ != nullptr)
+            udpSocket_->close();
+    }
+
+    void connect()
+    {
+        port_ = static_cast<uint16_t>(connection_info_.udp_port);
+        ip_ = connection_info_.udp_address;
+
+        closeConnections();
+
+        if (ip_ == "") {
+            throw std::invalid_argument("UdpIp setting is invalid.");
+        }
+
+        if (port_ == 0) {
+            throw std::invalid_argument("UdpPort setting has an invalid value.");
+        }
+
+        Utils::log(Utils::stringf("Using UDP port %d, local IP %s, remote IP %s for sending sensor data", port_, connection_info_.local_host_ip.c_str(), ip_.c_str()), Utils::kLogLevelInfo);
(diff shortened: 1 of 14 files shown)


rajat2004:
@madratman Rebased this on master

rajat2004:
@madratman Renamed `phys_api_` to `pawn_api`, will soon do the other refactorings to reduce inconsistencies between Multirotor and Car
Upto you whether you prefer to merge this PR as is, and then have the others as a follow-up PR or have them all together(which I think might make it difficult to review and introduce much more number of changes than desired in a single PR)

AirLib/include/vehicles/car/firmwares/ardurover/ArduRoverApi.hpp
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    void sendSensors()
+    {
+        if (sensors_ == nullptr)
+            return;
+
+        const auto& gps_output = getGps()->getOutput();
+        const auto& imu_output = getImu()->getOutput();
+        // const auto& baro_output = getBarometer()->getOutput();
```
madratman: not a big deal, but might be possible to use `VehicleApiBase::getImuData(const std::string& imu_name()` directly? Same for other sensors? 
I think there shouldn't be an edge case where gps and imu are not there by default, as we're not error checking in getGps() and getImu() (via something like `uint count_imus = getSensors().size(SensorBase::SensorType::Imu);`).

madratman:
I agree let's get this in first

AirLib/include/vehicles/car/firmwares/ardurover/ArduRoverApi.hpp
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    virtual GeoPoint getHomeGeoPoint() const override
+    {
+        return home_geopoint_;
+    }
+
+    virtual void getStatusMessages(std::vector<std::string>&) override
+    {
+    	// Commented out so that terminal is not spammed with messages
```
madratman: minor: fix whitespace

AirLib/include/vehicles/car/firmwares/ardurover/ArduRoverApi.hpp
```diff
@@ -0,0 +1,309 @@
...
+    AirSimSettings::MavLinkConnectionInfo connection_info_;
+
+    std::shared_ptr<mavlinkcom::UdpSocket> udpSocket_;
+
+    uint16_t port;
+    std::string ip;
+
+    const SensorCollection* sensors_;
+
+    CarControls lastCarControls_;
```
madratman: let's maintain consistency w.r.t naming style of member variables

`port_`, `ip_`, `last_car_controls_`, `last_rc_data_`

madratman: there's another inconsistency b/w arduroverapi.hpp and physxcarapi.hpp
why is a diff member variable, `lastrcdata` being set in `setCarControls()` method? Shouldn't it be a single `last_controls_` a la physxcarapi?

madratman: Am I missing something? Not sure where lastRCData_ plays a role anywhere else in the code? 
Is the update() sequence different from physxcarapi?

AirLib/include/vehicles/car/firmwares/ardurover/ArduRoverApi.hpp
rajat2004: Removed, makes it consistent with the ArduCopterApi code

AirLib/include/vehicles/car/firmwares/ardurover/ArduRoverApi.hpp
rajat2004: Fixed the naming of the variables
You are correct, the `lastRCData_` is not being used anywhere now. It was there since I earlier had a piece of code which was using it to send the RC data received by AirSim(from the wheel, keyboard or actual RC connected) to ArduPilot, but it actually wasn't ever going to happen since the `return true` [here](https://github.com/microsoft/AirSim/pull/2383/files#diff-853eb839e4b1471419ff881432a324acR74) actually disables taking the input from those external devices, so removed it and then this variable was left

AirLib/include/vehicles/car/firmwares/ardurover/ArduRoverApi.hpp
rajat2004: This is a problem in the implementations of PX4, ArduCopter, etc. And therefore has been copied everywhere
Can be done, but I think would be better to do it for all the vehicles together

madratman: Ah yes, true. Makes sense
Alright, thanks for taking care of the other two fixes. I'm happy to merge it in. Nice work!

AirLib/include/vehicles/car/firmwares/ardurover/ArduRoverApi.hpp
```diff
@@ -0,0 +1,307 @@
...
+        sendSensors();
+        recvRoverControl();
+    }
+
+    virtual const SensorCollection& getSensors() const override
+    {
+        return CarApiBase::getSensors();
+    }
+
+    // TODO:VehicleApiBase implementation
```
madratman: what does this todo comment refer to?

AirLib/include/vehicles/car/firmwares/ardurover/ArduRoverApi.hpp
```diff
@@ -0,0 +1,307 @@
...
+#include "common/AirSimSettings.hpp"
+
+// Sensors
+#include "sensors/imu/ImuBase.hpp"
+#include "sensors/gps/GpsBase.hpp"
+#include "sensors/magnetometer/MagnetometerBase.hpp"
+#include "sensors/barometer/BarometerBase.hpp"
+#include "sensors/lidar/LidarBase.hpp"
+
+#include "UdpSocket.hpp"
```
madratman: perhaps better to specify relative path to mavlinkcom/udpsocket header file.

AirLib/include/vehicles/car/firmwares/ardurover/ArduRoverApi.hpp
rajat2004: To implement the `armDisarm` and other API functions below it

AirLib/include/vehicles/car/firmwares/ardurover/ArduRoverApi.hpp
```diff
@@ -0,0 +1,307 @@
...
+        Utils::log("armDisarm() - Not Implemented", Utils::kLogLevelInfo);
+        return false;
+    }
+
+    virtual GeoPoint getHomeGeoPoint() const override
+    {
+        return home_geopoint_;
+    }
+
+    virtual void getStatusMessages(std::vector<std::string>&) override
```
madratman: should be     virtual void getStatusMessages(std::vector<std::string>& messages) override?

AirLib/include/vehicles/car/firmwares/ardurover/ArduRoverApi.hpp
rajat2004: That causes errors in the Unreal packaging, all the Mavlinkcom headers are present in the "include" folder
This is the way the Mavlinkcom header files are included everywhere, since the relative paths change when stored in the Unreal folder

madratman: ah yes, that's correct

AirLib/include/vehicles/car/firmwares/ardurover/ArduRoverApi.hpp
rajat2004: Specifying the variable name `messages` then requires `unused` in the function for compilation on Windows
Easier way to avoid that, though can add that back if you prefer. Such a usage is present everywhere, will change all that in that case

AirLib/include/vehicles/car/firmwares/ardurover/ArduRoverApi.hpp
madratman: ah, dug some more. this is coz sensor APIs were added in https://github.com/microsoft/AirSim/pull/1920, much later than ArduCopterSoloApi.hpp / MavLinkMultirotorApi .hpp

AirLib/include/vehicles/car/firmwares/ardurover/ArduRoverApi.hpp
madratman: yes, let's stick to unused for now, to maintain consistency

madratman: yep, it's better to make an unused workaround everywhere in the code in a single commit / pr

AirLib/include/vehicles/car/firmwares/ardurover/ArduRoverApi.hpp
rajat2004: Makes sense, now these functions can be used directly
Should reduce code duplication also across the different Firmwares

AirLib/include/vehicles/car/firmwares/ardurover/ArduRoverApi.hpp
rajat2004: Ahh, missed this last comment. I've updated the ArduRover file to use `unused`, the others can be done in a different PR

madratman:
All comments resolved, thanks! Nice work, @rajat2004.

rajat2004:
Ahh great, thanks for the in-depth review and merging @madratman !
Could you have a look at https://github.com/microsoft/AirSim/pull/2356, this enabled Travis builds on Linux, Windows and MacOS, and should help prevent compilation problems on different platforms being introduced(which actually happened after the ArduCopter PR, due to the `unused` function being missing)

PR #2400, opened 2020-01-28
Typo fix

piyushchauhan:


madratman:
there are multiple json files in there. hence the plural. we can rewrite as "`settings.json` files"

PR #2402, opened 2020-01-30
UE4.24: Fixes compiling errors in combination with Unreal Engine 4.24.x

### LINKED ISSUE #2348: UE4 4.24 can not support Airsim,how to solve
# Read This First

## If you are reporting a bug
* Make sure to write **all reproduction steps**
* Include full error message in text form
* Search issues for error message before filing issue
* Attach screenshot if applicable
* Include code to run if applicable

## If you have question
* Add clear and concise title
* Add OS, AirSim version, Python version, Unreal version if applicable
* Include context on what you are trying to achieve
* Include details of what you already did to find answers

**What's better than filing issue? Filing a pull request :).**

------------------------------------ (Remove above before filing the issue) ------------------------------------

LSBOSS:
The following was done to fix https://github.com/microsoft/AirSim/issues/2348 

- Changing first entry of ECameraDirectorMode in CameraDirector.h to index 0 instead of 1
- Adjusting initial_view_mode in AirSimSettings.hpp accordingly 
- Modified a couple of include pathes (e.g. #include "ConstructorHelpers.h" vs. #include "UObject/ConstructorHelpers.h")


rajat2004:
PR does work with 4.18 as well
Some documentation and examples changes might also be needed

LSBOSS:
> PR does work with 4.18 as well
> Some documentation and examples changes might also be needed

Hi @rajat2004,

can you please elaborate more on this? 

I see mentioning of the exact Unreal engine version in docs/build_linux.md, docs/build_windows.md and docs/docker_ubuntu.md. I built UE4.24 with my patch on both Windows and Linux and under Linux with ue4-docker as well. 

So I'd say it's safe to bump this value from 4.18 to 4.24? 

@Examples: I took a quick look at some examples and don't see how they are affected by my patch. Maybe you can enlighten me here :) 

Thank you very much

rajat2004:
@LSBOSS Hi, sorry for the misleading comment, shouldn't do such things while drowsy :)

What I meant was, after the changes in the PR, it still works with UE4 4.18 as well. If it were not the case, i.e. it would break support for 4.18, then a new branch for 4.24 might be needed, or update to the entire repo which would come with a lot of compatibility headaches and support issues
Since it works with 4.18, it should be okay to merge into master IMO


Regarding the examples, mistakenly, I thought this would affect camera recordings as well which is not the case, so no change should be needed there
Will post or make a PR against your branch :) if any documentation changes are required, but don't think that's the case either

Hope this clears up any confusion!

mflotynski:
Is there any reason why this PR hasn't been merged yet? 

BTW. Thanks @LSBOSS for the patch :bowing_man:

madratman:
@LSBOSS thanks for the patch! Works well with old UE versions on linux and windows. 
Apologies everyone, we have been occupied with other internal projects, but we're ramping up with PRs on github now. Thanks for your patience.

PR #2453, opened 2020-03-10
Fast image capture

ironclownfish:
Improves the speed of image capture.
-Locks the render target, reads the capture, then unlocks, instead of using ReadPixels().
-Removed some unneeded complexity, like PNG compression and float-valued rgb
-Removed extra mallocs/copies
-Attempt to simplify the code, reduce looping and branching.

More speedup is possible by using a buffer pool for the dest when reading the render target. This will follow in a future commit.


ironclownfish:
This has been superceded by #2472 2472

PR #2482, opened 2020-03-26
Added ability to get position vertex buffer of static meshes from the unreal scene

madratman:
Original PR by @nikonodein in #1478. Reposting their description.  

This commit does work for my purposes. However there may be some improvements we want to discuss before merging it in to master.

1) Meshresponse struct:
- Do we want to contain the vertices in seperate X,Y,Z containers?

```
//Right now
struct MeshResponse {
    std::vector<float> vertices; //size is num of vertices *3
}

//Change 

struct MeshResponse {
    std::vector<float> x_vertices;
    std::vector<float> y_vertices;
    std::vector<float> z_vertices;
}
```

2) Coordinate system:
- The meshes coordinate are in the Unreal coordinate system. In this code I just pipe that data directly to the MeshResponse. 
- Converting it to NED is tricky cause then you have to specify a vehicle, which I believe this function should not require.
- For my purposes I just deal with the conversions on the client side, however, I also have the extra functionality in my codebase to get the offset between the NED and the unreal origin.

3) Efficiency of function:
- The function is quite inefficient cause it calls a FlushRenderingCommands() for every single Mesh.
- This to me doesn't actually matter that much, because you should never be calling this function that often. We only call this function once at the start of our main loop, and then never do it again.
- Could be improved by preallocating the buffers, and then call FlushRenderingCommands() once, but in the grand scheme of things doesn't really matter.

4) Meshes to ignore by default:
- By default I ignore the skybox mesh and the camera meshes, but we might not want to do that.

Example of what I receive from this function. Note I ignore the ground plane meshes.
mesh_responses

### DIFF
--- a/AirLib/include/api/RpcLibAdapatorsBase.hpp
+++ b/AirLib/include/api/RpcLibAdapatorsBase.hpp
@@ -694,6 +694,70 @@ public:
             return d;
         }
     };
+
+	struct MeshResponse {
+		Vector3r position;
+		Quaternionr orientation;
+
+		std::vector<float> vertices;
+		std::vector<uint32_t> indices;
+		std::string name;
+
+		MSGPACK_DEFINE_MAP(position, orientation, vertices, indices, name);
+
+		MeshResponse()
+		{}
+
+		MeshResponse(const msr::airlib::MeshResponse& s)
+		{
+			position = Vector3r(s.position);
+			orientation = Quaternionr(s.orientation);
+
+			vertices = s.vertices;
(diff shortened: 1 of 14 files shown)


madratman:
Fwiw, one can also extend this PR to expose additional buffers (apart from position) under [FStaticMeshVertexBuffers](https://docs.unrealengine.com/en-US/API/Runtime/Engine/FStaticMeshVertexBuffers/index.html)  - color buffer, texture coord, tangents.

PR #2483, opened 2020-03-27
ROS tf2 - Stop rebroadcasting static frames

plusk01:
The `tf2` API in ROS creates two topics for transforms: `/tf` and `/tf_static`. The `/tf_static` topic is for transforms that do not change. Having two topics lets `/tf_static` leverage topic latching to minimize bandwidth usage.

In the AirSim ROS wrapper, the two static tfs that are created are for camera/lidar w.r.t body frame and for the odom frame (named using the vehicle name). These are all loaded from the settings. This changes stops rebroadcasting these static transforms on every state update.


madratman:
good catch, thanks

PR #2486, opened 2020-03-27
Minor cosmetics

plusk01:
code readability and getting rid of compilation warnings


plusk01:
superseded by #2485

madratman:
Thanks for an almost-contribution :). If you're looking to contribute, I think one thing that I missed is adding a docstring for moveByMotorPWMs and mentioning explicitly inputs should be 0-1.

PR #2500, opened 2020-04-01
Linux build improvements - clang8, no custom libcxx 

madratman:
- Tested with UE4.24, Ubuntu 18 (might have to enforce 4.24, right now we support >=4.22)
  - or add option for clang-7 for UE 4.22
- install `clang-, libc++-8-dev, libc++abi-8-dev` in `setup.sh`
- remove all the clang5.0 and custom libcxx download and building
- in commonsetup.cmake, throwing in a `-stdlib=libc++` and removing all the unreal specific flags seems to the job

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -65,15 +65,8 @@ else
     # variable for build output
     build_dir=build_debug
     if [ "$(uname)" == "Darwin" ]; then
-
-        if [[ -n $CIINSTALL ]]; then # use downloaded binaries on Travis
-            export CC=${LLVM_DIR}/bin/clang
-            export CXX=${LLVM_DIR}/bin/clang++
-        else
-            export CC=/usr/local/opt/llvm-5.0/bin/clang-5.0
-            export CXX=/usr/local/opt/llvm-5.0/bin/clang++-5.0
-        fi
-
+        export CC=/usr/local/opt/llvm@8/bin/clang
+        export CXX=/usr/local/opt/llvm@8/bin/clang++
     else
         export CC="clang-8"
         export CXX="clang++-8"
--- a/build.sh
+++ b/build.sh
@@ -62,25 +62,6 @@ if $gccBuild; then
         export CXX="g++"
     fi
 else
-    #check for correct verion of llvm
-    if [[ ! -d "llvm-source-50" ]]; then
-        if [[ -d "llvm-source-39" ]]; then
-            echo "Hello there! We just upgraded AirSim to Unreal Engine 4.24."
-            echo "Here are few easy steps for upgrade so everything is new and shiny :)"
-            echo "https://github.com/Microsoft/AirSim/blob/master/docs/unreal_upgrade.md"
-            exit 1
-        else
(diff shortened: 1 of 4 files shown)


build.sh
```diff
@@ -94,8 +75,9 @@ else
         fi
 
     else
-        export CC="clang-5.0"
-        export CXX="clang++-5.0"
+        echo "RATNESHRATNESHRATNESHRATNESHRATNESHRATNESHRATNESHRATNESHRATNESHRATNESHRATNESHRATNESHRATNESH"
```
madratman, line 78: happens to the best of us

setup.sh
```diff
@@ -60,33 +51,22 @@ if $gccBuild; then
...
-
-            # brew install llvm@3.9
-            brew tap llvm-hs/homebrew-llvm
-            brew install llvm-5.0
-            export C_COMPILER=/usr/local/opt/llvm-5.0/bin/clang-5.0
-            export COMPILER=/usr/local/opt/llvm-5.0/bin/clang++-5.0
-        fi
+        # brew install llvm@3.9
+        brew tap llvm-hs/homebrew-llvm
+        brew install llvm-8.0
```
rajat2004, line 58: Installing from `brew` on Travis times out, since for some reason, it tries to build it. Did try different things, but nothing work
Will need to check if it's the case with `llvm-8` but if yes, then will need to add back the Travis specific lines set above and update the version in `travis.yml`

setup.sh
```diff
@@ -98,7 +78,7 @@ if [ "$(uname)" == "Darwin" ]; then # osx
     fi
 
     brew install wget
-    brew install coreutils
+    brew install coreutilsah ok so
```
rajat2004, line 81: Wait, are these for real!?

setup.sh
madratman: ah I removed them in haste as libcxx wasn't needed to be compiled

setup.sh
madratman: well, great. I am switching b/w DMs and unreal and vscode, and my machine is dumping text everywhere

rajat2004:
PRs working properly for me, manually removed the `llvm-build` directory, but no problems
Some new warnings are there, which is expected when upgrading to a new compiler
Travis build of the current PR - https://travis-ci.com/github/rajat2004/AirSim/builds/157408468

rajat2004:
@madratman I've opened https://github.com/madratman/AirSim/pull/1 against this branch on your fork, which updates the compiler for OSX
Travis build - https://travis-ci.com/github/rajat2004/AirSim/builds/157476649

rajat2004:
Also, what should be done with the Unity compilation? That is still using clang-5

madratman:
travis is working for upstream airsim as well - https://travis-ci.org/github/microsoft/AirSim/builds/669789574 / https://travis-ci.org/github/microsoft/AirSim/pull_requests
but for some reason, not showing up on gh PR page

PR #2506, opened 2020-04-01
Features: add an api for setting trace line color and thickness

songxiaocheng:
There is a little problem that the alpha channel is not working. I also tried to manually call `DrawDebugLine` but the trace lines are always opaque. It seems like a UE issue, so I kept this argument. For now, setting RGB and thickness works well.

### DIFF
--- a/Unreal/Plugins/AirSim/Source/PawnSimApi.h
+++ b/Unreal/Plugins/AirSim/Source/PawnSimApi.h
@@ -87,6 +87,7 @@ public: //implementation of VehicleSimApiBase
         return params_.vehicle_name;
     }
     virtual void toggleTrace() override;
+    virtual void setTraceLine(const std::vector<float>& color_rgba, float thickness) override;
 
     virtual void updateRenderedState(float dt) override;
     virtual void updateRendering(float dt) override;
@@ -189,4 +190,7 @@ private: //vars
 
     std::unique_ptr<msr::airlib::Kinematics> kinematics_;
     std::unique_ptr<msr::airlib::Environment> environment_;
+
+    FColor trace_color_ = FColor::Purple;
+    float trace_thickness_ = 3.0f;
 };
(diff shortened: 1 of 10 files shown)


Unreal/Plugins/AirSim/Source/PawnSimApi.h
```diff
@@ -188,4 +189,7 @@ class PawnSimApi : public msr::airlib::VehicleSimApiBase {
 
     std::unique_ptr<msr::airlib::Kinematics> kinematics_;
     std::unique_ptr<msr::airlib::Environment> environment_;
+
+    FColor trace_color_=FColor::Purple;
+    float trace_thickness_=3.0f;
```
madratman, line 194: please change to 
```cpp
FColor trace_color_ = FColor::Purple; 
float trace_thickness_ = 3.0f; 
```
for readability

madratman:
Can we maintain consistency w.r.t. simPlot APIs (those APIs ask for rgba list in python in 0-1 range) (see plot_markers.py)

the lines here might make alpha channel work https://github.com/microsoft/AirSim/pull/2304/files#diff-405ee0a8e5b40c6316073f72165e0228R177, 
https://github.com/microsoft/AirSim/pull/2304/files#diff-405ee0a8e5b40c6316073f72165e0228R170

songxiaocheng:
Hi @madratman, I've updated as you suggested. 
As for the alpha channel problem, the two lines code you mentioned is not working either.
Actually, in my test, the alpha channel is also failed to work for the `simPlot*` APIs. Do you have the same problem?

songxiaocheng:
Now the build passed.

songxiaocheng:
The Travis check status fails to update and this PR have already got passed.

madratman:
Unable to resolved the transparency, but apart from that, it works fine.  
Thanks for the contribution!

rajat2004:
What is the range of thickness here? 0-100? and is it in pixels

Issue #856, opened 2018-03-01
rotateToYaw(yaw) returns at yaw but drone rotates more before stopping


wagnboss:
I am using the Python API.
MultirotorClient class method rotateToYaw(yaw, max_wait_seconds, margin) always over shoots the target yaw.  If large enough max_wait_seconds argument is supplied, method return when yaw is within margin of target yaw, but drone continues to rotate for a while. For example, see following pyton code:
from AirSimClient import *
import time
import math

factor = 180./math.pi

def print_info():
    pitch, roll, yaw = client.getPitchRollYaw()
    print("pitch %5.1f, roll %5.1f, yaw %5.1f" % (pitch*factor, roll*factor, yaw*factor))
    print()
    
client = MultirotorClient()
client.confirmConnection()
client.enableApiControl(True)
client.armDisarm(True)

delay = 5
max_wait_seconds = delay

client.takeoff()
time.sleep(delay)
print_info()

print("rotateToYaw (90)")
client.rotateToYaw(90., max_wait_seconds, 1)
print_info()    # it reached 90
time.sleep(delay)
print_info()    # it continued rotating

-----------------------------
which produces the following output:
Waiting for connection: 
after takoff
pitch   0.0, roll   0.0, yaw   0.0

rotateToYaw (90)
pitch   0.0, roll   0.0, yaw  90.5

pitch   0.0, roll   0.0, yaw 114.4

-------------------------------
Is this is normal/expected/desired behavior? 
If so, how do I get the drone to STOP rotating when it reaches target yaw?  
I've tried adding the following line immediately after the call to rotateToYaw, but it does not work.

client.rotateByYawRate(0.,1)    #try to stop rotation

sytelus:
You might want to play around to turn PID controller. Overshoots might happen because p is too high for example. There is plenty of literature on web on PID tuning...

PID parameters: 
Play with rate and level PID parameters: https://github.com/Microsoft/AirSim/blob/master/AirLib/include/vehicles/multirotor/firmwares/simple_flight/firmware/Params.hpp#L51

Code: https://github.com/Microsoft/AirSim/blob/master/AirLib/include/vehicles/multirotor/firmwares/simple_flight/firmware/AngleLevelController.hpp

mwanggh:
I found this command to work when the drone is hovering. If you control the head direction in move, you may use the drivetrain parameter. 

client.hoverAsync().join()
client.rotateToYawAsync().join

[...]

PR #2516, opened 2020-04-02
Fix bug: rotateToYaw not working

### LINKED ISSUE #1682: rotateToYawAsync() does nothing
Hopefully I am just using this wrong, but when trying to rotate using rotateToYawAsync() nothing happens to the drone. I simply take off then call the rotate method.

client.rotateToYawAsync(90,5,1)

Has anyone had any luck using this. Is there a better way to have a drone turn in place? Is that possible?

songxiaocheng:
Fix #1682, #856

Changes to fix `rotateToYaw`:
- `isYawWithinMargin` should receive the target yaw degree as the first argument
- The original stop criterion does not stop rotation due to inertia, so the resulted yaw is not guaranteed.

By the way, the `rotateByYawRate` implementation were made consistent with other API.


madratman:
Works well. Thanks for the contribution.

PR #2520, opened 2020-04-03
Specify C++17 standard on macOS

saihv:
master fails to build on my machine (I'm on 10.15.4 with XCode 11.4) with complaints about std::filesystem, although Travis seems fine?

This PR contains the fix I had to use.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ before_install:
 script:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
       ./setup.sh;
-      ./build.sh;
+      ./build.sh || travis_terminate 1;
       echo "Starting Unity Build!";
       cd Unity && ./build.sh 2> Unity_error.txt || ((cat Unity_error.txt | grep -i "error:" -B 10 -A 10) && exit 1);
     elif [[ "$TRAVIS_OS_NAME" == "windows" ]]; then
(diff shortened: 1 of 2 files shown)


rajat2004:
Travis uses macOS 10.13 and Xcode 9.4.1 by default. The versions can be changed using `xcode_image` - https://docs.travis-ci.com/user/reference/osx/

rajat2004:
Also, the OSX build is failing - https://travis-ci.org/github/microsoft/AirSim/jobs/670441638#L6866
```
ld: library not found for -lc++fs
clang-8: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [output/bin/MavLinkTest] Error 1
```

As to why the build isn't showing up as failed, my thinking is the root `build.sh` exits on failure (`set -e`), but Travis only checks the return code of the last command being run, which would be the Unity build.
So the `build.sh` exits with code 1, Travis continues with the `echo` and Unity build, which is where things get confusing, there's no output after that, which might indicate some problem in the beginning of the script itself, since everything goes into the `stderr` rather than `stdout` and hence is captured in the file, but the word `error:` was not found, and so no output. But then why isn't `exit 1` giving a return code of 1??
In the Unity build command, `||` is used to check the exit code of the command on the left side and running the right only if the left one exits with a non-zero code. I've tested this on my local system with some simple bash scripts, so am not sure why it's behaving in this way

One addition to make Travis exit if the root `build.sh` exits with failure is to add `|| travis_terminate 1` to it, so that it doesn't continue to the Unity build

saihv:
Thanks! - I totally missed that. It definitely makes sense to differentiate between AirLib and Unity builds on Travis

saihv:
[Well, doesn't look like this is going to work on 10.14.](https://stackoverflow.com/questions/58667853/does-use-of-c17-stdfilesystem-require-macos-10-15-xcode-11-1)

rajat2004:
That's too bad, according to the Travis forums, [10.15 was supposed to come in December](https://travis-ci.community/t/macos-catalina-build-environment/5608/5)
In this case, then it should build locally on your system properly while failing to work on Travis
Should we keep the OSX build on Travis enabled then? Will need to update the docs to say it'll work on Catalina only

saihv:
Yep, it does build locally on 10.15. Thanks for the doc update, will merge both soon

PR #2522, opened 2020-04-04
Linux build improvements

saihv:
Placeholder PR for WIP build improvements on Linux.

Features:

* Add and default to release mode in `build.sh`, with ability to take a `--debug` flag for debug. 
* Add compiler optimizations when building as Release. (O2 or O3?)
* Add option to use precompiled AirLib in the UE project to preserve compiler optimizations. (Can we control this through build.sh as well?)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -308,6 +308,8 @@ cmake_build/
 /[Ee]igen
 /build_debug
 /build_gcc_debug
+/build_release
+/build_gcc_release
 /cmake/CPackSourceConfig.cmake
 /cmake/CPackConfig.cmake
 /cmake/arch.c
--- a/.gitignore
+++ b/.gitignore
@@ -307,9 +307,7 @@ cmake_build/
 /boost
 /[Ee]igen
 /build_debug
-/build_gcc_debug
 /build_release
-/build_gcc_release
 /cmake/CPackSourceConfig.cmake
 /cmake/CPackConfig.cmake
 /cmake/arch.c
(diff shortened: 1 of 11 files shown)


rajat2004:
Just some things which came to my mind, I haven't tested this

- Add `build_release`, `build_gcc_release` to clean.sh
- Should we add usage info to `build.sh`? Will be imp, once the args increase
- Will this make any difference in the packaging of the project?

saihv:
Packaging might actually be broken right now, regardless of this PR. I am not entirely sure why but the introduction of clang-8 is causing problems on the engine's side  (shadowed enum error inside Engine source code), problems that were fixed for 4.25 pending release. It would be great if you can try packaging Blocks using master!

https://github.com/EpicGames/UnrealEngine/commit/cb388710a7fbe43eaa82a6d8c43b1632f25f6386

Agree with the other two points. I will try to add them

rajat2004:
Oh, wasn't ware that packaging was broken. Then how were the release binaries packaged? Did that require manual application of the commit to the UE code and then build it?

Will try packaging Blocks, is this the case on both Linux and Windows?

saihv:
I believe it's a Linux-only problem, because of clang-8 (but I haven't tried it in windows).

The 1.3.0 releases were built with a version of master prior to the introduction of clang-8. Even now, it's possible to do a simple patch to the UE source code (the commit I posted above) and build packages on Linux.

madratman:
Packaging works for me, although I am using an unreal engine "installed build"
I did see a cooking error. I was able to get around that by hitting "launch" (which cooks the contents), and then making the binary. 
There's an explicit option to "cook contents" which also might solve the issue.

Issue #2379, opened 2020-01-14
Customising Python API for FPV camera FOV control - (+simSetCameraFov function)


ibrahim0v:
Hello @madratman I hope and think that you'll be the one to help me :)

I am currently working on a project where I need an controllable gimbal camera but with the additive of zoom function. I found out that I can zoom in and out of a view by altering field of view parameter of the camera. Therefore I have already done that in airSim. Now in my version, when FPV mode is entered with "F" key, the camera orientation can be controlled with WASDQE keys and its fov can be controlled with the arrow keys. Anyways, now I want to be able to do this using the API you provided. The camera orientation can already be configured using the **simSetCameraOrientation**.  
So I started working on declaring and using a function called **simSetCameraFov** that function as its name implies but couldn't reach a point yet. So please your help would be appreciated. I will be working on it and I would like to post a merge request for an API update + the manual fpv gimbal control so the community can get use of it.

ibrahim0v:
I am able to change FOV of the camera now using my simSetCameraFov function but something is updating the FOV back to settings.fov which is 90 degress by default. I cannot find the module that is reverting it back to 90deg. 

@clovett @sytelus @madratman Can someone please guide me

ibrahim0v:
I have solved it. Now I have successfully added simSetCameraFov function as api for FPV camera zoom control. I can post a pull request if anyone cares

saihv:
A pull request would be very welcome!

ibrahim0v:
@saihv okay I'll post a PR as soon as I finish up some cleaning in my code :+1:

alwynmathew:
@idanlevin what is the FOV you achieved? Will it be able to modify your setting to 180 degree fisheye camera?

ibrahim0v:
I have limited it between 150degs to 20degs but the limits are customisable. I have seen 180 and the upside down view for degs above 180 (upto 360 I guess)

alwynmathew:
@ibrahim0v Have you posted this PR yet? I need 180 degree Horizontal FOV, how would I do it? Any help would be appreciated.

ibrahim0v:
I'll post a PR ASAP :+1:

rajat2004:
https://github.com/microsoft/AirSim/pull/2534 has been merged which closes this

ibrahim0v:
thanks @rajat2004 I couldn't send my PR :) Yours is a better way anyways :+1:

Issue #1569, opened 2018-11-25
Zoom Camera / dynamic FOV change


petergerten:
Is there any way to change the camera FOV programmatically to simulate a zoom camera?
Something similar to simSetCameraOrientation but for FOV.

petergerten:
looks like it would have to be implemented in PIPCamera.cpp similar to the other settings that are updateable.

agnarhs:
I have been looking for information regarding how to dynamically change camera settings; exposure, zoom,  from the airsim API and then I found this year old thread. Do you know if there exist a solution now?

petergerten:
@agnarhs unfortunately not. I am still looking for a solution as well.

PR #2534, opened 2020-04-06
simSetCameraFov() API to control FoV dynamically

saihv:
Addresses feature requests in https://github.com/microsoft/AirSim/issues/1569, https://github.com/microsoft/AirSim/issues/2379 etc.

### DIFF
--- a/PythonClient/airsim/client.py
+++ b/PythonClient/airsim/client.py
@@ -155,6 +155,9 @@ class VehicleClient:
     def simSetCameraOrientation(self, camera_name, orientation, vehicle_name = ''):
         # TODO: below str() conversion is only needed for legacy reason and should be removed in future
         self.client.call('simSetCameraOrientation', str(camera_name), orientation, vehicle_name)
+        
+    def simSetCameraFov(self, camera_name, fov_degrees):
+        return self.client.call('simSetCameraFov', camera_name, fov_degrees)
 
     def simGetGroundTruthKinematics(self, vehicle_name = ''):
         kinematics_state = self.client.call('simGetGroundTruthKinematics', vehicle_name)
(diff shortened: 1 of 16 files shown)


rajat2004:
Unity build on Windows is failing, needs a non-implemented function in Unity as well to make it compile

PythonClient/airsim/client.py
```diff
@@ -155,6 +155,9 @@ def simGetCameraInfo(self, camera_name, vehicle_name = ''):
     def simSetCameraOrientation(self, camera_name, orientation, vehicle_name = ''):
         # TODO: below str() conversion is only needed for legacy reason and should be removed in future
         self.client.call('simSetCameraOrientation', str(camera_name), orientation, vehicle_name)
+
```
madratman, line 158: Please add a docstring! (We need to add docstring to all python apis soon)
I'm brewing up something a la https://microsoft.github.io/AirSim-NeurIPS2019-Drone-Racing/api.html post #2540 is in

Unreal/Plugins/AirSim/Source/PIPCamera.h
```diff
@@ -42,6 +42,7 @@ class AIRSIM_API APIPCamera : public ACameraActor
     bool getCameraTypeEnabled(ImageType type) const;
     void setupCameraFromSettings(const APIPCamera::CameraSetting& camera_setting, const NedTransform& ned_transform);
     void setCameraOrientation(const FRotator& rotator);
+    void setCameraFoV(const float& fov_degrees);
```
ironclownfish, line 45: Why const float& instead of float?

saihv: Just to maintain consistency (see other declarations)

madratman: passing by const ref (and const correctness) is a good thing. 
But, I don't think there's anything to be gained with const ref to a primitive type (int, float, char, etc). it might be actually more expensive

saihv: Ah okay, yes - it might be expensive because of dereference costs. I can just remove it

madratman: cool thanks. merging

Issue #2536, opened 2020-04-06
Failed to find /AirSim/Blueprints/BP_PIPCamera


plusk01:
**System**: Ubuntu 18.04 / 64 GB RAM / Intel i9 / GTX 1060 6 GB / SSD
**Versions**: UE4.22 / ROS Melodic / AirSim master
**Environment**: Blocks (ran from UE4 Editor)
**Settings**: default

After fast-forwarding the last ~50 commits (since about a week ago), running `./clean.sh` and then `./setup.bash` and `./build.sh` again, I try and open Blocks with UE4.22.3:

```bash
$ ./Engine/Binaries/Linux/UE4Editor $HOME/dev/AirSim/Unreal/Environments/Blocks/Blocks.uproject
```

I see the following error:

![image](https://user-images.githubusercontent.com/3934603/78598344-2813ea80-781d-11ea-8f52-d8ef4b5b2646.png)

saihv:
@plusk01 Thanks for finding the cause! You are right, the BP should have been saved with 4.22 for compatibility purposes, I missed that. 

As to the content in the blueprint, the workflow is to 

1. Disable Capture Every Frame.
2. Enable Always Persist Rendering State.
3. Re-enable Capture Every Frame. 

Please feel free to submit a PR if you are able to save BP_PIPCamera under 4.22.

PR #2544, opened 2020-04-07
UE4.22 Compatibility

plusk01:
Followed workflow outlined in https://github.com/microsoft/AirSim/issues/2536#issuecomment-610173134 for `SceneCaptureComponent`.

Verified that `"AutoExposureBias": 10` in `settings.json` worked with `NoDisplay` set.

Addresses #2536.


saihv:
@plusk The Persist Rendering State box still seems to be unchecked in BP_PIPCamera. Could you please make sure it was saved correctly?

image

saihv:
Please ignore the previous comment, I made a mistake in my project setup. It does work as expected!

PR #2553, opened 2020-04-08
Switch more details to VS2019

bys1123:


build.cmd
```diff
@@ -10,12 +10,12 @@ set "buildMode="
 REM //check VS version
 if "%VisualStudioVersion%"=="" (
     echo(
-    echo oh oh... You need to run this command from x64 Native Tools Command Prompt for VS 2017.
+    echo oh oh... You need to run this command from x64 Native Tools Command Prompt for VS 2019.
     goto :buildfailed_nomsg
 )
-if "%VisualStudioVersion%"=="14.0" (
+if "%VisualStudioVersion%"=="15.0" (
```
madratman, line 16: less than 16 could be better

PR #2555, opened 2020-04-09
[docs] Update CHANGELOG.md

rajat2004:
As mentioned in the Changelog, only keeping the bigger updates, not including bug-fixes, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,20 +15,21 @@ Cars in AirSim
 [![AirSim Car Demo Video](docs/images/car_demo_video.png)](https://youtu.be/gnz1X3UNM5Y)
 
 ## What's New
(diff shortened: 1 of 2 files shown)


rajat2004:
There's a [What's New](https://github.com/microsoft/AirSim#whats-new) section in the main Readme, details from that will need to be included. Should that be kept or just point to Changelog?

madratman:
Please remove whitespace changes in the diff - https://stackoverflow.com/questions/3515597/add-only-non-whitespace-changes. 
What's new should have the key highlights

rajat2004:
So the current changes to What's new are okay? And then some more detailed info to Changelog such as imp bug fixes?

rajat2004:
Removed the whitespace changes (need to fix that in some other PRs as well)
Will add some other bug-fixes, etc in Changelog

rajat2004:
I think the current list should be enough, for more details, the commit list is always there

rajat2004:
Looks okay to me, Changelog now has info about changes in April as well

madratman:
" * [Support for PX4 master](https://microsoft.github.io/AirSim/px4_setup/) " -> this is incorrect. 
we support v1.10 now. Chris is working on the master support in the latest PR.

rajat2004:
@madratman Fixed, when I wrote this, I thought it was working with the master branch as well
Changelog now includes changes till today as well

Issue #2559, opened 2020-04-10
Some problem abot the OSX


Gatsby23:
![image](https://user-images.githubusercontent.com/18000675/78952625-cbf4e300-7b08-11ea-8c62-8ba09a9e77c3.png)
My mac is 10.15.4 OSX. There are some problem when I run the./build.sh. I have see the pollin and pollout function in the class definition, but I don't know why it will be crashed.

rajat2004:
Hi,
Could you test https://github.com/microsoft/AirSim/pull/2560 and see if it fixes the problem? Also, which version of XCode are you using?

Gatsby23:
> Hi,
> Could you test #2560 and see if it fixes the problem? Also, which version of XCode are you using?

My Xcode is 11.4, should I change my xcode ?

rajat2004:
Not sure, just getting more info as to why the error is occurring, since this hasn't popped up before

Gatsby23:
> Not sure, just getting more info as to why the error is occurring, since this hasn't popped up before

OK, I'm sorry, but how to use the #2560? Should I git clone and check to this branch ?

rajat2004:
The change is just 1 line right now, you can manually change that and try, else

```
cd AirSim
git checkout master; # update if not already: git pull upstream master --rebase
git clean -fdx
git fetch upstream pull/2560/head:PR/2560; git checkout PR/2560;
./clean.sh
./build.sh;
```

saihv:
@Gatsby23 Can you provide some more details about your XCode version? Did you run the latest `setup.sh` before `build.sh`? 

master branch builds fine on my macOS - 10.15.4, XCode 11.4 (11E146)

```
➜  airsim git:(master) ✗ clang -v     
Apple clang version 11.0.3 (clang-1103.0.32.29)
Target: x86_64-apple-darwin19.4.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

Gatsby23:
> @Gatsby23 Can you provide some more details about your XCode version? Did you run the latest `setup.sh` before `build.sh`?
> 
> master branch builds fine on my macOS - 10.15.4, XCode 11.4 (11E146)
> 
> ```
> ➜  airsim git:(master) ✗ clang -v     
> Apple clang version 11.0.3 (clang-1103.0.32.29)
> Target: x86_64-apple-darwin19.4.0
[...]

PR #2560, opened 2020-04-10
[Mavlinkcom/UdpSocket] Fix select() compilation error on OSX

rajat2004:
Error reported here - https://github.com/microsoft/AirSim/issues/2559
Not sure if this is the correct fix though, needs testing

@saihv You have a OSX machine right? If possible, could you test this? Not sure why this hasn't come up till now

### DIFF
--- a/MavLinkCom/src/impl/UdpSocketImpl.hpp
+++ b/MavLinkCom/src/impl/UdpSocketImpl.hpp
@@ -19,9 +19,9 @@ static bool socket_initialized_ = false;
 #else
 
 // posix
-#include <sys/types.h> 
+#include <sys/types.h>
 #include <sys/socket.h>
-#include <arpa/inet.h>
+#include <sys/select.h>
 #include <netinet/in.h>
 #include <cerrno>
 #include <netdb.h>


saihv:
Interesting - will give it a shot today, but it looks like the error is about an overflow setting on a file descriptor (socket)? Is that related to `select`?

rajat2004:
That's what I'm not sure about. `select` seems to be the most probable one, but from the OSX man pages, seems correct. It's a compile-time error which seems pretty strange that it hasn't popped up till now
Can't do much more investigtion without some more info

BTW, is there any easy way to test out OSX on a non-Mac machine? Maybe VirtualBox?

saihv:
So this does build fine, but I am not sure how to test it properly w.r.t. the issue - because master works without issues as well..

> BTW, is there any easy way to test out OSX on a non-Mac machine? Maybe VirtualBox?

The feasibility exists, because macOS runs on Intel processors; but [Apple's EULA states that running the OS on non-Mac computers is illegal](https://images.apple.com/legal/sla/docs/macOS1012.pdf)

rajat2004:
Great, thanks for testing! So I think that if the error can't be reproduced and is found to be the result of some faulty configuration, then this can be closed

rajat2004:
Closing this till the error gets reproduced, can be reopened then. This anyways doesn't seem to be the correct fix

Issue #2552, opened 2020-04-08
AirSim as an RL environment


Mathias-Ooms:
OS: Ubuntu 18.04.4 LTS
Using AirSim Neighborhood binary v1.3.0-Linux
Python version: 3.6.9
My hardware: 
![Selection_004](https://user-images.githubusercontent.com/25950931/78812627-0016b980-79cc-11ea-80a9-7060a998ab4f.png)

Dear developers,
I'm trying to achieve an AirSim RL environment (much like the OpenAI environments). Following the car RL example in the documentation, I've created the following class below. However I stumbled upon some issues:

- I'm trying to achieve a high frame rate from the client.simGetImages() for depthview. The highest frame rate I got was around 15 FPS with the following settings (below). I've tried to change the viewmode to "NoDisplay" but this drops the frame rate to 1 FPS and lowering the resolution in the CaptureSettings didn't help either.

- Running the binary for a long time results sometimes in a sudden crash (log below), I'm running the binary as follows: 
./AirSimNH.sh -vulkan -novsync -benchmark (also tried with -opengl4 flag, but eventually crashes as well) 

- Is there a way to disable newly added animals?

Much thanks in advance!

------------------------------------------------------------------------------------------------------------------------------

The Environment class with main loop to measure FPS:

    class AirSim:
    def __init__(self):
        self.client = airsim.CarClient()
        self.client.confirmConnection()
        self.client.enableApiControl(True)
        self.car_controls = airsim.CarControls()

    def reconnect(self):
        self.client = airsim.CarClient()
        self.client.confirmConnection()
        self.client.enableApiControl(True)
        self.car_controls = airsim.CarControls()

    # return depth view images (84x84)
    def getImg(self):
        responses = self.client.simGetImages([airsim.ImageRequest("0", airsim.ImageType.DepthPerspective, True, False)])
        img = self.transformInput(responses)
        return img

    def transformInput(self, responses):
        img1d = np.array(responses[0].image_data_float, dtype=np.float)
        img1d = 255 / np.maximum(np.ones(img1d.size), img1d)
        img2d = np.reshape(img1d, (responses[0].height, responses[0].width))

        from PIL import Image
        image = Image.fromarray(img2d)
        im_final = np.array(image.resize((84, 84)).convert('L'))

        return im_final

    def interpret_action(self, action):
        self.car_controls.brake = 0
        self.car_controls.throttle = 1
        if action == 0:
            self.car_controls.throttle = 0
[...]

PR #2561, opened 2020-04-10
Add simDisableActor API

rajat2004:
Mentioned in https://github.com/microsoft/AirSim/issues/2552 
My first try with UE-specific coding, so things would probably be wrong :sweat_smile: 

Based on https://unrealcpp.com/hide-and-disable-actor/, https://docs.unrealengine.com/en-US/API/Runtime/Engine/GameFramework/AActor/index.html and some forum posts

### DIFF
--- a/AirLib/include/api/RpcLibClientBase.hpp
+++ b/AirLib/include/api/RpcLibClientBase.hpp
@@ -106,6 +106,8 @@ public:
 
 	std::vector<std::string> simSwapTextures(const std::string& tags, int tex_id = 0, int component_id = 0, int material_id = 0);
 
+    void simDisableActor(const std::string& object_name);
+
 protected:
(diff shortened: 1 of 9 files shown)


rajat2004:
Tested this to be working properly on Linux, with the simulation working properly after `simDisableActor()` was called. Object doesn't occur in the visual rendering and no collision as well

Thoughts: 
Object still appears in the list when `simListSceneObjects` is called. Should this be removed?
 Add `enableActor()` API? This might lead to it being called on things which haven't been disabled, or maybe something new also

Will need to look into whether the above both things can be done or not

rajat2004:
Rebased on master

rajat2004:
Closing this in favour of #2651
FYI @madratman

PR #2565, opened 2020-04-12
[Pythonclient] Add missing docstrings

rajat2004:
WIP


rajat2004:
As a side note, many PR's Travis status are not being updated, might be due to https://travis-ci.community/t/github-status-not-posted-on-commits-on-repositories-using-legacy-service-integration/7798

rajat2004:
Also, would be great to have a proper list of APIs auto-generated and added to the Wiki, that'll make things much easier. What should be used for that, Sphinx, pdoc, or something else?

madratman:
I've added a first version of sphynx doc generator. I'll do the next step soon: call `build_api_docs.sh` from `build_docs.sh`; copy sphynx html to gh-pages and rename it to api.html; and then finally we need a little button on main github.io site to link to API / or add it to the readme

rajat2004:
Need to add docstring for all the APIs to make it appear on the list, which is quite painful :sweat_smile:
Also fixing the types of the params will be good to make it jump to the type definition

rajat2004:
Quite a lot of the APIs have been done, the rest should be done soon hopefully
Note that I have removed `return` from some APIs since the return type for them was `void`. In those cases, checked that the return was `None`, and there was no change in the behaviour

rajat2004:
I've made a new PR https://github.com/microsoft/AirSim/pull/2599 while has some fixes noticed while working on this PR
Wanted to keep this one just for docs and not include other fixes which make things difficult while reviewing

PR #2634, opened 2020-04-29
Fix some PX4 issues with latest master branch

lovettchris:
Fix memory corruption in setThreadName helper function.
Print message when home GPS location is set.
Delay connecting UDP connection to avoid socket errors.

### DIFF
--- a/AirLib/include/vehicles/multirotor/firmwares/mavlink/MavLinkMultirotorApi.hpp
+++ b/AirLib/include/vehicles/multirotor/firmwares/mavlink/MavLinkMultirotorApi.hpp
@@ -972,6 +972,7 @@ private: //methods
         got_first_heartbeat_ = false;
         is_hil_mode_set_ = false;
         is_armed_ = false;
+        has_home_ = false;
         is_controls_0_1_ = true;
         Utils::setValue(rotor_controls_, 0.0f);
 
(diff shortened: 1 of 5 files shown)


rajat2004:
One thing which I had noticed sometime back is that PX4 is using the [system clock](https://github.com/microsoft/AirSim/blob/master/AirLib/include/vehicles/multirotor/firmwares/mavlink/MavLinkMultirotorApi.hpp#L1309) for sending the simulation time. This is in contrast to in Ardupilot where the [stepped time](https://github.com/microsoft/AirSim/blob/master/AirLib/include/vehicles/multirotor/firmwares/arducopter/ArduCopterApi.hpp#L407) is being used

I really have no idea whether this causes any effect or anything, just though I should post my doubt

lovettchris:
Interesting question about the clock, I'll ping my PX4 friends and see if the actual time delta between ticks makes any difference when PX4 is running in lockstep mode.  I doubt it will, but I'll check, thanks.

madratman:
This seems to be working well with master, v1.11.0-beta1, v.10.1. @TomSeestern, @bys1123, it would be great if you guys can give it a shot as well. (I haven't tried sending cmds with QGC). Just observing the terminal output

bys1123:
Thanks your great works! 
cc' @Jaeyoung-Lim @MaEtUgR

PR #2642, opened 2020-04-30
[ros] Add mavros_msgs to build dependencies

jsharpe:
If mavros is being built in the same catkin workspace then airsim_ros_pkgs needs to depend upon mavros_msgs for the packages to be built in the correct order.

### DIFF
--- a/ros/src/airsim_ros_pkgs/package.xml
+++ b/ros/src/airsim_ros_pkgs/package.xml
@@ -15,6 +15,7 @@
   <build_depend>image_transport</build_depend>
   <build_depend>message_generation</build_depend>
   <build_depend>message_runtime</build_depend>
+  <build_depend>mavros_msgs</build_depend>
   <build_depend>nav_msgs</build_depend>
   <build_depend>roscpp</build_depend>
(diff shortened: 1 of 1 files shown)


madratman:
This doesn't really fix the issue. 
`sudo apt-get install ros-melodic-mavros*` can be added to the docs

madratman:
But ok, if some one is compiling it from source, we should have that in list of deps.

jsharpe:
> But ok, if some one is compiling it from source, we should have that in list of deps.
> Wouldn't it also be a runtime dep?
RIght - it should just be a <depend> node, also a number of the other msgs dependencies could be converted to just a <depend> to remove duplication. Also now I look at it why is it depending upon message_generation at exec time? Seems like there is some clean up to be done in the packages.xml file

madratman:
ah yes, it should be
```xml
<build_depend>message_generation</build_depend> 
<exec_depend>message_runtime</exec_depend>
```

Yeah, I agree the `<depend>` tag can half the number of lines for sure, as it's kinda like a catch-all. Please feel free to add that on to this PR. If not, I can do that, and merge this one. Lemme know.

PR #2668, opened 2020-05-06
Allows to pass the settings.json file location via the --settings argument

ahorn42:
Allows to pass the settings.json file location via the --settings argument as a path in addition to passing its content.

Also updated documentation, which did not list the other loading possibilities added with #715.

### DIFF
--- a/Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.cpp
+++ b/Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.cpp
@@ -357,9 +357,10 @@ bool ASimHUD::getSettingsText(std::string& settingsText)
         readSettingsTextFromFile(FString(msr::airlib::Settings::Settings::getUserDirectoryFullPath("settings.json").c_str()), settingsText));
 }
 
-// Attempts to parse the settings text from the command line
+// Attempts to parse the settings file path or the settings text from the command line
 // Looks for the flag "--settings". If it exists, settingsText will be set to the value.
-// Example: AirSim.exe -s '{"foo" : "bar"}' -> settingsText will be set to {"foo": "bar"}
+// Example (Path): AirSim.exe --settings "C:\path\to\settings.json"
+// Example (Text): AirSim.exe -s '{"foo" : "bar"}' -> settingsText will be set to {"foo": "bar"}
 // Returns true if the argument is present, false otherwise.
 bool ASimHUD::getSettingsTextFromCommandLine(std::string& settingsText) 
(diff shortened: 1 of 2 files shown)


Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.cpp
```diff
@@ -372,6 +373,11 @@ bool ASimHUD::getSettingsTextFromCommandLine(std::string& settingsText)
         FString commandLineArgsFString = FString(commandLineArgs);
         int idx = commandLineArgsFString.Find(TEXT("-settings"));
         FString settingsJsonFString = commandLineArgsFString.RightChop(idx + 10);
+
+		if (readSettingsTextFromFile(settingsJsonFString.TrimQuotes(), settingsText)) {
+			return true;
+		}
+
```
rajat2004, line 380: Please change the indentation from Tabs -> 4 spaces as is followed in the file

docs/settings.md
```diff
@@ -1,10 +1,15 @@
 # AirSim Settings
 
 ## Where are Settings Stored?
-Windows: `Documents\AirSim`
-Linux: `~/Documents/AirSim`
+AirSim is searching for the settings definition in 4 different ways in the following order. The first match will be used:
 
-The file is in usual [json format](https://en.wikipedia.org/wiki/JSON). On first startup AirSim would create `settings.json` file with no settings. To avoid problems, always use ASCII format to save json file.
+1. Looking at the path specified by the `--settings` command line argument.
```
rajat2004, line 6: This requires the full path of the settings file correct? Mentioning that and adding an example command like in the comments above will be great

rajat2004 (commented):
Wasn't aware of this, thanks!

Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.cpp
ahorn42: changed it ;)

docs/settings.md
ahorn42: added an example and a remark that it must be an absolute path

ahorn42:
just updated the pull request according to your suggestions :)

rajat2004:
Thanks :) One last thing which will be good to do would be to squash the commits into 1, that keeps the history cleaner and makes life easy for the maintainers

ahorn42:
@rajat2004 done ;)

TomSeestern (approved):

PR #2682, opened 2020-05-10
Fix some warnings

rajat2004:
There are more warnings, many unused variables which can be seen when running the Unity build

Right now only testing till compilation

### DIFF
--- a/MavLinkCom/src/impl/MavLinkConnectionImpl.cpp
+++ b/MavLinkCom/src/impl/MavLinkConnectionImpl.cpp
@@ -430,8 +430,6 @@ void MavLinkConnectionImpl::readPackets()
             }
             else if (frame_state == MAVLINK_FRAMING_OK)
             {
-                int msgId = msg.msgid;
-
                 // pick up the sysid/compid of the remote node we are connected to.
                 if (other_system_id == -1) {
                     other_system_id = msg.sysid;
--- a/MavLinkCom/src/serial_com/TcpClientPort.cpp
+++ b/MavLinkCom/src/serial_com/TcpClientPort.cpp
@@ -227,7 +227,6 @@ public:
 	// write to the serial port
 	int write(const uint8_t* ptr, int count)
 	{
-		socklen_t addrlen = sizeof(sockaddr_in);
 		int hr = send(sock, reinterpret_cast<const char*>(ptr), count, 0);
 		if (hr == SOCKET_ERROR)
 		{
@@ -262,7 +261,6 @@ public:
 
 		while (!closed_)
 		{
-			socklen_t addrlen = sizeof(sockaddr_in);
 			int rc = recv(sock, reinterpret_cast<char*>(result), bytesToRead, 0);
 			if (rc < 0)
 			{
--- a/cmake/cmake-modules/CommonSetup.cmake
+++ b/cmake/cmake-modules/CommonSetup.cmake
@@ -59,7 +59,7 @@ macro(CommonSetup)
                 ${RPC_LIB_DEFINES} ${CMAKE_CXX_FLAGS}")
 
             if (${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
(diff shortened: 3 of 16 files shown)


cmake/cmake-modules/CommonSetup.cmake
```diff
@@ -54,7 +54,7 @@ macro(CommonSetup)
         else ()
             set(CMAKE_CXX_FLAGS "\
                 -std=c++17 -ggdb -Wall -Wextra \
-                -Wno-variadic-macros -Wno-parentheses -Wno-unused-function -Wno-unused \
+                -Wno-variadic-macros -Wno-parentheses -Wno-unused-function -Wno-unused -Wno-unknown-warning-option \
```
rajat2004, line 57: Should we remove the `-Wno-unused` flag here?

saihv: -Wno-unused is meant to ignore warnings involving unused variables, are you seeing those even with this flag on?

rajat2004: No, the warnings are coming from the Unity build where it hasn't been disabled
I'm not sure if we should disable them or not, I haven't yet fixed all the warnings

saihv:
I would advise against removing variables, especially in complex libraries like MavlinkCom - while they may be unused in that particular piece of code, they could have been meant to exist for the purpose of extensibility or some sort of parity. 

If a variable has a purpose and is named appropriately it does not detract from the code as a whole. Warnings are a smaller problem, and I recommend we handle it through compiler flags.

rajat2004:
I agree, remove all the unused variables probably doesn't make much sense, that's why I haven't yet removed variables from places like AdaptiveController and some others in Mavlinkcom
Should I remove the commits for Mavlinkcom which are there right now?

rajat2004:
Once the warnings which should be ignored and fixed are decided, then will clean up the commits
Also, the Unity cmake needs to be looked at, there shouldn't be 2 sets of build flags, etc.

MavLinkCom/src/impl/MavLinkConnectionImpl.cpp
```diff
@@ -430,8 +428,6 @@ void MavLinkConnectionImpl::readPackets()
             }
             else if (frame_state == MAVLINK_FRAMING_OK)
             {
-                int msgId = msg.msgid;
-
```
rajat2004, line 434: This can be easily removed since `msg.msgid` is used in many places

MavLinkCom/src/serial_com/TcpClientPort.cpp
```diff
@@ -257,12 +255,10 @@ class TcpClientPort::TcpSocketImpl
 
 	int read(uint8_t* result, int bytesToRead)
 	{
-		int bytesRead = 0;
 		// try and receive something, up until port is closed anyway.
 
 		while (!closed_)
 		{
-			socklen_t addrlen = sizeof(sockaddr_in);
```
rajat2004, line 265: This can also be removed, `addrlen` doesn't make much sense here
Also in 1 another place where this local variable is declared but not used, both can be removed

rajat2004:
I think I'll add back the other variables in Mavlinkcom but remove the ones commented above if others agree

rajat2004:
Commits cleaned up, only removed variables which didn't fit in the context

ros/src/airsim_ros_pkgs/src/pd_position_controller_simple.cpp
```diff
@@ -127,6 +127,10 @@ bool PIDPositionController::local_position_goal_srv_cb(airsim_ros_pkgs::SetLocal
         reset_errors(); // todo
         return true;
     }
+
+    // Already have goal, and have reached it
+    ROS_INFO_STREAM("[PIDPositionController] Already have goal and have reached it");
+    return false;
```
rajat2004, line 133: I'm not sure if this is correct or not, we should probably set the goal in this case also

PR #2683, opened 2020-05-10
Fix Python API example files

rajat2004:
Changes - 

1. Fix hello_car.py print, temp folder for images, cleanup
2. Delete legacy_hello_car.py
3. Add utility functions to check for NaN in Pose (#2695)
4. Fix cv_mode.py
5. [Unity] Don't send Alpha channel in image
6. [Utils] Use cv2 to write images (#2892)

Just some things noticed when trying to figure out why Unity Car APIs aren't working which has been reported on multiple issues. I'm not seeing anything wrong or different from the Unreal implementation right now. Ping @madratman @saihv  

Unity returns RGBA while Unreal does RGB (Not after this PR)


PythonClient/computer_vision/cv_mode.py
```diff
@@ -35,17 +45,18 @@
...
 
     pose = client.simGetVehiclePose()
     pp.pprint(pose)
 
     time.sleep(3)
 
 # currently reset() doesn't work in CV mode. Below is the workaround
-client.simSetPose(airsim.Pose(airsim.Vector3r(0, 0, 0), airsim.to_quaternion(0, 0, 0)), True)
\ No newline at end of file
+client.simSetVehiclePose(airsim.Pose(airsim.Vector3r(0, 0, 0), airsim.to_quaternion(0, 0, 0)), True)
```
rajat2004: Actually `client.reset()` does work in CV mode

madratman:
My guess would be we can change the buffer type of Unity from RGBA to RGB by not copying over the alpha channel?
It would be the equivalent of https://github.com/microsoft/AirSim/pull/1910/files#diff-598dac89bb5af3518db2ff6858f0a389L118. Can you try finding these lines in Unity?

rajat2004:
@madratman I think this is the one we need - https://github.com/microsoft/AirSim/blob/master/Unity/UnityDemo/Assets/AirSimAssets/Scripts/Utilities/DataCaptureScript.cs#L153
I could comment it out, will backwards compatibility be a problem. since no. of channels will change? I'll still test it out and commit the change to this PR

madratman:
No, I think it would be for the better as it establishes consistency in num of channels in UE and Unity,

rajat2004:
Made the Unity change, and reversed the channel change in the script.
2 other things I noted from testing -

1. Unity API (atleast Car movement one, need to check others also) is broken, Car doesn't move
2. Uncompressed images are corrupted I think, they don't get saved, and `cv2.imshow` shows a blank screen. I got to know about this 1-2 days back when doing some work with Recording where setting `Compress` to `False` didn't save the images

madratman:
Merging this.. let's track the remaining issues in new PRs

rajat2004:
#2937 fixes the Unity Car API problem, still needs a bit more work though

PR #2624, opened 2020-04-25
Upgrade setCameraOrientation to setCameraPose

saihv:
setCameraPose now takes a pose argument instead of just orientation. Commanded pose is expected to contain position in NED, **relative to default camera location**. 

Status:

- [x] Base implementation of setCameraPose for Unreal
- [x] Testing in Unreal
- [x] Unity implementation
- [ ] Testing in Unity
- [ ] Documentation
- [ ] Replacements in examples/ros wrapper

PR #2710, opened 2020-05-21
Upgrade setCameraOrientation to setCameraPose

rajat2004:
Replaces #2624 

From the original PR - 

setCameraPose now takes a pose argument instead of just orientation. Commanded pose is expected to contain position in NED, **relative to default camera location.**

Status:

- [x] Base implementation of setCameraPose for Unreal
- [x] Testing in Unreal
- [x] Unity implementation
- [x] Testing in Unity (https://github.com/microsoft/AirSim/pull/2624#issuecomment-627699236)
- [x] Documentation
- [x] Python examples replacement
- [x] Replacements in ROS wrapper


ros/src/airsim_ros_pkgs/src/airsim_ros_wrapper.cpp
```diff
@@ -804,7 +809,9 @@ void AirsimROSWrapper::drone_state_timer_cb(const ros::TimerEvent& event)
         if (has_gimbal_cmd_)
         {
             std::unique_lock<std::recursive_mutex> lck(drone_control_mutex_);
-            airsim_client_.simSetCameraOrientation(gimbal_cmd_.camera_name, gimbal_cmd_.target_quat, gimbal_cmd_.vehicle_name);
+            // Only camera rotation, no translation movement of camera
+            airsim_client_.simSetCameraPose(gimbal_cmd_.camera_name, get_airlib_pose(0, 0, 0, gimbal_cmd_.target_quat),
+                                            gimbal_cmd_.vehicle_name);
```
rajat2004, line 814: This change needs a review, in regards to whether new fields should be added to the Gimbal message, current implementation, etc.

rajat2004:
Might be better to merge this after #2712 just to make sure

mihailk239:
Please add same functional for python client.

rajat2004:
@mihailk239 The Python API change is there in the PR, see `client.py`
Should there be a backwards compatible layer, which calls the new API with 0 position? @madratman @saihv

ros/src/airsim_ros_pkgs/src/airsim_ros_wrapper.cpp
madratman: I think it's okay, as Gimbal usually implies an orientation only.

PR #2712, opened 2020-05-23
[Travis] Add ROS compilation test, dependencies script

rajat2004:
Came to mind when working on #2710

Should the AirSim dependencies be separated into a different script? Don't think there's much of a need for this, but let me know if that's better


rajat2004:
Should we separate the AirSim dependencies into a different script?

rajat2004:
@saihv Could you have a look at this?

madratman:
lgtm, thanks!

rajat2004:
Thanks for merging!

PR #2472, opened 2020-03-23
Faster image capture

ironclownfish:
A new PR for speeding up image capture, better than my previous PR https://github.com/microsoft/AirSim/pull/2453.

- Includes previous PR, which:
  - Locks the render target, reads the capture, then unlocks, instead of using ReadPixels().
  - Removes some unneeded complexity, like PNG compression and float-valued rgb
  - Removes extra mallocs/copies
  - Attempts to simplify the code, reduce looping and branching.    

- And adds the following 
  - Implementation of a BufferPool, and application of it to reduce memory usage in image capture.
  - Removed all the remaining extra memcopies, as far as I can tell.
  - Cleaned up some of the sloppy parts of last PR, removed a couple unrelated commits, and re-instated simGetImages (the plural one)

PR #2465, opened 2020-03-19
Add simDisableCamera API

rajat2004:
Closes https://github.com/microsoft/AirSim/issues/2464, https://github.com/microsoft/AirSim/issues/2389

Without `simDisableCamera`-
```
$ python3 test_camera.py
Connected!
Client Ver:1 (Min Req: 1), Server Ver:1 (Min Req: 1)

Time spent for 30 frames: 0:00:01.008567
Time spent for 30 frames: 0:00:01.033204
Time spent for 30 frames: 0:00:01.029388
Time spent for 30 frames: 0:00:01.053637
Time spent for 30 frames: 0:00:01.059051
Time spent for 30 frames: 0:00:01.067620
Time spent for 30 frames: 0:00:01.054935
Time spent for 30 frames: 0:00:01.060000
Time spent for 30 frames: 0:00:01.040280
Time spent for 30 frames: 0:00:01.101260
Time spent for 30 frames: 0:00:01.060938
Time spent for 30 frames: 0:00:01.108110
Time spent for 30 frames: 0:00:01.150538
Time spent for 30 frames: 0:00:01.093519
Time spent for 30 frames: 0:00:01.060111
Time spent for 30 frames: 0:00:01.051841
High resolution image captured.
Time spent for 30 frames: 0:00:01.925659
Time spent for 30 frames: 0:00:02.788289
Time spent for 30 frames: 0:00:02.831024
Time spent for 30 frames: 0:00:02.813883
```

With `simDisableCamera`-
```
Time spent for 30 frames: 0:00:01.092915
Time spent for 30 frames: 0:00:01.057923
Time spent for 30 frames: 0:00:01.062711
Time spent for 30 frames: 0:00:01.062194
Time spent for 30 frames: 0:00:01.138591
Time spent for 30 frames: 0:00:01.090212
Time spent for 30 frames: 0:00:01.048497
Time spent for 30 frames: 0:00:01.060650
Time spent for 30 frames: 0:00:01.096157
Time spent for 30 frames: 0:00:01.049348
Time spent for 30 frames: 0:00:01.104762
Time spent for 30 frames: 0:00:01.102033
Time spent for 30 frames: 0:00:01.055395
Time spent for 30 frames: 0:00:01.097566
Time spent for 30 frames: 0:00:01.094629
Time spent for 30 frames: 0:00:01.138703
High resolution image captured.
Time spent for 30 frames: 0:00:01.414585
Time spent for 30 frames: 0:00:01.041129
[...]

PR #2713, opened 2020-05-23
Faster Image Capture

rajat2004:
Rebased and working version of #2472. Needs more testing

See some benchmarks on my system here - https://github.com/microsoft/AirSim/pull/2472#issuecomment-633098747

One thing I noticed is that with the PR, we're getting 4 channels RGBA instead of 3 in master, this now becomes inline with Unity which is also giving 4 channels

My later commits might be a bit haphazard, will clean up later

Binaries - https://drive.google.com/open?id=14t9rmTGR3Au0R5V8GoF5DfYKYcLOJSAU
Linux - `-opengl` gives correct images, default Vulkan gives scrambled data, see comment below

On Linux, with binary, I'm getting about 2x improvement from ~15 (latest release) to ~35 FPS on my system 

Image types `Scene`, `Segmentation`, `Normals` work correctly, however `Depth` images is giving 8 channels somehow?

If someone wants to test using this PR from source, use this branch - https://github.com/rajat2004/AirSim/tree/faster_img_cap_test (It's updated to latest master, and will be pushing any commits there first)


rajat2004:
Update: The PR has been tested to be working correctly on Windows, on Linux, with `RLM_ReadOnly` it crashes due to an assert in the Engine code. `RLM_WriteOnly` gives garbage data on both Win & Linux.
Commented out the assert in Engine code in Linux, but still getting garbage data with `RLM_ReadOnly`. Working correctly with OpenGL, however FPS drops than with Vulkan, though still higher than master. Seems to be a bug in Vulkan, maybe will post on some UE4 forums

Linux testing has been done with both 4.24 & 4.25

Any suggestions would be great!

SijmenHuizenga:
This is amazing. Thank you so much for this amazing work <3

At [our repo](https://github.com/FS-Online/Driverless-Competition-Simulator/tree/airsim-patching-maddness) we have been trying to get this to work as well. We are on linux with vulkan, commented out the assert and are using `RLM_ReadOnly`. We are getting garbage data as well. What is most peculiar is that the image (garbage data) does not change when driving around. We will try to get it to work some more but might have to move to Windows.

rajat2004:
The last 2 commits are for fixing the problem being discussed in https://github.com/microsoft/AirSim/pull/2465, with the script from the PR as well included
From my testing, I didn't notice any difference in the case when only a single camera was being used continuously

Before modification -

```
Time spent for 50 frames: 0:00:02.471294
Time spent for 50 frames: 0:00:02.256129
Time spent for 50 frames: 0:00:02.173395
Time spent for 50 frames: 0:00:02.185919
High resolution image captured.
Time spent for 50 frames: 0:00:02.815300
Time spent for 50 frames: 0:00:05.559206
Time spent for 50 frames: 0:00:05.435409
Time spent for 50 frames: 0:00:05.387521
Time spent for 50 frames: 0:00:05.406460
High resolution image captured.
Time spent for 50 frames: 0:00:05.677712
Time spent for 50 frames: 0:00:05.367931

Img-benchmark script (single low-res camera)
result: 34.556743901198814 avg_fps for 364 num of images (--save_images)
```

After modification
```
result: 34.83061267025649 avg_fps for 365 num of images

Time spent for 50 frames: 0:00:02.140232
Time spent for 50 frames: 0:00:02.166878
Time spent for 50 frames: 0:00:02.150540
Time spent for 50 frames: 0:00:02.181311
Time spent for 50 frames: 0:00:02.154143
High resolution image captured.
Time spent for 50 frames: 0:00:02.980690
Time spent for 50 frames: 0:00:02.293952
Time spent for 50 frames: 0:00:02.265318
Time spent for 50 frames: 0:00:02.233132
Time spent for 50 frames: 0:00:02.238350
High resolution image captured.
Time spent for 50 frames: 0:00:02.598863
Time spent for 50 frames: 0:00:02.237810
Time spent for 50 frames: 0:00:02.235234
Time spent for 50 frames: 0:00:02.221081
Time spent for 50 frames: 0:00:02.262449
```

rajat2004:
> At [our repo](https://github.com/FS-Online/Driverless-Competition-Simulator/tree/airsim-patching-maddness) we have been trying to get this to work as well. We are on linux with vulkan, commented out the assert and are using `RLM_ReadOnly`. We are getting garbage data as well. What is most peculiar is that the image (garbage data) does not change when driving around. We will try to get it to work some more but might have to move to Windows.

Yeah, the garbage data is a pain, and I think this probably won't be fit to merge until that get's fixed.
On Linux, OpenGL does work though, add `-opengl` option when starting Editor from the terminal, I'm getting correct images when using OpenGL, and almost the same to sometimes even better FPS than Vulkan

rajat2004:
@madratman Is there any specific reason to remove the `save_images` option? Just curious, It was useful for checking if the images are correct when doing `cv2.imshow` reduced the FPS on my system

madratman:
I think it was cluttering the script. It's fine to run it with imshow once to verify things, and then without it to get actual numbers.

WouterJansen:
I tested this out. Performance does seem amazing now from a first glance. However I noticed a few issues. Do note I'm on a pretty old version of AirSim (I'd say the version from mid 2019 for Unreal 4.18) and am using 4.22 atm. 

In the old version the _RenderRequest_ functionality used BGR instead of RGB, which I always thought was odd (some CV2 specific choice?) so I changed it to RGB by changing the following:
```
 for (const auto& item : results[i]->bmp) {	
                        *ptr++ = item.B;	
                        *ptr++ = item.G;	
                        *ptr++ = item.R;	
 }
```
to RGB. This made my ROS code a lot easier to use. I couldn't not find in this PR how it was done now. Would be great if you could let me know if I could change it as well to RGB. 

I also experienced a crash. Again, I'm on 4.22 with a pretty old version of Airsim with many custom changes so not sure how relevant this crash report is. Was running from Editor. I can't seem to replicate the crash but figured I'd just post it here. I'll let you know if I get it more often. 

Using this request setup:

```
requests = []
requests.append(airsimpy.ImageRequest(front_center, get_camera_type("Scene"), False, False))
requests.append(airsimpy.ImageRequest(front_center, get_camera_type("Segmentation"), False, False))
requests.append(airsimpy.ImageRequest(front_center, get_camera_type("DepthPlanner"), True, False))
responses = client.simGetImages(requests)
```

Got a breakpoint crash here:
```
FTextureRenderTargetResource* UTextureRenderTarget::GameThread_GetRenderTargetResource()
{
	check( IsInGameThread() );
	return static_cast< FTextureRenderTargetResource*>( Resource );
}
```
the exception:
```
Exception thrown: read access violation.
**this** was nullptr.
```

callstack:
```
>	UE4Editor-Engine.dll!UTextureRenderTarget::GameThread_GetRenderTargetResource() Line 46	C++
 	UE4Editor-Renderer.dll!FScene::UpdateSceneCaptureContents(USceneCaptureComponent2D * CaptureComponent) Line 769	C++
 	UE4Editor-Engine.dll!USceneCaptureComponent::UpdateDeferredCaptures(FSceneInterface * Scene) Line 393	C++
 	UE4Editor-Renderer.dll!FRendererModule::BeginRenderingViewFamily(FCanvas * Canvas, FSceneViewFamily * ViewFamily) Line 3389	C++
 	UE4Editor-Engine.dll!UGameViewportClient::Draw(FViewport * InViewport, FCanvas * SceneCanvas) Line 1470	C++
 	UE4Editor-Engine.dll!FViewport::Draw(bool bShouldPresent) Line 1515	C++
 	UE4Editor-UnrealEd.dll!UEditorEngine::Tick(float DeltaSeconds, bool bIdleMode) Line 1821	C++
 	UE4Editor-UnrealEd.dll!UUnrealEdEngine::Tick(float DeltaSeconds, bool bIdleMode) Line 403	C++
 	UE4Editor.exe!FEngineLoop::Tick() Line 3967	C++
 	[Inline Frame] UE4Editor.exe!EngineTick() Line 62	C++
 	UE4Editor.exe!GuardedMain(const wchar_t * CmdLine, HINSTANCE__ * hInInstance, HINSTANCE__ * hPrevInstance, int nCmdShow) Line 168	C++
 	UE4Editor.exe!WinMain(HINSTANCE__ * hInInstance, HINSTANCE__ * hPrevInstance, char * __formal, int nCmdShow) Line 261	C++
 	[External Code]	
```

I also get a lot of warnings related to `bCaptureEveryFrame`. Should this be turned off? To make it work I simply followed what was said here : https://github.com/microsoft/AirSim/issues/2536#issuecomment-610173134_.
`
Warning: CaptureScene: Scene capture with bCaptureEveryFrame enabled was told to update - major inefficiency.`

rajat2004:
@WouterJansen Thanks for the detailed testing!

>  Do note I'm on a pretty old version of AirSim (I'd say the version from mid 2019 for Unreal 4.18) and am using 4.22 atm.

When testing this PR, you're on the latest master (well, even more cutting edge). Testing with 4.22 is great since I haven't done that

> 
> In the old version the _RenderRequest_ functionality used BGR instead of RGB, which I always thought was odd (some CV2 specific choice?) 

OpenCV uses BGR as it's default order, so probably yes

> 
> This made my ROS code a lot easier to use. I couldn't find in this PR how it was done now. Would be great if you could let me know if I could change it as well to RGB.

This currently just copies the entire block of pixels rendered by UE, see [this](https://github.com/microsoft/AirSim/pull/2713/files#diff-598dac89bb5af3518db2ff6858f0a389R45-R54). So even I'm not sure how to do that, need to look into that. But should be possible to write a wrapper code after calling the API to convert to desired format.

> 
> I also experienced a crash almost on every run now now. Again, I'm on 4.22 with a pretty old version of Airsim with many custom changes so not sure how relevant this crash report is. Was running from Editor. I'll let you know if I get more information on it.
> 
> Using this request setup:
> 
> ```
> requests = []
> requests.append(airsimpy.ImageRequest(front_center, airsim.ImageType.Scene, False, False))
> requests.append(airsimpy.ImageRequest(front_center, airsim.ImageType.Segmentation, False, False))
> requests.append(airsimpy.ImageRequest(front_center, airsim.ImageType.DepthPlanner, True, False))
> responses = client.simGetImages(requests)
> ```
> 
> I also get a lot of warnings related to `bCaptureEveryFrame`. Should this be turned off? To make it work I simply followed what was said here : [#2536 (comment)](https://github.com/microsoft/AirSim/issues/2536#issuecomment-610173134)_.
> ` Warning: CaptureScene: Scene capture with bCaptureEveryFrame enabled was told to update - major inefficiency.`

I'm not seeing these on Linux at least, and been some time since I tested on Win. Maybe was introduced by the recent revert to the asset which was disabling the flag?
So just some queries about this part. Are you getting the crashes after disabling `bCaptureEveryFrame`? Or before as well? What was the performance difference before and after disabling it? @saihv Since you've worked on this before, might have insights on what this does.

Also, could the crash be happening due to a specific image type? If possible, can you check if commenting out, say the Depth image doesn't cause it to crash?
Crash could be happening due to UE 4.22 as well. I'll try to do some more testing soon on my system.

WouterJansen:
I'm not on the master, I'm on a older version of Airsim with just this PR implemented on top of it :) 

The RGB/BGR thing was solved by just changing the encoding in my ROS publish of the image. Not a big deal just need to alter my ROS node. 

I'll do some more testing to see if it's a particular image type that causes the crash. 

I did not test before turning `bCaptureEveryFrame` on. I simply updated the camera blueprint as in that linked issue/comment to turn both those features on (_Capture Every Frame_ and _Always Persist Rendering State_).  I'll do some testing with turning on and off these settings. 


I did notice another issue. Seems like turning on `pixels_as_float` returns a `ImageResponse` with no floating image information. 
my request:
image
my response:
image
As you can see the response has `pixels_as_float` set to `False` `image_data_float` is empty.

WouterJansen:
Was doing another test. Now with only a scene capture (request:scene, compress=false,pixels_as_float=false).

Got a new crash in `SparseArray.h`:
`Assertion failed: Index < GetMaxIndex() [File:d:\unrealengine\engine\source\runtime\core\public\Containers/SparseArray.h] [Line: 86] `


`>	[Inline Frame] UE4Editor-Engine.dll!TSparseArray<TSetElement<FTickFunction *>,TSparseArrayAllocator<FDefaultAllocator,FDefaultBitArrayAllocator> >::AllocateIndex(int) Line 86	C++
 	UE4Editor-Engine.dll!TSparseArray<TSetElement<FTickFunction *>,TSparseArrayAllocator<FDefaultAllocator,FDefaultBitArrayAllocator> >::AddUninitialized() Line 125	C++
 	UE4Editor-Engine.dll!TSet<FTickFunction *,DefaultKeyFuncs<FTickFunction *,0>,FDefaultSetAllocator>::Emplace<FTickFunction * const &>(FTickFunction * const & Args, bool * bIsAlreadyInSetPtr) Line 510	C++
 	UE4Editor-Engine.dll!FTickTaskLevel::AddTickFunction(FTickFunction * TickFunction) Line 1111	C++
 	UE4Editor-Engine.dll!FTickFunction::SetTickFunctionEnable(bool bInEnabled) Line 1789	C++
 	UE4Editor-Engine.dll!UActorComponent::SetComponentTickEnabled(bool bEnabled) Line 862	C++
 	UE4Editor-Engine.dll!UActorComponent::Activate(bool bReset) Line 1524	C++
 	UE4Editor-AirSim.dll!APIPCamera::enableCaptureComponent(const msr::airlib::ImageCaptureBase::ImageType type, bool is_enabled) Line 447	C++
 	[Inline Frame] UE4Editor-AirSim.dll!APIPCamera::setCameraTypeEnabled(msr::airlib::ImageCaptureBase::ImageType) Line 248	C++
 	UE4Editor-AirSim.dll!UnrealImageCapture::getSceneCaptureImage(const std::basic_string<char,std::char_traits<char>,std::allocator<char> > & camera_name, msr::airlib::ImageCaptureBase::ImageType image_type, msr::airlib::ImageCaptureBase::ImageResponse & response) Line 33	C++
 	UE4Editor-AirSim.dll!PawnSimApi::getImages(const std::vector<msr::airlib::ImageCaptureBase::ImageRequest,std::allocator<msr::airlib::ImageCaptureBase::ImageRequest> > & requests, std::vector<msr::airlib::ImageCaptureBase::ImageResponse,std::allocator<msr::airlib::ImageCaptureBase::ImageResponse> > & responses) Line 191	C++
 	UE4Editor-AirSim.dll!msr::airlib::RpcLibServerBase::{ctor}::__l2::<lambda>(const std::vector<msr::airlib_rpclib::RpcLibAdapatorsBase::ImageRequest,std::allocator<msr::airlib_rpclib::RpcLibAdapatorsBase::ImageRequest> > & request_adapter, const std::basic_string<char,std::char_traits<char>,std::allocator<char> > & vehicle_name) Line 108	C++
 	[Inline Frame] UE4Editor-AirSim.dll!rpc::detail::call_helper<0>::call(msr::airlib::RpcLibServerBase::{ctor}::__l2::std::vector<msr::airlib_rpclib::RpcLibAdapatorsBase::ImageResponse,std::allocator<msr::airlib_rpclib::RpcLibAdapatorsBase::ImageResponse> > <lambda>(const std::vector<msr::airlib_rpclib::RpcLibAdapatorsBase::ImageRequest,std::allocator<msr::airlib_rpclib::RpcLibAdapatorsBase::ImageRequest> > &, const std::basic_string<char,std::char_traits<char>,std::allocator<char> > &) f, std::tuple<std::vector<msr::airlib_rpclib::RpcLibAdapatorsBase::ImageRequest,std::allocator<msr::airlib_rpclib::RpcLibAdapatorsBase::ImageRequest> >,std::basic_string<char,std::char_traits<char>,std::allocator<char> > > &) Line 75	C++
 	[Inline Frame] UE4Editor-AirSim.dll!rpc::detail::call_helper<1>::call(msr::airlib::RpcLibServerBase::{ctor}::__l2::std::vector<msr::airlib_rpclib::RpcLibAdapatorsBase::ImageResponse,std::allocator<msr::airlib_rpclib::RpcLibAdapatorsBase::ImageResponse> > <lambda>(const std::vector<msr::airlib_rpclib::RpcLibAdapatorsBase::ImageRequest,std::allocator<msr::airlib_rpclib::RpcLibAdapatorsBase::ImageRequest> > &, const std::basic_string<char,std::char_traits<char>,std::allocator<char> > &)) Line 56	C++
 	[Inline Frame] UE4Editor-AirSim.dll!rpc::detail::call_helper<2>::call(msr::airlib::RpcLibServerBase::{ctor}::__l2::std::vector<msr::airlib_rpclib::RpcLibAdapatorsBase::ImageResponse,std::allocator<msr::airlib_rpclib::RpcLibAdapatorsBase::ImageResponse> > <lambda>(const std::vector<msr::airlib_rpclib::RpcLibAdapatorsBase::ImageRequest,std::allocator<msr::airlib_rpclib::RpcLibAdapatorsBase::ImageRequest> > &, const std::basic_string<char,std::char_traits<char>,std::allocator<char> > &)) Line 56	C++
 	[Inline Frame] UE4Editor-AirSim.dll!rpc::detail::call(msr::airlib::RpcLibServerBase::{ctor}::__l2::std::vector<msr::airlib_rpclib::RpcLibAdapatorsBase::ImageResponse,std::allocator<msr::airlib_rpclib::RpcLibAdapatorsBase::ImageResponse> > <lambda>(const std::vector<msr::airlib_rpclib::RpcLibAdapatorsBase::ImageRequest,std::allocator<msr::airlib_rpclib::RpcLibAdapatorsBase::ImageRequest> > &, const std::basic_string<char,std::char_traits<char>,std::allocator<char> > &)) Line 84	C++
 	UE4Editor-AirSim.dll!rpc::detail::dispatcher::bind::__l2::<lambda>(const clmdep_msgpack::v2::object & args) Line 74	C++
 	[External Code]	
`

rajat2004:
> I'm not on the master, I'm on a older version of Airsim with just this PR implemented on top of it :)

Oh ok, that makes sense

>
> I did notice another issue. Seems like turning on `pixels_as_float` returns a `ImageResponse` with no floating image information.
> As you can see the response has `pixels_as_float` set to `False` `image_data_float` is empty. It also has `compress` turned on while my request had it disabled. Not sure if this was the case before this PR but floating point requests did work as expected before.
> 
> Edit: seems no matter what I do, _compress_ is always `True `and `pixels_as_float` always `False `on my response.

I think it might just be due to some pieces being missed out during implementing this, such as [this](https://github.com/microsoft/AirSim/pull/2713/files#diff-598dac89bb5af3518db2ff6858f0a389L109-L110), float will need some looking into.

rajat2004:
> Was doing another test. Now with only a scene capture (request:scene, compress=false,pixels_as_float=false).
> 
> Got a new crash in `SparseArray.h`:
> `Assertion failed: Index < GetMaxIndex() [File:d:\unrealengine\engine\source\runtime\core\public\Containers/SparseArray.h] [Line: 86] `

This seems to be somewhat common crash, like [this](https://issues.unrealengine.com/issue/UE-73467), could be due to 4.22? Not sure

WouterJansen:
> I think it might just be due to some pieces being missed out during implementing this, such as [this](https://github.com/microsoft/AirSim/pull/2713/files#diff-598dac89bb5af3518db2ff6858f0a389L109-L110), float will need some looking into.

Ok that makes sense. Well I could work around the float not being implemented yet, but would be good to have back at some point. 

Did some more testing, doesn't really seem to matter which camera type I use. _Scene_ seems to make the crashes happen easily, just like all the others. Turning on and off _Capture Every Frame_ and _Always Persist Rendering State_ also made no difference. It's relatively random.  It's more easy to trigger when looping it instead of the occasional request:

```python
    while True:
        requests = []
        requests.append(airsimpy.ImageRequest("front_center", airsimpy.ImageType.Scene, False, False))
        responses = client.simGetImages(requests)
        time.sleep(0.1)
```

I'll keep an eye out on this PR. If these crashes are only an issue that I am having I guess it's worth a full update of my Airsim to the latest versions :)

rajat2004:
So am doing a bit of testing today, but yeah, am able to trigger crash similar to this - https://github.com/microsoft/AirSim/pull/2713#issuecomment-644182511

@WouterJansen regarding `bCaptureEveryFrame`, did you disable that in every camera type in the BP Editor? Since it needs to be done in all I think. 
Should it actually be turned off though? It's description suggests it's better left on I think - `Whether to update the capture's contents every frame. If disabled, the component will render once on load and then only when moved.`

WouterJansen:
> @WouterJansen regarding `bCaptureEveryFrame`, did you disable that in every camera type in the BP Editor? Since it needs to be done in all I think.
> Should it actually be turned off though? It's description suggests it's better left on I think - `Whether to update the capture's contents every frame. If disabled, the component will render once on load and then only when moved.`

If I recall correctly, in my old version of AirSim  `bCaptureEveryFrame` was turned off. So I turned it on to align with the current AirSim settings. I did attempt toggling it off and on to see if it was the cause of the crash but doesn't seem like it. And yeah, I think it should be turned on as you say as well.

rajat2004:
Worked on the float part a bit, my testing branch is here - https://github.com/rajat2004/AirSim/tree/faster_img_cap_test, compare against this branch can be seen [here](https://github.com/rajat2004/AirSim/compare/pr/2472_rebased...rajat2004:faster_img_cap_test) 
What I did was to convert BufferPool to template and make `image_data_float` essentially equal to `image_data_uint`

I'm still confused about the `float` part a bit, just trying to lay down my thoughts here, please let me know where I am going wrong

- All 3 Depth images are of `float` type by default, others are `uint8`
- User can request any image to be of Float or non-Float type
- If image data is `float` (requested or through `image_type`), it's present in `image_data_float`. Also `pixels_as_float` should be `True` in this case
- User has to handle after getting the response, such as to access the correct field `image_data_uint8` or `image_data_float`. For saving float data, we have to use `write_pfm`

In my testing branch, right now the Depth images are of type `float`. User-requested data-type doesn't matter (which I think should be ok, they can be converted to one to other)
For some reason, I'm getting an 8 channel image, like when using `get_pfm_array()`, the error message - `ValueError: cannot reshape array of size 294912 into shape (144,256)`

petergerten:
Is there currently any recommended branch or commit to use if I want to take a look at a working 'faster' version?

rajat2004:
@petergerten You could try out this or even https://github.com/rajat2004/AirSim/tree/faster_img_cap_test branch, the latter one has some more commits, but `depth` image still doesn't work, will try to bring it up-to-date and work on it. Any commits will be made on the `faster_img_cap_test` branch, and then will port them over to here
You can try out the binaries I've linked in the first comment if haven't already done so. Note that on Linux, running the binaries as well as in Editor will require `-opengl` since Vulkan gives garbage data for some reason

rajat2004:
@madratman @saihv I was working on the `faster_img_cap_test` branch, and was always getting a linking error after merging the master branch into it. After taking a long break and going through the commits, figured out that it was happening due to the recent changes in `build.sh` & `AirSim.Build.cs` to use precompiled AirLib in UE
The problem was that I had changed `BufferPool` to a  template class in Airlib, objects of which were being used in UE code.This caused the linking error due to objects with the datatypes not being compiled when Airlib was built, and so definitions weren't found when linking. 
I added a commit which reversed the change to `Build.cs` and `build.sh`, and now it compiles.

madratman:
I think we should move BufferPool to the unreal/plugins part of the code, as it's only used in the Unreal part.

WouterJansen:
Any luck with the crashes?

rajat2004:
@WouterJansen No, it's still crashing. It seems to be more frequent on Windows than Linux, people have tried out 15 min runs with the binary without crashing,  the editor does crash more. Also if it crashes once, then it mostly crashes on the next run as well. There might be some deeper issue at hand here, since it's crashing where the campers is activated.
I was thinking of having a look at some other implementations, maybe Carla. Unfortunately I mostly won't be able to work on this for the next few days atleast

WouterJansen:
This is by far the current biggest bottleneck for me to use AirSim to the full potential. 

As soon as I add more then 1 camera to my `simGetImages()` request it just gets exponentially worse. So something like a stereo depth camera with 4 requests (2xRGB+2xdepth) makes it only produce images at 3FPS on good hardware. 

If any additional testing or help is required here I would love to assist.

phil333:
> I'll maybe try to work on this over the weekend, anyone is more than welcome to try it out and figure out the problems, everything is open-source.
> Just did a short-run on UE 4.25, doesn't seem to be crashing atleast on short runs. [Win Binary](https://drive.google.com/file/d/1pnOEuqlQuJH1bepcsv5oMnn2xVTYgHBu/view?usp=sharing)
> 
> But fixing the crashes and making it work on Vulkan seems to be very difficult to me right now, any ideas or suggestions are welcome.
> I also was thinking of a different short-tern approach separate from this PR, to collect the images in the request parallelly, UE has a `ParallelFor` method which might work

H rajat2004. At least from my side the ParallelFor method sounds interesting. One of my issues is that I am testing Stereo Cameras in Unreal, and the lack of synchronisation between multiple cameras is a major issue.

rajat2004:
I did try out the `ParallelFor` implementation but didn't work, the editor hangs, maybe cause the image rendering has to happen on the game thread, so running parallelly still won't work. Here's my branch - https://github.com/rajat2004/AirSim/tree/parallel-img-cap if anyone wants to test, note that code is quite messy

Unreal/Plugins/AirSim/Source/UnrealImageCapture.cpp
```diff
@@ -15,117 +14,46 @@ UnrealImageCapture::UnrealImageCapture(const common_utils::UniqueValueMap<std::s
...
+    response.camera_name = camera_name;
+    response.time_stamp = render_request.latest_result_.time_stamp;
+    response.width = padded_width;
+    response.height = height;
+    response.image_type = image_type;
+    response.image_data_uint8 = std::move(render_request.latest_result_.pixels);
+
+    // Disable camera after capturing image, this reduces resource consumption when images are not being taken
+    // Particulary when a high-resolution camera is used occasionally
+    camera->setCameraTypeEnabled(image_type, false);
```
cthoma20-arc, line 56: @rajat2004 Hi Rajat. Thanks for your hard work on this pull request. I've been looking at this issue for a little, and so far, after commenting out this line, everything works pretty much perfectly (please verify this, I made a few other modifications so it's possible I made a change somewhere else). I ran it twice for around 50 minutes each without a single crash. My guess is this line is causing a race condition somewhere. Maybe it needs to be run on a different thread? Please let me know if that change also works for you.

Best,
Tom

rajat2004: Awesome, thanks for testing and figuring out the problems! I'll also try it out, if it works out then can be moved into a separate API call which the user calls the disable cameras when not needed

rajat2004:
Here's a Linux Blocks binary with the suggested fix by @cthoma20-arc - https://drive.google.com/file/d/1sK-S0CX6cqGtuMeQLqz2OhAVVg6dZAq1/view?usp=sharing
I've added the commit at [faster_img_cap_test](https://github.com/rajat2004/AirSim/tree/faster_img_cap_test) branch, getting that PR back in sync with this just seems too much painful and not worth it, probably will open a new PR itself later to replace this one (again :sweat_smile: ) since comments and all are getting longer and longer
Someone trying it out and maybe even creating a Windows binary would be great. Currently, the Azure pipelines doesn't publish the packaged Ubuntu & Win binaries anywhere, that would be a great help, since packaging takes a lot of time, especially on a laptop, plus pipelines does it faster :)

Unreal/Plugins/AirSim/Source/RenderRequest.cpp
```diff
@@ -3,181 +3,57 @@
...
-    result->width = size.X;
-    result->height = size.Y;
-    FReadSurfaceDataFlags flags(RCM_UNorm, CubeFace_MAX);
-    flags.SetLinearToGamma(false);
-
-    return flags;
+    //Try to wait just long enough for the render thread to finish the capture.
+    //Start with a long wait, then check for completion more frequently.
+    //TODO: Optimize these numbers.
+    for (int best_guess_cap_time_microseconds = 500; !fast_cap_done_; best_guess_cap_time_microseconds = 200)
```
cthoma20-arc, line 34: Would it be possible to use a [condition variable](http://www.cplusplus.com/reference/condition_variable/condition_variable/) and a mutex here to block until the rendering is done instead of polling fast_cap_done_?

Unreal/Plugins/AirSim/Source/RenderRequest.cpp
```diff
@@ -3,181 +3,57 @@
...
+        std::this_thread::sleep_for(std::chrono::microseconds(best_guess_cap_time_microseconds));
```
cthoma20-arc, line 35: Would it be possible to use a [condition variable](http://www.cplusplus.com/reference/condition_variable/condition_variable/) and a lock to wait for the render thread to finish instead of polling fast_cap_done_?

cthoma20-arc (commented):
This is an awesome pull request, thanks again for fixing this issue 😃. I have a few more suggestions. I hope they will be useful.

Cheers,
Tom

Unreal/Plugins/AirSim/Source/RenderRequest.cpp
rajat2004: Changed to `mutex`, `condition_variable`, see https://github.com/rajat2004/AirSim/commit/901bd015286817229a01c3be23c4fb54a73a43d1
Didn't notice any major difference, not expected also

Unreal/Plugins/AirSim/Source/RenderRequest.cpp
```diff
@@ -3,181 +3,57 @@
...
-                assert(end_draw_handle_.IsValid());
-                game_viewport_->OnEndDraw().Remove(end_draw_handle_);
-            });
-
-            // while we're still on GameThread, enqueue request for capture the scene!
-            for (unsigned int i = 0; i < req_size_; ++i) {
-                params_[i]->render_component->CaptureSceneDeferred();
-            }
+    UAirBlueprintLib::RunCommandOnGameThread([this]() {
+        fast_cap_done_ = false;
```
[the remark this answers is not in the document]
rajat2004: Yup, done, see https://github.com/rajat2004/AirSim/commit/901bd015286817229a01c3be23c4fb54a73a43d1

AirLib/include/common/common_utils/BufferPool.h
```diff
@@ -0,0 +1,37 @@
...
+{
+public:
+    BufferPtr GetBufferExactSize(size_t size);
+    BufferPtr GetBufferAtLeastSize(size_t size, size_t max_size = std::numeric_limits<size_t>::max());
+
+private:
+    class BufferCollection
+    {
+    public:
+        BufferCollection(size_t size) : Size(size) {}
```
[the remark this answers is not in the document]
rajat2004: I'm not very sure about this, will need to think more and check if memory leak is happening
If destructor is defined, then copy constructor, assignment might also be required, atleast from what I understand
Any pointers, suggestions and discussions on this would be great!

rajat2004:
Sorry for the late reply, I think I'll mostly open a new PR with the other branch, there are still major points which need to be fixed, but can be discussed on that PR
All the magic was done by @ironclownfish & @madratman, thanks a lot! I'm just putting on some fixes, etc on top :)

rajat2004:
Opened #3018 which replaces this, have added the major problems right now in the description there, any testing, reviews would be great!

AirLib/include/common/common_utils/BufferPool.h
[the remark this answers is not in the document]
cthoma20-arc: Please correct me if I'm wrong, but I believe a memory leak will occur since we are allocating Buffers using `new` but the buffers are never `delete`d. We make unique_ptrs using the allocated buffers, but we use a custom deleter so they aren't deleted. This isn't really a problem at the moment since a single static BufferPool is being used, but it could be an issue in the future if someone uses more BufferPool objects. The destructor should probably delete the pointers in AvailableBuffers so that that memory is freed. I think it would probably make sense to delete the copy constructor and assignment operator, since it doesn't really make sense to copy this kind of object.
Thanks again :)

Unreal/Plugins/AirSim/Source/RenderRequest.cpp
cthoma20-arc: Awesome! Thank you

PR #2743, opened 2020-06-02
ROS - All sensors and car support

e1d1s1:
Many improvements to ROS package "airsim_ros_pkgs"

- ROS Support for all sensor types
- Support for ENU frames typical of ground vehicles
- Better support for multiple agents
- Option to publish sim /clock topic
- ROS Joystick support for car via python
- Bring CarApiBase interface to parity with MultirotorApiBase


ros/src/airsim_ros_pkgs/package.xml
```diff
@@ -15,11 +15,17 @@
   <build_depend>image_transport</build_depend>
   <build_depend>message_generation</build_depend>
   <build_depend>message_runtime</build_depend>
+  <build_depend>mavros_msgs</build_depend>
   <build_depend>nav_msgs</build_depend>
   <build_depend>roscpp</build_depend>
   <build_depend>rospy</build_depend>
   <build_depend>sensor_msgs</build_depend>
+  <!--build_depend>hector_uav_msgs</build_depend-->
```
rajat2004, line 23: Can be removed

tools/install_ros_deps.sh
```diff
@@ -30,4 +30,5 @@ fi
 
 sudo apt-get install gcc-8 g++-8
 sudo apt-get install ros-$ROS_DISTRO-mavros*
+sudo apt-get install ros-$ROS_DISTRO-tf2-sensor-msgs
```
rajat2004, line 33: Need to add to the [doc](https://github.com/microsoft/AirSim/blob/master/docs/airsim_ros_pkgs.md)

rajat2004 (commented):
This will be a great improvement to the ROS wrapper, thanks!
Can you add some details about the new sensor publishers, etc in [airsim_ros_pkgs.md](https://github.com/microsoft/AirSim/blob/master/docs/airsim_ros_pkgs.md)
Just a light lookover, will try to test it out soon

e1d1s1:
Should be all set for merge. Let me know if you see any other issues.

phil333:
nice work on the car interface. I found 1 little issue: Naming the car anything other than drone_1 means no control api, because the ros interface is looking for /airsim_node/drone_1/car_cmd and this is hard-coded. This is not a big deal though if you mention it in the documentation. Would be great to see this addition on the main branch.

e1d1s1:
> nice work on the car interface. I found 1 little issue: Naming the car anything other than drone_1 means no control api, because the ros interface is looking for /airsim_node/drone_1/car_cmd and this is hard-coded. This is not a big deal though if you mention it in the documentation. Would be great to see this addition on the main branch.

You should be able to change the topic via launch config:
node: car_joy
param: car_control_topic

The vehicle name/ID comes out of the airsim config, so if you alter this topic to match it "should" be fine.

PR #2749, opened 2020-06-05
Auto-build docs from master branch

rajat2004:
Commits picked from #2658

Needs work by the maintainers to add script with tokens, etc

From the linked PR comments - 

Added a commit which runs a currently non-existent `deploy_docs.sh` script on Travis master
I'm not sure of how the docs are updated, but probably `mkdocs gh-deploy` isn't used since it erases all history on the branch.

Maybe something like https://github.com/mkdocs/mkdocs/pull/1082? I went for Travis instead of a Github Action since we also Sphinx in the mix to build the Python API docs
There are several other ways to write the script, just need to find one which works best

Useful blog post - https://tech.michaelaltfield.net/2020/07/18/sphinx-rtd-github-pages-1/

Travis setup needs to be looked at

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,8 @@ before_install:
       sudo apt-get update;
       sudo apt-get -y install git wget unzip;
       sudo apt-get -y install build-essential software-properties-common cmake rsync libboost-all-dev;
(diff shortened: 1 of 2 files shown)


tool3210 (approved):
lgtm

rajat2004:
Replaced by #3224

PR #2770, opened 2020-06-13
XBox Controller fix

### LINKED ISSUE #2768: Xbox One controller drone always going up
Hi,
I have a issue when i only use Airsim with unreal, my drone is always trying to go up. It's not doing it when i use it with Ardupilot and Qgroundcontrol.

Thanks for helping me with this problem :)

[The controller](https://www.amazon.ca/PDP-Wired-Controller-Xbox-One/dp/B073X47SG4/ref=sr_1_8?crid=2TCRVSSI573T2&dchild=1&keywords=pdp+xbox+one+controller&qid=1591900390&sprefix=pdp+x%2Caps%2C167&sr=8-8)

rajat2004:
Fixes #2768 
Pending testing from @agentfeux

Zero2One conversion is being done here - https://github.com/microsoft/AirSim/blob/master/Unreal/Plugins/AirSim/Source/PawnSimApi.cpp#L227
Before this PR, it was changing an already 0-1 value to 0-1, so it introduced a positive offset

This change won't affect Win since it's kept behind #ifdefs

### DIFF
--- a/Unreal/Plugins/AirSim/Source/SimJoyStick/SimJoyStick.cpp
+++ b/Unreal/Plugins/AirSim/Source/SimJoyStick/SimJoyStick.cpp
@@ -316,7 +316,7 @@ public:
                     else { //XBox
                         switch(event_.number) {
                         case 0: state.left_x = normalizeAxisVal(event_.value, true, false, false); break;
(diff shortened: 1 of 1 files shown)


rajat2004:
Marking this as Ready for Review, would be great if anyone could test it out!

zimmy87:
tested this locally and it works for me

PR #2771, opened 2020-06-13
spelling error and 'vector' ambiguous symbol

xwnb:
a spelling error
Compiler Error C2872: 'vector' : ambiguous symbol in vs 2019

### DIFF
--- a/docs/image_apis.md
+++ b/docs/image_apis.md
@@ -25,14 +25,13 @@ png_image = client.simGetImage("0", airsim.ImageType.Scene)
 
 int getOneImage() 
 {
-    using namespace std;
     using namespace msr::airlib;
     
-    //for car use CarRpcLibClient
-    msr::airlib::MultirotorRpcLibClient client;
+    // for car use CarRpcLibClient
+    MultirotorRpcLibClient client;
 
-    vector<uint8_t> png_image = client.simGetImage("0", VehicleCameraBase::ImageType::Scene);
-    //do something with images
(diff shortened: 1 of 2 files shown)


docs/image_apis.md
```diff
@@ -106,7 +106,7 @@ int getStereoAndDepthImages()
     msr::airlib::MultirotorRpcLibClient client;
 
     //get right, left and depth images. First two as png, second as float16.
-    vector<ImageRequest> request = { 
+    msr::airlib::vector<ImageRequest> request = {
```
rajat2004, line 109: I think this should be `std`, don't think AirSim defines it's own vector

xwnb: Agree. In the file Common.hpp, it has "using vector = std::vector<T>", so std::vector or msr::airlib::vector both work. However, it occurs  Compiler Error C2872 if missing a namespace.

xwnb: Agree. In the file _Common.hpp_, it has "_using vector = std::vector<T>_", so both _std::vector_ or _msr::airlib::vector_ work. However, it occurs  **Compiler Error C2872** if missing a namespace.

rajat2004: That makes sense, I think then the problem is that we have 2 namespaces here, we should probably remove `using namespace std;` from the block, see [HelloDrone](https://github.com/microsoft/AirSim/blob/master/HelloDrone/main.cpp) or [HelloCar](https://github.com/microsoft/AirSim/blob/master/HelloCar/main.cpp)

Issue #2444, opened 2020-03-05
Distance Sensor in AirSim


YKembal:
I am trying to use the distance sensor (for a Vechile not a Drone) in airsim and it is working properly except that it only measure the distance from the bottom of the car to the ground and i don't know where to change it to make it measure the distance from the front of the car not from it's bottom , PLEASE HELP.

This is an example of the current distance sensor output :

<DistanceSensorData> {   'distance': 0.05738349258899689,
    'max_distance': 40.0,
    'min_distance': 0.20000000298023224,
    'relative_pose': <Pose> {   'orientation': <Quaternionr> {   'w_val': 1.0,
    'x_val': 0.0,
    'y_val': 0.0,
    'z_val': 0.0},
    'position': <Vector3r> {   'x_val': 0.0,
    'y_val': 0.0,
    'z_val': 0.0}},
    'time_stamp': 1583421822500312700}

rajat2004:
Distance sensor currently doesn't support modification of orientation or position through the settings, could have a look at using Lidar for this
Another possibility is to add such a transforming code which takes the orientation from settings and applies it, which should be similar to the already existing code for Lidar

YKembal:
> Distance sensor currently doesn't support modification of orientation or position through the settings, could have a look at using Lidar for this
> Another possibility is to add such a transforming code which takes the orientation from settings and applies it, which should be similar to the already existing code for Lidar

Can you give an example on how to taake the orientaaation of a sensor from settings and change it and apply it as you have said ?

rajat2004:
The settings for Lidar are parsed here - https://github.com/microsoft/AirSim/blob/master/AirLib/include/common/AirSimSettings.hpp#L1175
This takes in position and orientation of the Lidar, which are then used in [`LidarSimpleParams.hpp`](https://github.com/microsoft/AirSim/blob/master/AirLib/include/sensors/lidar/LidarSimpleParams.hpp#L70)
Some transformation also happens in [`LidarSimple.hpp`](https://github.com/microsoft/AirSim/blob/master/AirLib/include/sensors/lidar/LidarSimple.hpp#L89)

This should be a good starting place for adding such a feature. Files across different sensors have similar names and implementations, to make it easier for developing new features, which is the case across most of the AirSim codebase for different vehicles, etc.

EdwinMeriaux:
> Distance sensor currently doesn't support modification of orientation or position through the settings, could have a look at using Lidar for this
[...]

Issue #1029, opened 2018-04-27
How to use the distance sensor?


galgyeony:
I am very happy to see the new "distance" sensor but don't know how to use it.

As far as I know, the distance sensor internally uses getRayLength(), which is actually implemented by UE4.
However, I do not know how to acquire the result (i.e., to log or display the distance value).

Is there an API or another way to do so?

Thank you very much for your support.

sytelus:
There is no Python API yet. We are planning to expose this soon.

galgyeony:
I know that there are some GPS related sensor APIs (e.g., getGPSLocation) and IMU related ones (e.g., get*Kinematics).
But, for me, it is still difficult to find any API for magnetometor and barometer.

Is there any API for magnetometor and barometer? If not, are you planning to expose them?

Thank you!

dileep-krishnan:
Hi @sytelus ,
Any update on the above feature?
If not, can you point me to an existing API as example so that I can build make the API myself.

Thanks a lot!!

paulerikf:
@dileep-krishnan Did you figure this out?  Alternatively, does anyone have an example?

dahewujiang:
hi, @sytelus is there anything updated for distance sensor ? I cannot  find docs for how to use this sensor. You know that it's necessary to mount sensors (e.g.,ultrasonic sensor,millimeter-wave radar) on cars  if  we want to test our driverless car. i think that we need more sensors model (ultrasonic sensor and millimeter-wave radar are best ).

stavBodik:
@sytelus @paulerikf @dileep-krishnan 

for C++ users

 I also could not find anyway to get Barometer pressure...but :

Please look at BarometerSimple.hpp

Function : getOutputInternal(...)

 auto pressure = EarthUtils::getStandardPressure(altitude);

. . .

madratman:
I just added APIs for Barometer, IMU, GPS, Magnetometer, Distance Sensor in https://github.com/Microsoft/AirSim/pull/1920 

See the updated docs https://github.com/Microsoft/AirSim/blob/master/docs/sensors.md
and the API here https://github.com/Microsoft/AirSim/blob/master/docs/sensors.md#sensor-apis

madratman:
For details on distance sensor / point lidar, see the discussion in the original PR here https://github.com/Microsoft/AirSim/pull/791 and the original feature request here https://github.com/Microsoft/AirSim/issues/405

[...]

PR #2807, opened 2020-06-27
Distance sensor Upgrades

### LINKED ISSUE #2723: Visualizing sensors data in Client side (Unreal)
I am trying to visualise sensors data (Distance and Lidar sensors data ) in unreal engine, what the right way to approach this

### LINKED ISSUE #2806: Distance Detection Right in Front of Drone
Hello, 

So I would like to be able to detect the distance from just one point in front of the drone it see if there is anything immediately in front of it.

I know I cannot just use the distance sensor (since they can only for now detect distance from the drone to floor) But is there anyone way I can shift that to just in front of the drone. Or can I just use the Lidar to only detect 1 point in front of the drone?

rajat2004:
1. Add setting for drawing debug point with Distance Sensor
2. Expose Distance sensor settings, also move sensor above Car vehicle by 1 meter by default 
3. Fix `DistanceSensorData` struct in `types.py` (Should be float, not Quaternion)
4. Update docs
5. Use Custom sensor orientation in Mavlink Distance Sensor message (Needs testing with PX4)
6. ROS wrapper change
7. Example script for Distance sensor with 2 cars

Default sensor-
Screenshot from 2020-06-28 08-42-07

Downward-facing (altitude measurement)-
Screenshot from 2020-06-28 11-35-03

2 Cars with distance sensor
Screenshot from 2020-07-24 10-30-09

Closes #2723, closes #2806
Related issues - #2444, #1029

Question: Should there be fields in settings for other params such as `update_frequency`, noise, etc? Currently haven't added those

### DIFF
--- a/AirLib/include/api/RpcLibAdapatorsBase.hpp
+++ b/AirLib/include/api/RpcLibAdapatorsBase.hpp
@@ -672,7 +672,7 @@ public:
         DistanceSensorData()
         {}
 
-        DistanceSensorData(const msr::airlib::DistanceBase::Output& s)
+        DistanceSensorData(const msr::airlib::DistanceSensorData& s)
         {
             time_stamp = s.time_stamp;
             distance = s.distance;
@@ -681,9 +681,9 @@ public:
             relative_pose = s.relative_pose;
         }
 
-        msr::airlib::DistanceBase::Output to() const
+        msr::airlib::DistanceSensorData to() const
         {
-            msr::airlib::DistanceBase::Output d;
+            msr::airlib::DistanceSensorData d;
 
             d.time_stamp = time_stamp;
             d.distance = distance;
--- a/AirLib/include/api/RpcLibClientBase.hpp
+++ b/AirLib/include/api/RpcLibClientBase.hpp
@@ -84,7 +84,7 @@ public:
     msr::airlib::BarometerBase::Output getBarometerData(const std::string& barometer_name = "", const std::string& vehicle_name = "") const;
     msr::airlib::MagnetometerBase::Output getMagnetometerData(const std::string& magnetometer_name = "", const std::string& vehicle_name = "") const;
     msr::airlib::GpsBase::Output getGpsData(const std::string& gps_name = "", const std::string& vehicle_name = "") const;
-    msr::airlib::DistanceBase::Output getDistanceSensorData(const std::string& distance_sensor_name = "", const std::string& vehicle_name = "") const;
+    msr::airlib::DistanceSensorData getDistanceSensorData(const std::string& distance_sensor_name = "", const std::string& vehicle_name = "") const;
 
     // sensor omniscient APIs
     vector<int> simGetLidarSegmentation(const std::string& lidar_name = "", const std::string& vehicle_name = "") const;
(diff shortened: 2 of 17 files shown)


mirarmdn:
How does the distance sensor airsim measure the distance from the car to the obstacles in front? because from the simulation that has been done, the result of the distance always increases even though the distance from the car to the obstacle is getting closer
anyone can help, please

rajat2004:
@mirarmdn Distance sensor uses a ray cast to measure the distance. You're right, I just tested, and there's some problem with Car vehicle, the distance values don't make sense. It's working correctly on Multirotor vehicle though. Need to look into this

mirarmdn:
@rajat2004 What should I do? I desperately need the distance sensors to measure the distance from the car to the obstacle. I've looked into several sources, but I haven't found an answer. Can you help me
Thank you

rajat2004:
@mirarmdn I've pushed the fix for the Car vehicle, the problem was that the sensor was inside the vehicle, and so gave the wrong values. Moving it by 1 meter above fixes the problem, this is also being done in Lidar as well. Added an example script as well for measuring the distance between 2 vehicles which I used for testing 

Screenshot from 2020-07-24 10-30-09

madratman:
This PR is looking good and complete to me. @rajat2004 do you want to add anything else here?

rajat2004:
@madratman Don't think so, I was thinking of adding support for distance sensor in ArduCopter, Rover vehicles but that will take time and requires a PR in AP repository to be merged first, so this can go in and will open a new PR later when that happens.
One point s whether fields like noise, etc should be exposed in settings or not, right now haven't done that.

madratman:
Yes, let's add those features in next PR. I'm getting this in.

fracnaive:
Problem:
It always get the same result ZERO no matter how i change "X","Y","Z" of Distance Sensors when i run the distance_sensor_multi.py.
my settings.json:
{"SeeDocsAt": "https://github.com/Microsoft/AirSim/blob/master/docs/settings.md","SettingsVersion": 1.2,"SimMode": "Car", "Vehicles": {"Car1": {"VehicleType": "PhysXCar","AutoCreate": true, "Sensors": { "Distance": { "SensorType": 5, "Enabled" : true,"X": 0, "Y": 0, "Z": 5, "DrawDebugPoints": true}}}, "Car2": {"VehicleType": "PhysXCar", "AutoCreate": true, "X": 10, "Y": 0, "Z": 0, "Sensors": {"Distance": { "SensorType": 5, "Enabled" : true,"X": 0, "Y": 0, "Z": 5,"DrawDebugPoints": true }}}}}
result:
Connected!

Distance sensor data: Car1: 0.0, Car2: 0.0
Distance sensor data: Car1: 0.0, Car2: 0.0
Distance sensor data: Car1: 0.0, Car2: 0.0
Distance sensor data: Car1: 0.0, Car2: 0.0
Distance sensor data: Car1: 0.0, Car2: 0.0
Distance sensor data: Car1: 0.0, Car2: 0.0
Distance sensor data: Car1: 0.0, Car2: 0.0
Distance sensor data: Car1: 0.0, Car2: 0.0
Distance sensor data: Car1: 0.0, Car2: 0.0
Distance sensor data: Car1: 0.0, Car2: 0.0

image

rajat2004:
@fracnaive Your sensor is probably below the vehicle, in the ground itself. NED is used here, so +ve Z will be downwards.
Also for the future, it's preferable to create a new issue

fracnaive:
I get it，thanks！

Issue #2409, opened 2020-02-07
simGetLidarSegmentation() do not match getLidarData().point_cloud


xiaoaoran:
I tried to get segmentation ground truth via `client.simGetLidarSegmentation()`:

`lidarData = self.client.getLidarData()`
`lidarSeg = self.client.simGetLidarSegmentation()`

`print(len(lidarData.point_cloud))`
`>>>3858`
`print(len(lidarSeg))`
`>>>1336`
So the number of lidar point in `lidarData` should be 3858/3=1286, which is not equal to the number of ground truth points in `lidarSeg `

but @ironclownfish  said 

> api function: simGetLidarSegmentation().
Returns the segmentation of each lidar point's collided object in the last lidar update. i.e. the segmentation IDs align with getLidarData().point_cloud
[https://github.com/microsoft/AirSim/pull/2011#issue-286773742](url)

Anyone could tell me why? Thanks!

xiaoaoran:
Please help me with this problem

xiaoaoran:
Hello?

TheFloe1995:
I also stumbled across this issue. It seems that you have to call both functions _right after each other_. 

In a quick experiment I called both functions directly after each other at 10 FPS while moving. Most of the time the number of points was equal in both lists. But in 7/300 cases they were different. 

Seems to be timing dependent which is quite inconvenient in my opinion. Are there any plans to change that behaviour, e.g. having a single function call to get the point cloud + segmentation IDs?

rajat2004:
I've opened #2810 which combines segmentation data into `LidarData` struct fetched using `getLidarData`

PR #2810, opened 2020-06-29
Combine Lidar Segmentation API into getLidarData

rajat2004:
Having separate APIs means that if we want the segmentation data corresponding to the point cloud, it has to be called just after `getLidarData`
Even this is not foolproof, and might get mismatched data. Related issue #2409 
Combining the 2 APIs seemed to be to a cleaner solution. Also add generic `findSensorByName` function in VehicleApiBase

A simple test-script -
```python
import setup_path
import airsim
import time

client = airsim.CarClient()
client.confirmConnection()

count = 0

for i in range(1000):
    lidarData = client.getLidarData()
    # lidarSegData = client.simGetLidarSegmentation()
    lidarSegData = lidarData.segmentation
    
    if not lidarData.point_cloud:
        print("Null entry, skipping")
        continue

    print(f"Lidar: {len(lidarData.point_cloud)}")
    # time.sleep(0.1)
    print(f"LidarSeg: {len(lidarSegData)}")

    if(len(lidarSegData) != len(lidarData.point_cloud)//3):
        print("Not equal")
        count+=1
    time.sleep(0.1)

print(count)
```

For envs like Neighbourhood, I'm getting 30-40 out of 1000 mismatched entries (which isn't a lot actually)

Note: I haven't added the workaround for empty `segmentation` vecotr similar to what is there for `point_cloud` in `RpcLibAdaptorsBase` since https://github.com/rpclib/rpclib/issues/152 has been fixed. I've opened #2812 for removing the workaround, hopefully they rebase cleanly
With the removal of the workaround, checking whether the point_cloud is empty using `if not lidarData.point_cloud:` works

Edit: Well, after thinking about it a bit more, it somewhat makes sense that the segmentation API is a different one since this data won't actually be present in real life, therefore `sim` prefix. This problem can also be detected pretty easily, so if we want to keep the APIs separate, the first option could be to close this PR (or just use the last commit)
Another could be to create a new Lidar sensor which has the segmentation info as well. This will also lead to the removal of collecting segmentation data in the normal Lidar sensor, which probably will improve performance (but will need to check how much it's affected)
Let me know whatever seems best

### DIFF
--- a/AirLib/include/api/RpcLibAdapatorsBase.hpp
+++ b/AirLib/include/api/RpcLibAdapatorsBase.hpp
@@ -465,8 +465,9 @@ public:
         msr::airlib::TTimePoint time_stamp;    // timestamp
         std::vector<float> point_cloud;        // data
         Pose pose;
+        std::vector<int> segmentation;
 
-        MSGPACK_DEFINE_MAP(time_stamp, point_cloud, pose);
+        MSGPACK_DEFINE_MAP(time_stamp, point_cloud, pose, segmentation);
 
         LidarData()
         {}
@@ -476,6 +477,7 @@ public:
             time_stamp = s.time_stamp;
             point_cloud = s.point_cloud;
             pose = s.pose;
+            segmentation = s.segmentation;
         }
 
         msr::airlib::LidarData to() const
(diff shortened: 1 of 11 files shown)


zimmy87:
tested this locally and it works for me so I'm going ahead and merging this

PR #2812, opened 2020-06-30
Remove workaround for empty struct in RPC

rajat2004:
https://github.com/rpclib/rpclib/issues/152 has been fixed
Testing was with #2810 where empty Lidar data was produced sometimes, no crash, and can be checked easily in Python e.g.
`if not lidar_data.point_cloud:`

### DIFF
--- a/AirLib/include/api/RpcLibAdapatorsBase.hpp
+++ b/AirLib/include/api/RpcLibAdapatorsBase.hpp
@@ -405,12 +405,6 @@ public:
             image_data_uint8 = s.image_data_uint8;
             image_data_float = s.image_data_float;
 
-            //TODO: remove bug workaround for https://github.com/rpclib/rpclib/issues/152
-            if (image_data_uint8.size() == 0)
-                image_data_uint8.push_back(0);
-            if (image_data_float.size() == 0)
-                image_data_float.push_back(0);
-
             camera_name = s.camera_name;
             camera_position = Vector3r(s.camera_position);
             camera_orientation = Quaternionr(s.camera_orientation);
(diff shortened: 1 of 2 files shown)


madratman:
Thanks for this. Does this PR affect the client side behavior when empty image is returned. 
(yes, the client side would need to check for the same, as is the case now. But my question is if you've tested the same for images as well?)

rajat2004:
The client-side behaviour which would change is that it'll be possible to check if the image is empty with `if not image_data_uint8` or something, or the always present way to check if length matches with HxWxC. This will affect a check of whether the length is 1 or not, but that's wrong to begin with. Will test with images to confirm.
This does give an idea to add a utility function to check if image is empty or not

rajat2004:
Well, it's surprisingly difficult to get empty images, fetched hundreds of images, even cranked up the res to 4k but still no empty images. Great that it happens so rarely but makes it that much harder to reproduce.
If anyone is able to test or has any ideas, do tell me. A possible way could be to add code to randomly clear the vector when constructing the RPC message

tool3210 (approved):
lgtm

zimmy87:
Tested this locally with the Lidar scenario and it worked for me. I'm also unsure of how to test this with images. I tried configuring a camera to have 0 width and 0 height, but that generates an exception in the camera setup logic.

rajat2004:
@zimmy87 Thanks for testing! I haven't tested more than my previous comment indicated, but feel that the idea mentioned there to add a piece of code which randomly does not put the image in should work. Will try to test it out soon

zimmy87:
I hard-coded the ImageResponse to have empty image data and the simGetImages API still worked for me, so I think this is a safe enough fix to make.

rajat2004:
Great, thanks a lot for testing!
If you don't mind, could you have a look at #3180 and see whether GitHub Actions needs to be enabled or something? It would give at least some basic CI checks until Azure Pipelines is up again which can do the entire build with UE4

jonyMarino:
Hi Rajat! Will do. I think the Azure pipelines are working. We didn't call it here because Zimmy compiled in his local machine.

PR #2814, opened 2020-06-30
Fix documented example image reshape in image_apis.md

TRex22:
# What does this PR do?
- Resolves a minor issue in the documentation where `3` dimensions were used and not `4`
- Adds a `Quick Tip` around image resizing

### DIFF
--- a/docs/image_apis.md
+++ b/docs/image_apis.md
@@ -71,7 +71,7 @@ response = responses[0]
 img1d = np.fromstring(response.image_data_uint8, dtype=np.uint8) 
 
 # reshape array to 4 channel image array H X W X 4
-img_rgb = img1d.reshape(response.height, response.width, 3)
+img_rgb = img1d.reshape(response.height, response.width, 4)
 
 # original image is fliped vertically
 img_rgb = np.flipud(img_rgb)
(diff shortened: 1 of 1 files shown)


docs/image_apis.md
```diff
@@ -71,7 +71,7 @@ response = responses[0]
 img1d = np.fromstring(response.image_data_uint8, dtype=np.uint8) 
 
 # reshape array to 4 channel image array H X W X 4
-img_rgb = img1d.reshape(response.height, response.width, 3)
+img_rgb = img1d.reshape(response.height, response.width, 4)
```
rajat2004, line 74: UE currently gives RGB image, so I think the comment above needs to be fixed rather than the code

TRex22: In my testing, with the latest environments in releases the returned image has 4 channels RGBA. Im not sure where the alpha channel is coming from but this change would be accurate to how the current AirSim python client behaves and what data you get back.

Potentially there is a bug somewhere and I can investigate it

TRex22: I wanted to make this change because it tripped me up and wasted days of me trying to debug my code - then I realised the dimensions didnt match. I was using a different 1d -> 3d image resize mechanism (not the one defined in the docs) so I didnt get an error but rather scrambled data

rajat2004: Is that so, from what I remember, it gives RGB, I'll test again.
Are you using Unity by any chance? I think that returns RGBA

TRex22: No definitely Unreal. I tested this with the neighbourhood environment

rajat2004: Sorry for the late response, however, I can say that the API is returning an RBG image rather than RGBA for sure. 
Relevant UE code - https://github.com/microsoft/AirSim/blob/master/Unreal/Plugins/AirSim/Source/RenderRequest.cpp#L114-L116
Unity is also being changed to RBG in #2683

PR #2834, opened 2020-07-07
Add Recording APIs

### LINKED ISSUE #2832: Record images using API instead of Button
I am working with a script that runs several simulations with AirSim, and I would like to automate the clicking of the Record button. Is there anyway I could do this using the Python API?

rajat2004:
Closes #2832 
Contains non-implemented Unity methods, Python docstrings and adds info in `apis.md` as well


madratman:
Please split into enable and disable Recording.

rajat2004:
@madratman done! I've changed the names to `startRecording`, `stopRecording` to align with the methods that they call

PR #2845, opened 2020-07-10
Fixes #2815: Non-linear bias corrupts SurfaceNormals when requested through AirSim

### LINKED ISSUE #2815: Surface normals having a weird bias applied when requesting via AirSim
I'm working with surface normals requested via AirSim and observed a strange behavior. There seems to be a small non-linear biased applied to the surface normals' color values when requesting via airsim.ImageRequest and type airsim.ImageType.SurfaceNormals. 

I created a test scene with 9 axis aligned cubes and a camera looking down the positive x-axis. I then request a surface normal image via AirSim and let Unreal Engine 4 produce one be utilizing the highres screen shot function with "Include Buffer Visualization Targets" enabled. Here's the result:

**Exported via UE4:**
image

**Exported via AirSim:**
image

The camera facing side of a the cube (the cyan one) should have a world normal of (-1, 0, 0) which according to the responsible material in the UE4 AirSim plugin (Plugins\AirSim\Content\HUDAssets\NormalsMaterial.uasset) should result in a color of (0, 0.5, 0.5) in float space or (0, 127, 127) in RGB space: 
image

But as you can clearly see, the color of said cube side on the AirSim export is (0, 187, 187) which results in a normal of (-0.87, 0.39, 0.39) after normalization. There's clearly a bias applied and I didn't found a workaround yet. 

**settings.json:**
```json
{
    "SettingsVersion": 1.2,
    "SimMode": "ComputerVision",
    "Vehicles": {
        "MyVehicle": {
            "VehicleType": "ComputerVision",
            "Cameras": {
                "my_camera": {
                    "CaptureSettings": [
                        {
                            "ImageType": 6,
                            "FOV_Degrees": 60,
                            "Width": 1024,
                            "Height": 1024
                        }
                    ],
                    "X": 0.0,
                    "Y": 0.0,
                    "Z": 0.0,
                    "Pitch": 0.0,
                    "Roll": 0.0,
                    "Yaw": 0.0
                }
            }
        }
    }
}
```

**Minimal working Python code:**
```python
import setup_path
import airsim

client = airsim.VehicleClient(timeout_value = 7200)
client.confirmConnection()

responses = client.simGetImages([airsim.ImageRequest("my_camera", airsim.ImageType.SurfaceNormals)])
airsim.write_file("normal.png", responses[0].image_data_uint8)
```

**UE4 Map file**
[BrokenSurfaceNormalsUE4Map.zip](https://github.com/microsoft/AirSim/files/4851396/BrokenSurfaceNormalsUE4Map.zip)

**Version infos**
* Unreal Engine: 4.24.2
* AirSim Plugin build from commit: Mon Jun 1 10:30:23 2020 +0530
* AirSim Python library: 1.2.8 
* Python: 3.7.2 32bit

LSBOSS:
This fix forces usage of linear gamma for RenderTargets utilized for capturing SurfaceNormals and Segmentations, resulting in correct colors being output. (See #2815 for discussion about what is considered being correct) 

As a bonus, the post process material for extracting world normals and packing them into a RGB texture has been marked as *unlit* for some eventual speed ups and also because the buffer visualization material of Unreal Editor is doing the very same :)


Unreal/Plugins/AirSim/Source/PIPCamera.cpp
```diff
@@ -286,12 +286,21 @@ void APIPCamera::setupCameraFromSettings(const APIPCamera::CameraSetting& camera
...
         if (image_type >= 0) { //scene capture components
-            if (image_type==0 || image_type==5 || image_type==6 || image_type==7)
-                updateCaptureComponentSetting(captures_[image_type], render_targets_[image_type], false, 
-                    image_type_to_pixel_format_map_[image_type], capture_setting, ned_transform);
-            else
-                updateCaptureComponentSetting(captures_[image_type], render_targets_[image_type], true, 
-                    image_type_to_pixel_format_map_[image_type], capture_setting, ned_transform); 
+            switch (image_type) {
+                case 0:
+                case 7:
```
rajat2004, line 291: Having direct int values makes it more difficult to understand, might be better to use enum fields such as `ImageType::Scene` etc.

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
```diff
@@ -286,12 +286,21 @@ void APIPCamera::setupCameraFromSettings(const APIPCamera::CameraSetting& camera
...
+                    updateCaptureComponentSetting(captures_[image_type], render_targets_[image_type], false, image_type_to_pixel_format_map_[image_type], capture_setting, ned_transform, false);
+                    break;
+
+                case 5:
+                case 6:
+                    updateCaptureComponentSetting(captures_[image_type], render_targets_[image_type], true, image_type_to_pixel_format_map_[image_type], capture_setting, ned_transform, true);
```
rajat2004, line 297: So does forcing Linear Gamma Space change the values in segmentation image like you mentioned in the issue?
(I have no idea what this does though :sweat_smile: )

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
LSBOSS: Makes absolute sense - I just bluntly copied what was there before :) Let me fix this in the next commit to this pull request.

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
LSBOSS: It absolutely does. Now the values of the segmentation output correspond to those of mentioned seg_rgbs.txt.

The ForeLinearGamma flag basically tells UE4 to not apply any transformation of input colors (which UE4 is usally doing) when running through gamma correction. The background is that UE4 internally stores color information in sRGB space wherever it can for a better color reproduction. That's a non linear space that, if displayed on your monitor directly, seems to be too dark for the human eye. Here come's the gamma correction into action and is "correcting" things back to normal. 

That's all fine for ImageType::Scene but not for textures not in sRGB space and where exact colors are important. E.g. Segmentation and SurfaceNormals.

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
LSBOSS: DONE 👍

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
madratman: Thanks for another good PR. 

Here's a segmentation test. Use this settings.json
```json
{
    "SettingsVersion": 1.2,
    "SimMode": "ComputerVision",
    "Vehicles": {
        "MyVehicle": {
            "VehicleType": "ComputerVision",
            "Cameras": {
                "my_camera": {
                    "CaptureSettings": [
                        {
                            "ImageType": 5,
                            "FOV_Degrees": 90,
                            "Width": 1024,
                            "Height": 1024
                        }
                    ],
                    "X": 0.0,
                    "Y": 0.0,
                    "Z": 0.0,
                    "Pitch": 0.0,
                    "Roll": 0.0,
                    "Yaw": 0.0
                }
            }
        }
    }
}
```

with the following:
```python
import airsim
import numpy as np

client = airsim.VehicleClient(timeout_value = 7200)
client.confirmConnection()

responses = client.simGetImages([airsim.ImageRequest("my_camera", airsim.ImageType.Segmentation, False, False)])

img1d = np.fromstring(responses[0].image_data_uint8, dtype=np.uint8) #get numpy array
img_rgb = img1d.reshape(responses[0].height, responses[0].width, 3) #reshape array to 3 channel image array H X W X 3

print(np.unique(img_rgb.reshape(-1, img_rgb.shape[-1]), axis=0)) #red
```

With PR, we get the following BGR values: 
```
[[ 17 194 164]
 [103 216  44]
 [121   9 150]
 [205  72  54]
 [225  35  73]]
```

With 1.3.1 binaries, we get:
```
[[ 73 226 210]
 [170 237 115]
 [183  53 202]
 [232 145 127]
 [241 104 146]]
```

PR #2847, opened 2020-07-11
Extend Recording to multiple vehicles

rajat2004:
New Additions -

1. Info of all vehicles is recorded (with additional column `VehicleName`), `VehicleName` parameter in `Cameras` field to specify specfic vehcile to use the camera from
2. `Folder` field in settings to change the directory to store data in  

Example settings - 
```
{
    "SettingsVersion": 1.2,
    "SimMode": "Car",

    "Vehicles": {
        "Car1": {
          "VehicleType": "PhysXCar",
          "AutoCreate": true
        },
        "Car2": {
          "VehicleType": "PhysXCar",
          "AutoCreate": true,
          "X": 0, "Y": 10, "Z": 0
        }
    },
    "Recording": {
        "RecordOnMove": false,
        "RecordInterval": 1,
        "Cameras": [
            { "CameraName": "0", "ImageType": 0, "PixelsAsFloat": false, "VehicleName": "Car1", "Compress": true },
            { "CameraName": "0", "ImageType": 5, "PixelsAsFloat": false, "VehicleName": "Car1", "Compress": true },
            { "CameraName": "0", "ImageType": 0, "PixelsAsFloat": false, "VehicleName": "Car2", "Compress": true }
        ]
    }

}
```

Recording - [2020-07-11-11-48-51.zip](https://github.com/microsoft/AirSim/files/4906392/2020-07-11-11-48-51.zip)

Testing -
- Multiple Vehicles, with no recording settings (Scene image from each vehicle)
- Multiple Vehicles, specific sensor settings (Settings above)
- Single vehicle, no settings (Scene from that vehicle, former behaviour)

TODO:
- [x] Update docs
- [ ] Unity implementation

Further possible additions -
- [x] Allow user to specify directory to store recordings in (Replacement for `Documents`)
- [ ] Currently saving no images is not possible (For some reason this leads to an error in writing to file, even though the file has been created, and there are lines in it already)
      - I suspect this might be a race condition for the first write, since there are no images, no rendering and saving of images is required, so writing of pose data is happening almost instantaneously, but the file to write the data in hasn't still been created.(Haven't confirmed this though, just a thought)

Any testing (especially of `Folder` on Windows) or suggestions would be great!

Came to mind when working on #2841 
Related older issue #1346, Also includes the fix proposed in #1414

Issue #2307, opened 2019-11-10
HELP! Where to get PFM image data redirected into another path? 


jquitgra:
# Read This First
I am developing a deep-reinforcement learning algorithm for AI testing with a team, and need a neural network for classification of the PFM images on the quadcopter camera. We see that once we hit record, the images go to our Documents/Airsim directory. However, we are trying to understand where in the scripts provided in Airsim are these images getting directed to our documents folder, so we can get the most recent folder that holds the most recent frame images for further processing.

saihv:
https://github.com/microsoft/AirSim/blob/master/Unreal/Plugins/AirSim/Source/Recording/RecordingFile.cpp#L114

Issue #1414, opened 2018-09-22
The RecordInterval Problem


jinlukang1:
I set my RecordInterval:6.00, but when I check my AirSim File, I found that 8 images were recorded in one second of 6 seconds, and when I ran again, it became 4 images, and sometimes 1 and sometimes 2, why?

this is my setting.json
```
{
	"SettingsVersion": 1.2,
	"SimMode": "ComputerVision",
	"ViewMode": "",
	"Recording": {
		"RecordOnMove": false,
		"RecordInterval": 6.00,
		"Cameras": [
			{ "CameraName": "0", "ImageType": 0, "PixelsAsFloat": false, "Compress": true },
			{ "CameraName": "0", "ImageType": 5, "PixelsAsFloat": false, "Compress": true }
		]
	},
	"CameraDefaults": {
	"CaptureSettings": [
		{
		  "ImageType": 0,
		  "Width": 1440,
		  "Height": 900,
		  "FOV_Degrees": 90,
		  "AutoExposureSpeed": 100,
		  "MotionBlurAmount": 0
		},
		{
		  "ImageType": 5,
		  "Width": 1440,
		  "Height": 900,
		  "FOV_Degrees": 90,
		  "AutoExposureSpeed": 100,
		  "MotionBlurAmount": 0
		}
	]
	},
	"SegmentationSettings": {
	"InitMethod": "CommonObjectsRandomIDs",
	"MeshNamingMethod": "OwnerName",
	"OverrideExisting": false
  }
}
```

Can Someone help me?

sytelus:
Just tested this and it works as expected. At each interval one image for each camera is captured.

Relevant code: https://github.com/Microsoft/AirSim/blob/master/Unreal/Plugins/AirSim/Source/Recording/RecordingThread.cpp#L81

madratman:
@jinlukang1 was this resolved?

jinlukang1:
No, I don't know how to solve this problem yet.
So I just close it after I have recorded it So when I run it again, it's right again.

royuqp:
Hi, I found that this problem occurs when you begin play more than once on the editor and then start recording. Each time you begin play, "Cameras" setting of "Recording" written in settings.json is internally accumulated and used for recording.

The following simple source code modification will fix it :)

`static void loadRecordingSetting(const Settings& settings_json, RecordingSetting& recording_setting)`
`{`
`    recording_setting.requests.clear();    // <- inserted`
`    Settings recording_json;`
`...`

Issue #1346, opened 2018-08-24
Does the native recording feature support recording from all vehicles in a scene?


saihv:
The code in RecordingThread.cpp, does it handle multiple drones/cars?

saihv:
Answer: it doesn't. But it isn't hard to extend it to operate with multiple vehicles, by passing vectors of ImageCapture and Kinematics objects to the record thread (instead of a single object from vehicle 0).

arincbulgur:
@saihv 

I am using [kshitiz38](https://github.com/kshitiz38/AirSim)'s version to work with both a drone and ground vehicle. However, I also want to record the kinematics of multiple vehicles. I tried to use your modified RecordingThread.cpp in the [link](https://github.com/saihv/AirSim/blob/master/Unreal/Plugins/AirSim/Source/Recording/RecordingThread.cpp) to build the AirSim. It created the plugin bits without an error. However, when I tried to add the plugins into my unrealproject, it's unable to build the project.

Is your modified RecordingThread.cpp not compatible with the current APIs? Or should I also change some other codes? I am a beginner level of C++ user. Can you help me about this issue?

PR #2861, opened 2020-07-18
Extend Recording to multiple vehicles

### LINKED ISSUE #3029: How to start recording when game begins?
I want to start recording right after game begins rather than me push R button,
Is this achievable?

### LINKED ISSUE #3123: I want to know how to record without imageFile.
I use record button in AirSim because I want to record speed of the car and car position. 
However, record button ain AirSim also record the ImegeFile. So processing while driving is heavy.  
Please teach me how to record without imageFile.

### LINKED ISSUE #3214: Invalid uncomressed images when recording from Unreal editor
## Bug report 

- AirSim Version/#commit: 1.3.1
- UE/Unity version: 4.25.4
- autopilot version: N/A
- OS Version: Win10

Recording uncompressed images from Unreal editor using rec button, output invalid images which can't be open when set `"Compress": false`. 
When record compress images - it works fine.

Capture

### Settings

Example:
```
  "Recording": {
    "RecordOnMove": false,
    "RecordInterval": 1,
    "Cameras": [
        { "CameraName": "0", "ImageType": 0, "PixelsAsFloat": false, "Compress": false },
        { "CameraName": "0", "ImageType": 5, "PixelsAsFloat": false, "Compress": false }
    ]
  }

```

1. Add uncompress camera to recording section in setting file
2. Record from unreal editor using 'R'
3. Check the images folder

rajat2004:
Replaces #2847 (branch got messed up)

Closes #3029
Closes #3123
Closes #3214

Other related issues - #2307

New Additions -

1. Info of all vehicles is recorded (with additional column `VehicleName`), `VehicleName` parameter in `Cameras` field to specify specific vehicle to use the camera from
2. `Folder` field in settings to change the directory to store data in
3. Allow Recording with no images (just pose, etc data) (#3123)
4. `Enabled` parameter for starting recording when AirSim starts without needing to press R or call API (#3029) (Any naming suggestions are welcome)
5. Add method to write PPM file (for uncompressed images) (#3214)

Also fixes a bug when pressing Esc and Play multiple times causes requests to add in the vector and save multiple images (https://github.com/microsoft/AirSim/issues/1414#issuecomment-505741691)

Example settings - 
```json
{
    "SettingsVersion": 1.2,
    "SimMode": "Car",

    "Vehicles": {
        "Car1": {
          "VehicleType": "PhysXCar",
          "AutoCreate": true
        },
        "Car2": {
          "VehicleType": "PhysXCar",
          "AutoCreate": true,
          "X": 0, "Y": 10, "Z": 0
        }
    },
    "Recording": {
        "RecordOnMove": false,
        "RecordInterval": 1,
        "Cameras": [
            { "CameraName": "0", "ImageType": 0, "PixelsAsFloat": false, "VehicleName": "Car1", "Compress": true },
            { "CameraName": "0", "ImageType": 5, "PixelsAsFloat": false, "VehicleName": "Car1", "Compress": true },
            { "CameraName": "0", "ImageType": 0, "PixelsAsFloat": false, "VehicleName": "Car2", "Compress": true }
        ]
    }

}
```

Recording - [2020-07-11-11-48-51.zip](https://github.com/microsoft/AirSim/files/4906392/2020-07-11-11-48-51.zip)

Testing -
- Multiple Vehicles, with no recording settings (Scene image from each vehicle)
- Multiple Vehicles, specific sensor settings (Settings above)
- Single vehicle, no settings (Scene from that vehicle, former behaviour)

TODO:
- [x] Update docs
- [ ] Unity implementation

Further possible additions -
- [x] Allow user to specify directory to store recordings in (Replacement for `Documents`)
- [x] Currently saving no images is not possible (For some reason this leads to an error in writing to file, even though the file has been created, and there are lines in it already)
      - I suspect this might be a race condition for the first write, since there are no images, no rendering and saving of images is required, so writing of pose data is happening almost instantaneously, but the file to write the data in hasn't still been created.(Haven't confirmed this though, just a thought)
      - Fixed (set `is_ready` flag at the end)

Any testing (especially of `Folder` on Windows) or suggestions would be great!

Came to mind when working on #2841 
Related older issue #1346, Also includes the fix proposed in #1414

### DIFF
--- a/AirLib/include/common/AirSimSettings.hpp
+++ b/AirLib/include/common/AirSimSettings.hpp
@@ -51,12 +51,16 @@ public: //types
     };
 
     struct RecordingSetting {
-        bool record_on_move;
-        float record_interval;
+        bool record_on_move = false;
+        float record_interval = 0.05f;
 
-        std::vector<msr::airlib::ImageCaptureBase::ImageRequest> requests;
+        std::map<std::string, std::vector<ImageCaptureBase::ImageRequest> > requests;
 
-        RecordingSetting(bool record_on_move_val = false, float record_interval_val = 0.05f)
+        RecordingSetting()
+        {
+        }
+
+        RecordingSetting(bool record_on_move_val, float record_interval_val)
             : record_on_move(record_on_move_val), record_interval(record_interval_val)
         {
         }
@@ -395,14 +399,14 @@ public: //methods
         loadCameraDirectorSetting(settings_json, camera_director, simmode_name);
         loadSubWindowsSettings(settings_json, subwindow_settings);
         loadViewModeSettings(settings_json);
-        loadRecordingSetting(settings_json, recording_setting);
         loadSegmentationSetting(settings_json, segmentation_setting);
         loadPawnPaths(settings_json, pawn_paths);
         loadOtherSettings(settings_json);
         loadDefaultSensorSettings(simmode_name, settings_json, sensor_defaults);
         loadVehicleSettings(simmode_name, settings_json, vehicles);
 
-        //this should be done last because it depends on type of vehicles we have
+        //this should be done last because it depends on vehicles (and/or their type) we have
+        loadRecordingSetting(settings_json, recording_setting);
         loadClockSettings(settings_json);
     }
 
@@ -582,8 +586,10 @@ private:
             std::to_string(settings_json.getInt("CameraID", 0)));
     }
 
-    static void loadRecordingSetting(const Settings& settings_json, RecordingSetting& recording_setting)
+    void loadRecordingSetting(const Settings& settings_json, RecordingSetting& recording_setting)
     {
+        recording_setting.requests.clear();
+
         Settings recording_json;
         if (settings_json.getChild("Recording", recording_json)) {
             recording_setting.record_on_move = recording_json.getBool("RecordOnMove", recording_setting.record_on_move);
@@ -600,16 +606,22 @@ private:
                                 req_camera_settings.getInt("ImageType", 0));
                         bool compress = req_camera_settings.getBool("Compress", true);
                         bool pixels_as_float = req_camera_settings.getBool("PixelsAsFloat", false);
+                        std::string vehicle_name =req_camera_settings.getString("VehicleName", "");
 
-                        recording_setting.requests.push_back(msr::airlib::ImageCaptureBase::ImageRequest(
+                        recording_setting.requests[vehicle_name].push_back(ImageCaptureBase::ImageRequest(
                             camera_name, image_type, pixels_as_float, compress));
                     }
                 }
             }
         }
-        if (recording_setting.requests.size() == 0)
-            recording_setting.requests.push_back(msr::airlib::ImageCaptureBase::ImageRequest(
-                "", ImageType::Scene, false, true));
+
+        if (recording_setting.requests.empty()) {
+            // Add Scene image for each vehicle
+            for (const auto& vehicle : vehicles) {
+                recording_setting.requests[vehicle.first].push_back(ImageCaptureBase::ImageRequest(
+                    "", ImageType::Scene, false, true));
+            }
+        }
     }
 
     static void initializeCaptureSettings(std::map<int, CaptureSetting>& capture_settings)
--- a/AirLib/include/common/AirSimSettings.hpp
+++ b/AirLib/include/common/AirSimSettings.hpp
@@ -345,8 +345,6 @@ private: //fields
     float settings_version_actual;
     float settings_version_minimum = 1.2f;
 
-    bool vehicles_specfied = false;         // Whether "Vehicles" element was specified, used for Recording settings
-
 public: //fields
     std::string simmode_name = "";
     std::string level_name = "";
@@ -391,7 +389,6 @@ public: //methods
     {
         initializeSubwindowSettings(subwindow_settings);
         initializePawnPaths(pawn_paths);
-        initializeVehicleSettings(vehicles);
     }
 
     //returns number of warnings
@@ -412,7 +409,7 @@ public: //methods
         loadPawnPaths(settings_json, pawn_paths);
         loadOtherSettings(settings_json);
         loadDefaultSensorSettings(simmode_name, settings_json, sensor_defaults);
-        loadVehicleSettings(simmode_name, settings_json, vehicles, vehicles_specfied);
+        loadVehicleSettings(simmode_name, settings_json, vehicles);
 
         //this should be done last because it depends on vehicles (and/or their type) we have
         loadRecordingSetting(settings_json);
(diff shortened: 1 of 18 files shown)


rajat2004:
The last commit fixes a bug in the PR which causes it to not save any images when VehicleName wasn't specified.
Example settings -
```
{
    "SettingsVersion": 1.2,
    "SimMode": "Car",
    "Recording": {
        "RecordOnMove": false,
        "RecordInterval": 1,
        "Cameras": [
            { "CameraName": "0", "ImageType": 0, "PixelsAsFloat": false, "Compress": true }
        ]
    }
}
```
And here, if we change `Compress` to `false`, the images are corrupted and can't be opened, which is what I had mentioned in https://github.com/microsoft/AirSim/pull/2683#issuecomment-663806866

AirLib/include/common/AirSimSettings.hpp
```diff
@@ -566,47 +575,74 @@ struct AirSimSettings {
...
             std::to_string(settings_json.getInt("CameraID", 0)));
     }
 
-    static void loadRecordingSetting(const Settings& settings_json, RecordingSetting& recording_setting)
+    void loadDefaultRecordingSettings()
     {
+        recording_setting.requests.clear();
+        // Add Scene image for each vehicle
+        for (const auto& vehicle : vehicles) {
+            recording_setting.requests[vehicle.first].push_back(ImageCaptureBase::ImageRequest(
```
rajat2004, line 608: The strange part which I'll take a look at later, `vehicles` map here doesn't contain a single vehicle when no vehicles are specified, but instead 3 (one for each simmode), and why I added `getDefaultVehicleName`, rather than using the first vehicle in the map. 
Not sure why it's being done like this, shouldn't a single default vehicle be created in `initializeVehicleSettings()` instead of 3?

Can be seen by adding a line here, like `Utils::log(vehicle.first)`.

Unreal/Plugins/AirSim/Source/Recording/RecordingThread.cpp
```diff
@@ -92,7 +98,7 @@ bool FRecordingThread::Init()
     else {
         finishing_signal_.wait();
     }
-    if (image_capture_ && recording_file_)
+    if (recording_file_)
```
zimmy87, line 101: In the event that an empty UniqueValueMap is passed as the vehicle_sim_apis parameter to FRecordingThread::startRecording(), do we still want to log a success here considering image_captures_ will be empty as well? Also, this method is always returning true, signaling a successful initialization; it looks like this could result in a null pointer exception if FRecordingThread::Run() is called after this conditional fails. Don't we want to return false when recording_file_ is null?

zimmy87 (changes requested):
Thanks for submitting this PR @rajat2004! I have a few small comments mostly dealing with naming and style conventions that I'd like to see addressed before merging.

AirLib/include/common/AirSimSettings.hpp
```diff
@@ -378,7 +389,7 @@ struct AirSimSettings {
     {
         initializeSubwindowSettings(subwindow_settings);
         initializePawnPaths(pawn_paths);
-        initializeVehicleSettings(vehicles);
+        // initializeVehicleSettings(vehicles);
```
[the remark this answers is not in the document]
rajat2004: Forgot about this, removed

AirLib/include/common/common_utils/Utils.hpp
```diff
@@ -704,6 +704,35 @@ class Utils {
                 }
             }
         }
+
+        file.close();
+    }
+
+    static void writePPMfile(const uint8_t* const image_data, int width, int height, const std::string &path)
```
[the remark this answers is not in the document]
rajat2004: 1. I passed it by reference since that's the convention across most of the AirSim code, and prevents more copies so seemed like the better option. Changed the PFM method to pass by reference as well
2. Agree, renamed the method to `writePFMfile`, ti's a somewhat big change, across 6 files, so let me know if its better to leave it as is, the commit can be dropped easily

Unreal/Plugins/AirSim/Source/Recording/RecordingThread.cpp
rajat2004: Your point is valid, as to what'll happen if the map is empty, however that'll mean there aren't any vehicles, and in that case AirSim will probably panic quite early on itself. `FRecordingThread::Run()` won't run till `is_ready_` is set, which is done in `startRecording()` above at the end. This check should be sufficient I feel for the purpose intended to just print the message. Due to `is_ready_`,  there shouldn't be a problem with `recording_file` being null, since that'll be set later on when `startRecording()` is called, and returning `false` will cause UE4 to stop then itself.

I do remember that I wanted to keep the code intact as possible, so might have missed some problems. Will take a closer look again

zimmy87:
Hey @rajat2004, the 2 failing checks are due to a build break caused by a different PR. The fix for this build break was committed yesterday, can you rebase this PR to pick up the fix and rerun all the checks?

AirLib/include/common/AirSimSettings.hpp
zimmy87: Due to how ASimModeBase::setupVehiclesAndCamera() is defined, only vehicles that match the current simmode get instantiated. The only other place where the vehicles map gets read is in AirSimSettings::getVehicleSetting(), which only gets called in VehicleSimApiBase, and VehicleSimApiBase objects have a 1:1 relationship with the vehicles that get instantiated. As a result, I am confident the change to initializeVehicleSettings() to only initialize a single default vehicle is a safe change to make.

zimmy87 (approved):
tested this locally and it looks good to me; will discuss with my team about merging

zimmy87:
Thank you for the contribution @rajat2004!

rajat2004:
Glad to help! Contributing to AirSim has definitely helped me improve my skills, and hope that this will be useful to others as well :)

Ryton:
At my end following minimal example causes corrupted png file(s) for uncompressed images,  The file size scales with image size, so i assume that the data is generated, just not stored away correctly.

settings.json
{
    "SettingsVersion": 1.2,
    "SimMode": "Car",
    "Recording": {
        "RecordOnMove": false,
        "RecordInterval": 1,
        "Cameras": [
         { "CameraName": "0", "ImageType": 0, "PixelsAsFloat": false, "Compress": true},
	 { "CameraName": "0", "ImageType": 0, "PixelsAsFloat": false, "Compress": false}
        ]
    }
}    

Two images are generated, yet while the first is OK, the second one is corrupted.
Used Airsim v 1.2, map "Default". 
Perhaps this was patched in the mean while, in version 1.3 or 1.4?

uncompressed_corrupt_img_0_0_1616710947452891700
compressed_OK_img_0_0_1616710947446439600

"pngcheck -v uncompressed_corrupt_img_0_0_1616710947452891700.png" returns:
  File: uncompressed_corrupt_img_0_0_1616710947452891700.png (110592 bytes)
  this is neither a PNG or JNG image nor a MNG stream

rajat2004:
@Ryton This PR hasn't been included in the 1.4 release, you'll have to use the latest master

Ryton:
Ah ok, thanks! i was not aware of this. 
Will do then!

Ryton:
Fixed now, using the latest master 
AND 
using the correct field name/structure in the json...
the capturesettings for the camera should NOT be directly alongside their position,

so NOT:
```
... ,
   "Cameras": {
        "Camera1": {
  "X": 0.135, "Y": -2.15,"Z": -4.50,
            "Width": 1920 ,
            "Height": 1200 
},
```
But rather, within CameraSettings:
```
...,
"Cameras": {
 "Camera1": {
			"CaptureSettings": [
      {
          "SensorType": 0,
          "Enabled": true,
          "Width": 1920 ,
          "Height": 1200 
	  }],
            "X": 0.135, "Y": -2.15,"Z": -4.50,
```

PR #2865, opened 2020-07-20
Update docker_ubuntu.md to recommend a compatible version of ubuntu

GuillaumeLeclerc:
Numpy now require python 3.6, which is not available without extra ppa on ubuntu 16.04. Updating the documentation to make sure that we recommend a version that works with the Dockerfile provided


rajat2004:
I think the Dockerfiles and the python script should also be updated to use 18.04 by default

PR #2855, opened 2020-07-15
Fixes #2854: [UE4] simGetSegmentationObjectID will always return -1

LSBOSS:
This PR fixes #2854 .

Calling simGetSegmentationObjectID from Python and AirSim running in Unreal Engine 4 always fails with -1 because the underlying function `UAirBlueprintLib::GetMeshStencilID` is comparing the passed parameter not against the actual mesh's or owner's name but the name of the mesh component instead (e.g. StaticMeshComponent0). 

This PR fixes that by miming the same calls as the actual initialization of IDs in `UAirBlueprintLib::InitializeMeshStencilIDs`.

PR #2866, opened 2020-07-20
add az pipelines for windows and linux

madratman:
This PR adds CI with self-hosted VMs for Windows and Linux builds of AirSim. 
The main new addition is building and packaging of the Unreal Engine Blocks project for both OS'es, with UE 4.25.1. 

The VMs and pipelines results are in a private Azure DevOps organization right now, however if the build fails, a reasonable error message does get looped back to the github UI. We may change this in future if it turns to be problematic for users. 

Here's a little snippet of a linux build: 

Screenshot from 2020-07-20 14-19-48

Both linux and windows build take roughly 30 minutes

### DIFF
--- a/Unreal/Environments/Blocks/update_from_git_ci.bat
+++ b/Unreal/Environments/Blocks/update_from_git_ci.bat
@@ -1,0 +1,34 @@
+@echo off
+REM //---------- set up variable ----------
+setlocal
+set ROOT_DIR=%~dp0
+
+set AirSimPath=%1
+
+REM default path works for Blocks environment
+if "%AirSimPath%"=="" set "AirSimPath=..\..\.."
+
+IF NOT EXIST "%AirSimPath%" (
+	echo "AirSimPath %AirSimPath% was not found"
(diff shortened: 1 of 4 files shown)


madratman:
Contributers may trigger builds for existing open PRs via a `/azp run microsoft.AirSim` github comment, as can be seen [here](https://github.com/microsoft/AirSim/pull/2668#issuecomment-661490238), [here](https://github.com/microsoft/AirSim/pull/2682#issuecomment-661490146), or [here](https://github.com/microsoft/AirSim/pull/2855#issuecomment-661489423). 

I've put a bunch of PRs in the queue
Screenshot from 2020-07-20 17-41-18

rajat2004:
This will be very helpful for making sure it compiles completely on both Win & Ubuntu, thanks!
Just wondering, is it possible to allow downloading of the packaged Blocks binary for the master branch, and maybe even the PRs? This would be pretty useful for people wanting to test out new features without needing to fetch the branch and rebuild from source.

Issue #2728, opened 2020-05-26
AirSim: trying to simulate some sort of wind


nmuchaxo:
Hello, I'm new to UE4 and AirSim.
As the title says, I'm trying to simulate wind and the affects of it (even if it's not that accurate) on a quadcopter.
I've done some research and one way I can do this, is using radial force (if I'm following a bad path, please let me know) and try to use it to make the drone drift...
Other way is trying to apply a direct force (I'm not really sure), kinda like an impact or a collision, without making the drone stop and this force can be positive, negative and influence the drone's velocity...

Is there a better way to do this? Any recommendations, or has anyone done this? 
TIA

Issue #2342, opened 2019-12-11
AirSim - Automatic takeoff and landing training with wind and external forces using neural networks


xabierolaz:
Hi, I'm new to AirSim and would like to know where should I begin my research. I'd like to train drones inside AirSim using deep learning, so it can learn how to optimize landings and takeoffs with adverse climate (mainly wind), so we can then transfer all this data to the real drone.

Any hint about where to start would be greatly appreciated, thanks for your time!

xabierolaz:
bump

aiaarif:
Maybe you can take a look here on how to add wind. ->  #284

petergerten:
@xabierolaz did you get further with your project?

xabierolaz:
> 
> 
> @xabierolaz did you get further with your project?

No, had to switch to gazebo and openAI

Issue #2081, opened 2019-07-16
Do you Think Possible to Add wind like This?


prid1129:
Hi guys, I'm Trying to add wind force to Airsim UAV, but I thought it is too complicated to Follow #284 .

Instead, I'm trying to use AddForce function of UE4 to FlyingPawn.cpp/h . which is in the Plugin.

So my question is, is it possible? I'm not sure about this manner cause I'm just a beginner both c++ and UE4.

Thx :)

jonyMarino:
Treated in #284

Issue #284, opened 2017-06-15
How to add wind?


realflight1:
What would be the best / easiest way to apply wind to the simulated vehicle? Considering that I have the wind speed and direction.

realflight1:
Looking at `FastPhysicsEngine` it seems like one place to integrate wind would be in `getDragWrench` by modifying the actual vehicle velocity ([`vel_vertex`](https://github.com/Microsoft/AirSim/blob/593da156953a2514d7e02936f96f8752fb5b04af/AirLib/include/physics/FastPhysicsEngine.hpp#L206)). By modifying I mean adding the wind velocity vector in body frame to `vel_vertex`.

My knowledge of physics and aerodynamics is pretty basic, so I would appreciate any feedback. I used the following as a reference: [Does wind speed things up or accelerate things?](https://physics.stackexchange.com/questions/28827/does-wind-speed-things-up-or-accelerate-things)

sytelus:
You are almost there! Yes, you can just add wind vector to vehicle velocity to calculate final drag. However that's just one part of the story. Basically that simulates that if wind is towards vehicle then vehicle will feel resistance and reduce velocity. Other part of the wind effect is that wind can actually push vehicle. This is given by [very similar equation](http://www.engineeringtoolbox.com/wind-load-d_1775.html). Currently getDragWrench() would ignore the velocity vectors that do not have component towards vehicle (because it only computes drag). I would say you need another method getWindWrench() which computes wind pressure using formula from above link.

realdealneil:
Has this feature ever been implemented?  I need to simulate wind to make my controller more robust.

noNameForMe1994:
Is there anyway can I add the wind to the environment and get the real-time resistance or energy consumption of drone from airsim?

RVBldr:
Internal VSTF 7403

ardianumam:
Hi,
Is this already implemented yet?

Hiroskey:
Can I control the wind in real-time (i.e. dynamically)?

love-mj:
Has the effect of wind on UAV been realized？

rajat2004:
I've opened #2867 which adds Wind implementation according to the discussion here. Any testing or review, especially of the physics part would be great!

PR #2867, opened 2020-07-20
Add Wind simulation

rajat2004:
Adds force calculation due to Wind. Based on #284 

TODO:
- [x] API for setting wind
- [x] Unity Implementation (API tested to be working, haven't touched the settings part)
- [x] Check if this needs to be sent in PX4 or ArduPilot interface (Not needed mostly, AP Testing, seems okay - https://drive.google.com/drive/folders/1D5qsdtY308aQ9Fevr11y4ba6dMIzTf71?usp=sharing)

Settings:
```
{
    "SettingsVersion": 1.2,
    "SimMode": "Multirotor",
    "Wind": {
        "X": 0, "Y": 30, "Z": 0
    }
}
```

API:
```
wind = airsim.Vector3r(0, 25, 0)
client.simSetWind(wind)
```

Video: https://drive.google.com/file/d/1Cl1QOY_pD4SQd-EtuUNMMmxYDGDHZagI/view?usp=sharing
Script video - https://drive.google.com/file/d/1Gx4zCSnifbVN3I-32583dM6GTiQP0ZLT/view?usp=sharing

Would be great if others can test it out as well. Review of the physics implementation would be very much appreciated, haven't dabbled in this part of the codebase earlier

Related issues: #284, #2081, #2728, #2342


nicocarb (commented):
I have implemented payload dynamics (single point of attachment to a cable) in AirSim and implemented it very similarly to how you have done. Can't speak to the exact physics of the wind, but it looks good.

rajat2004:
@nicocarb Thanks for the review! Payload attachment certainly seems interesting, would be nice to see the implementation if you're able to
I'm going to do some testing with ArduPilot, and I think those logs will be able to give more idea about the physics, maybe not to me, but definitely other much more experienced members of the AP community. One thing which is a bit strange is the vehicle remains tilted in the wind, maybe cause the wind to too high?

nicocarb:
@rajat2004 I apologize for the late response. I could perhaps explain my payload implementation. Not sure if I can show you my full code, however. I essentially created a payload object that has a volume, drag faces, a mass, and a cable length. Then, I take information about the current angle of the payload, current rate of change of angle of the payload, and drag force on the payload and use a small angle approximation to solve for the payload position in time in the World Frame relative to the drone's attachment point. Then, I apply a force vector equivalent to the tension force created by the payload in the direction of the payload position. Small angle approximations were appropriate for my use case; however, if small angle approximations are not valid, perhaps exploring a Euler approximation using payload acceleration and velocity may be successful. I was not able to get this to act appropriately. Most of the actual calculations occur in getNextKinematicsNoCollision(), which I've overriden. Perhaps we can discuss showing you an implementation of this from my side (I am just not sure as of right now whether I can). 

In terms of the wind, I imagine it must be too high. For a drone to fly against wind, attempting to hold its positions, its thrust vectors must attempt to counteract the wind force. The video you've attached seems to suggest this is what the controller is attempting to do. I may be incorrect, however.

rajat2004:
@nicocarb No problem at all, I was just interested in getting an overview on the implementation. Thanks for the explanation, looks pretty complex, great that you got it working.
Re the wind, yeah, the values I'm using are quite high, low values didn't get much visual response. From the ArduPilot testing (Drive link in description) as well, the drone bounces back to the original orientation when wind is reset to 0. Also, the tilt seems slightly lesser in AP than Simple flight, and bounce back isn't also that much sudden, which could be due to more sophisticated control in AP. Still need to do some log analysis, an AP member did have a look at the physics and seemed okay.

samuelctabor:
I think the physics implementation could be improved. Right now drag and wind are dealt with separately in `getWindWrench` and `getDragWrench` using the vehicle and wind velocities respectively. In reality, wind can only affect a vehicle through drag, and the only velocity needed for drag calculation is the velocity of the vehicle _relative to the air_.

The correct way to calculate the effect of wind would be to feed it into getDragWrench and subtract/add it (depending on convention) from the vehicle linear velocity `linear_vel` to give the velocity of the vehicle relative to the air.

rajat2004:
@samuelctabor Thanks for the review! Yes, I think you're correct, using the relative velocity does make more sense.
I've added a commit which makes those changes, could you have a look? This also reduces the code changes required by a lot.

Here's a video with SimpleFlight - https://drive.google.com/file/d/1ydMt8BFMTWRvJiTEuEByY6bfJmfswR4Z/view?usp=sharing
This however required reducing the wind speed in the example script, otherwise it would crash repeatedly to the ground when 25m/s speed was applied. So there's a behaviour change, was the previous implementation incorrect, or something more needs fixing?

samuelctabor:
@rajat2004 the previous implementation was incorrect :-)

25m/s is a lot - over 55mph - your new values are more realistic.

Note that `getWindWrench` can now be removed.

rajat2004:
Yup, the new behaviour did make more sense, but wanted to confirm. The `getWindWrench` method was kept for reference, cleaned up now.
Thanks a lot for the review @samuelctabor!

ibrahim0v:
hey @rajat2004 , I tried to test out your implementation. I keep getting this when I run simSetWind():

msgpackrpc.error.RPCError: rpclib: function 'simSetWind' (called with 1 arg(s)) threw an exception. The exception is not derived from std::exception. No further information available.

I could not trace it back. Could you guess my mistake?

rajat2004:
Quite strange, haven't seen that before.
You're running on Windows? Could you post the settings used, and maybe try the example script as well

ibrahim0v:
Okay my bad :) It was a silly mistake. Seems like I need some coffee :DD

rajat2004:
Ok, great that it got fixed. Still, curious to know what caused the problem :)

ibrahim0v:
I have given a tuple as the argument of airsim.Vector3r function. So instead of passing windX,windY,windZ I have passed (windX,windY,windZ)

ibrahim0v:
BTW, wind physics is beautiful but can be adjusted to me more realistic. I can run tests for my drone now. Thanks for your contribution. 
A.I: I have tested this on PX4 1.9.2 SITL

ibrahim-aydin (approved):
I tested it on the following environment:
1. Unreal Engine 4.24
2. PX4 1.9.2 SITL
3. AirSim 1.2.0 (+ some updates from 1.3.1 and a few fixes + my own customizations)

Couldn't face any problems. Wind implementation is great and fine for me now. Good job @rajat2004

rajat2004:
Great, thanks for the testing!
Agree, the physics could be improved, probably the job for someone more knowledgeable and motivated

rajat2004:
Thanks for merging! Probably should have marked this in the PR itself using keywords, but some issues can be closed now I think - #284, #2081, #2728, #2342

hebai-github:
@rajat2004 The wind implementation is great! Is it possible to create location-based injection of wind, i.e., at different locations, different winds are injected?

rajat2004:
An interesting idea, there currently isn't any such location-based separation of effects anywhere in the code AFAIK, a possible way could be to store multiple wind profiles for areas based on the coordinates, and then lookup the specfic area every time. Any another ideas which might work?
I mostly won't be able to work on this any time soon, any contributions are more than welcome!

hebai-github:
I was thinking to add another function in the physics engine, which takes the wind velocity and the corresponding location information as the input argument. If the drone's location is close to the input location, the wind velocity takes an effect. otherwise a default wind velocity is used. The wind profiling can be at the client side which looks up the wind velocity based on the vehicle's position. This way i think we can affect multiple drones with different wind velocities. 

I would like to give a try anyway. but i am very new to AirSim and a novice C++ programmer. Any advice would be appreciated.

rajat2004:
Sure, that's another way, it all depends on the goal and the kind of effect to be achieved. You'll have to set some threshold to decide the barrier around the location as well.
The files changed in this PR will give an idea on what files you'll need to modify, the file structure naming and all is fairly straightforward, plus other commits and PRs can give pointers on which files and lines need to be touched

aklein1995:
I have doubts with the wind implementation. Considering the example provided at:
https://github.com/Microsoft/AirSim/blob/main/PythonClient/multirotor/set_wind.py

As far as I understood, when you apply 10m/s in forward direction, if we do nothing at code (e.g., time.sleep(10)), we should expect the drone to move forward during 10 seconds. 

I am confused, because instead of seeing such behavior, the drone does not move. In fact, the drone does not move when you modify either the X or Y dimensions with values between 0 and 20 m/s. For example, for 15m/s in X axis:
`client.simSetWind(airsim.Vector3r(15, 0, 0))`
When these values are above 20m/s, the drone seems to try to stabilize. Similarly, in Z dimension this issue also exist but lower values seems to make the drone move (e.g., 10m/s works).

Am I missing something? Am I missing something? Does the drone have a way to stay stable on its own without coding it?

Lesh97:
I use Unity Airsim.
add settings.json Wind :{"X":0,"Y":25,"Z":0}
vehicle is px4multirotor
but wind is not working 
anyone can solve this? how to use it for unity?

PR #2873, opened 2020-07-22
Install CMake only if required

rajat2004:
CMake was being installed even if already present on Ubuntu 18.04

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -57,15 +57,18 @@ fi
 
 #give user perms to access USB port - this is not needed if not using PX4 HIL
 #TODO: figure out how to do below in travis
+# Install additional tools, CMake if required
 if [ "$(uname)" == "Darwin" ]; then # osx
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -74,46 +77,46 @@ else #linux
...
+        if [ "$(lsb_release -rs)" == "18.04" ]; then
+            sudo apt-get -y install \
+                apt-transport-https \
+                ca-certificates \
+                gnupg
+            wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | gpg --dearmor - | sudo tee /etc/apt/trusted.gpg.d/kitware.gpg >/dev/null
+            sudo apt-add-repository 'deb https://apt.kitware.com/ubuntu/ bionic main'
+            sudo apt-get -y install --no-install-recommends \
+                make \
+                cmake
```
rajat2004, line 95: We could just install from normal apt repositories if >=18.04 since they'll have recent cmake versions

PR #2874, opened 2020-07-22
[ros] fix simSetCameraPose bug

madratman:
I merged a bug accidentally while merging 2710. Should have waited for travis result.


rajat2004:
Sorry for the bug, and thanks for fixing! Not sure as to how this got in, I had checked the build earlier...
But good to see that CI is working as intended

PR #2881, opened 2020-07-26
Disable cameras after fetching images, projection matrix

### LINKED ISSUE #1766: Calling simGetImage or simGetImages once drops performance permanently. Why?
Calling simGetImage or simGetImages from python or C++ permanently impacts rendering framerate. In my case I am running at 60fps generally and can communicate with unreal via other airsim lightweight APIs without framerate costs.

As soon as simGetImage or simGetImages is called, the framerate drops to 20fps on my machine. Even if I call simGetImage once, the framerate is reduced. Can you explain where in the code the permanent cost is paid? You must be scheduling simGetImage tasks continuously in Unreal as RenderTasks, are you? Why does the performance drop permanently?

The recording button ( because it calls the same api ) will also drop performance at the same rate.

I understand that you optimized the performance ( ReadSurfaceData and not ReadPixels ) to bring it up to the current level, as you mention here:
https://stackoverflow.com/questions/43218678/unreal-engine-accessing-saving-in-game-images-to-disk-without-blocking-game-thr/43287505

Also interesting to mention that the actual performance drop is the same whether you are capturing several camera views or just one, which makes me feel that the cost is paid out of render thread management.

### LINKED ISSUE #2389: Performance impact after simGetImages() finished
Hi,

I have a camera with 640x480 resolution configured and use simGetImages() in a loop getting ~40fps.
All good.
Now I added a second camera with all the same characteristics but a very high resolution: 5000x4000.
This is to mimic a still image function.

After I call simGetImages() on the high res camera once, the main camera loop framerate drops from ~40 fps to ~1fps.
While a drop during high-res image capture is expected - it never recovers from this and stays at ~1 fps.

Does anybody has an idea what is happening here?

Peter

### LINKED ISSUE #2464: Add function to disable a camera via API
Currently all configured cameras become active as soon as they are used.
They will stay active even if they are not used and degrade the performance unnecessary.

In interactive mode they can be toggled on and off via key - this is not the case in headless mode or if the API is used.

There should be a function to deactivate a camera (and maybe to activate - or it activates automatically when an image is requested).

Related: https://github.com/microsoft/AirSim/issues/2389

rajat2004:
Replaces #2465 by doing that internally

Before -
```
Time spent for 30 frames: 0:00:01.302084
Time spent for 30 frames: 0:00:01.319725
Time spent for 30 frames: 0:00:01.312358
Time spent for 30 frames: 0:00:01.333144
High resolution image captured.
Time spent for 30 frames: 0:00:01.675652
Time spent for 30 frames: 0:00:04.904762
Time spent for 30 frames: 0:00:04.817640
Time spent for 30 frames: 0:00:04.800456
Time spent for 30 frames: 0:00:04.828672
High resolution image captured.
Time spent for 30 frames: 0:00:05.155805
Time spent for 30 frames: 0:00:04.922219
```

After -
```
Time spent for 30 frames: 0:00:01.280212
Time spent for 30 frames: 0:00:01.275583
Time spent for 30 frames: 0:00:01.258539
Time spent for 30 frames: 0:00:01.270891
High resolution image captured.
Time spent for 30 frames: 0:00:01.641478
Time spent for 30 frames: 0:00:01.654431
Time spent for 30 frames: 0:00:01.611860
Time spent for 30 frames: 0:00:01.590966
Time spent for 30 frames: 0:00:01.578362
High resolution image captured.
Time spent for 30 frames: 0:00:01.901373
Time spent for 30 frames: 0:00:01.706202
Time spent for 30 frames: 0:00:01.683336
Time spent for 30 frames: 0:00:01.614648
Time spent for 30 frames: 0:00:01.599740
High resolution image captured.
Time spent for 30 frames: 0:00:01.964565
```

Not sure as to what is causing the still present increase in time, but still much better than earlier
Tested whether disabling camera affects subwindow, no effect
Haven't tested the projection matrix one though

Closes #1766, closes #2389, closes #2464

### DIFF
--- a/PythonClient/multirotor/high_res_camera.py
+++ b/PythonClient/multirotor/high_res_camera.py
@@ -1,0 +1,61 @@
+import airsim
+from datetime import datetime
+
+'''
+Simple script with settings to create a high-resolution camera, and fetching it
+
+Settings used-
+{
+    "SettingsVersion": 1.2,
+    "SimMode": "Multirotor",
+    "Vehicles" : {
+        "Drone1" : {
+            "VehicleType" : "SimpleFlight",
+            "AutoCreate" : true,
+            "Cameras" : {
+                "high_res": {
+                    "CaptureSettings" : [
+                        {
+                            "ImageType" : 0,
+                            "Width" : 4320,
+                            "Height" : 2160
+                        }
+                    ],
+                    "X": 0.50, "Y": 0.00, "Z": 0.10,
+                    "Pitch": 0.0, "Roll": 0.0, "Yaw": 0.0
+                },
+                "low_res": {
+                    "CaptureSettings" : [
+                        {
+                            "ImageType" : 0,
+                            "Width" : 256,
+                            "Height" : 144
+                        }
+                    ],
+                    "X": 0.50, "Y": 0.00, "Z": 0.10,
+                    "Pitch": 0.0, "Roll": 0.0, "Yaw": 0.0
+                }
+            }
+        }
+    }
+}
+'''
+
+client = airsim.VehicleClient()
+client.confirmConnection()
+framecounter = 1
+
+prevtimestamp = datetime.now()
+
+while(framecounter <= 500):
+    if framecounter%150 == 0:
+        client.simGetImages([airsim.ImageRequest("high_res", airsim.ImageType.Scene, False, False)])
+        print("High resolution image captured.")
+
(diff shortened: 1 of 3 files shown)


rajat2004:
Follow up, this seems to be causing crashes when using the API continuously, strange that it didn't appear with my testing and others as well. I've opened #3039 which reverts this, but again needs testing to confirm

Issue #2917, opened 2020-08-06
Bug on Multirotor "moveOnPath"


ppueyor:
We discovered a bug in moveOnPath. The last segment of the path was not taken into account, so the drone didn't move to the last position on the segment list. We fixed it in our private repository.

The file that needs to be fixed is "MultirotorApiBase.cpp". 

Should we upload the new file here or we just need to indicate the lines and fixes done?

rajat2004:
A PR if possible would probably be the best way!

zimmy87:
This issue is by design; see my comment [here](https://github.com/microsoft/AirSim/pull/2919#issuecomment-738271735).

PR #2919, opened 2020-08-07
Fixed Quadrotor moveOnPath

ppueyor:
As explained in #2917 , the last segment of path of moveOnPath (that affects to more functions to move the drone) was not taken into account so the drone did not move if the stated distance was low (and bigger than the lookahead_error).

By adding these two conditions to the loop, the bug seems to be fixed.

### DIFF
--- a/AirLib/src/vehicles/multirotor/api/MultirotorApiBase.cpp
+++ b/AirLib/src/vehicles/multirotor/api/MultirotorApiBase.cpp
@@ -269,7 +269,7 @@ bool MultirotorApiBase::moveOnPath(const vector<Vector3r>& path, float velocity,
     float goal_dist = 0;
 
     //until we are at the end of the path (last seg is always zero size)
-    while (!waiter.isTimeout() && (next_path_loc.seg_index < path_segs.size()-1 || goal_dist > 0)
+    while (!waiter.isTimeout() && !waiter.isComplete() && (next_path_loc.seg_index < path_segs.size() || goal_dist > 0)
         ) { //current position is approximately at the last end point
 
         float seg_velocity = path_segs.at(next_path_loc.seg_index).seg_velocity;


rajat2004:
Is this related to #2324? Since both are modifying the same lines
Will try to test it out, a script which shows the problem would be great!

ppueyor:
I don't think that is related to #2324 . 

Which kind of script would you need?

rajat2004:
Nothing fancy as such, just a simple script with some waypoints which shows that the drone doesn't carry out the last part of the path. If you want, you could even add it in the `Pythonclient/multirotor` folder and push to this PR, this adds new examples for others to use and test

ppueyor:
You can find a simple script attached.
It is a simple C++ program where the drone should move to a new position, whose transform is drawn.

In this case if z < 0.75 (lookahead), the drone won't move (because the last part of the segment is not completed).
With the fix of this PR, the drone correctly moves to the stated position.


[move_to_z_fix.zip](https://github.com/microsoft/AirSim/files/5068129/move_to_z_fix.zip)

zimmy87:
I'm having an issue with this PR where MultirotorApiBase::moveOnPath gets stuck in an infinite loop during a call to moveToPositionAsync. Also See https://github.com/microsoft/AirSim/blob/b8a06accd1f5e50d77eb10d7046042b147c0da90/AirLib/src/vehicles/multirotor/api/MultirotorApiBase.cpp#L805-L810

The logic here looks like it depends upon setting the next path segment to the last segment to signal to the while loop in MultirotorApiBase::moveOnPath to terminate. Additionally, the last segment should always be a zero length segment due to https://github.com/microsoft/AirSim/blob/b8a06accd1f5e50d77eb10d7046042b147c0da90/AirLib/src/vehicles/multirotor/api/MultirotorApiBase.cpp#L263-L265

I attached the python script I used to reproduce this infinite loop in case you'd like to run the scenario yourself
[repro2917.zip](https://github.com/microsoft/AirSim/files/5632384/repro2917.zip)

zimmy87:
The original issue being addressed here is very similar to https://github.com/microsoft/AirSim/issues/3106 . The move* APIs are not intended to move the drone to an exact location, they are meant to move the drone to a area around the specified target within a certain distance. This is to simulate real-world behavior where drones have a certain margin of error when moving.

jonyMarino:
Closed due to lack of response. Feel free to ask to reopen the pull request if you can answer our questions.
Thanks!

PR #2920, opened 2020-08-07
Add support for Distance Sensor in ArduPilot

rajat2004:
Changes - 

1. Support for Distance Sensor in ArduPilot, also increase RC channels to 12
2. Fix style inconsistency, use unique_ptr in ArduPilot files
3. Unity: Update to C++14 standard

Win Blocks binary - https://drive.google.com/file/d/1eLaC58nnh0ms9BhHqvVMnXo9o_xiH3pr/view?usp=sharing

TODO:
- [x] ArduRover Implementation
- [x] Documentation in ArduPilot Wiki (AP Wiki PR - https://github.com/ArduPilot/ardupilot_wiki/pull/2948)

Corresponding AP PR - https://github.com/ArduPilot/ardupilot/pull/15007
Follow-up after #2807

Code duplication, any ideas on how this can be reduced?

Second-last commit fixes a style inconsistency, camel case is not used for variables there, also no idea why I used `shared_ptr`, maybe a remanent of earlier copy-paste from PX4 backend.
Last one updates C++ standard used in Unity to C++14, required for `make_unique` (see https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique), otherwise [this happens](https://travis-ci.com/github/rajat2004/AirSim/builds/182056675)

However, AirLib is built with C++17 standard in UE, which can cause problems later on. Should it be updated further to 17? (Tested it to be working locally)

### DIFF
--- a/AirLib/include/vehicles/car/firmwares/ardurover/ArduRoverApi.hpp
+++ b/AirLib/include/vehicles/car/firmwares/ardurover/ArduRoverApi.hpp
@@ -186,7 +186,34 @@ private:
 
         std::ostringstream oss;
 
-        const uint count_lidars = getSensors().size(SensorBase::SensorType::Lidar);
+        // Send Distance Sensors data if present
+        const uint count_distance_sensors = sensors_->size(SensorBase::SensorType::Distance);
+        if (count_distance_sensors != 0) {
+            // Start JSON element
+            oss << ","
+                    "\"rng\": {"
+                    "\"distances\": [";
+
+            // Add the first sensor separately since otherwise there's an extra ","
+            // Which leads to AP counting an extra sensor, and if there are 10 sensors (AP supports upto 10), last one will get ignored
+            auto distance_output = getDistanceSensorData("");
+            oss << distance_output.distance;
+
+            // Add remaining sensors in the array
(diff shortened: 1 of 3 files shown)


rajat2004:
@madratman @saihv @ironclownfish @jonyMarino Just asking, would it be possible to have the packaged Blocks binary from the Azure Pipelines available to download? Described here https://docs.microsoft.com/en-us/azure/devops/pipelines/artifacts/artifacts-overview?view=azure-devops (I haven't myself gone through it completely)

rajat2004:
> @madratman @saihv @ironclownfish @jonyMarino Just asking, would it be possible to have the packaged Blocks binary from the Azure Pipelines available to download? Described here https://docs.microsoft.com/en-us/azure/devops/pipelines/artifacts/artifacts-overview?view=azure-devops (I haven't myself gone through it completely)
> Also mentioned here - [#2866 (comment)](https://github.com/microsoft/AirSim/pull/2866#issuecomment-661624991)

Just a follow-up, could an API call like this work - https://docs.microsoft.com/en-us/rest/api/azure/devops/build/artifacts/get?view=azure-devops-rest-4.1?
Need to figure out the variables though, and mostly needs access

rajat2004:
Replaced by #3364

PR #2935, opened 2020-08-14
Update build_linux.md

thatPRT:
removed _build.sh_ from docs for Linux install of Unreal engine since it is no longer present

### DIFF
--- a/docs/build_linux.md
+++ b/docs/build_linux.md
@@ -51,8 +51,6 @@ cd AirSim
 
 ```bash
 ./setup.sh
-./build.sh
-# use ./build.sh --debug to build in debug mode
 ```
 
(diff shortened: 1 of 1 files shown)


docs/build_linux.md
```diff
@@ -51,8 +51,6 @@ cd AirSim
 
 ```bash
 ./setup.sh
-./build.sh
-# use ./build.sh --debug to build in debug mode
```
rajat2004, line 55: Really don't understand this change, `build.sh` is required, and is present in the repo - https://github.com/microsoft/AirSim/blob/master/build.sh

madratman:
Closing this. `build.sh` is definitely needed

Issue #2921, opened 2020-08-10
How to obtain the data information of all vehicles in the multi-vehicle state？


LittleTimeWxk:
Dear friends，
    Now I create many cars in simulator and I want to put the datas of them into a list, then I can get the information of the car one by one using python API.
    
     **My Question：How to get the data of all the vehicles created？**

     I don't understand the code structure of vehicle creation in codes, so I won't use it.Could you provide some help？Tutorials, suggestions and ideas are all available！

rajat2004:
Most of the APIs take in a `vehicle_name` parameter, which should be the name specified on the settings
Not passing the parameter uses the first vehicle

LittleTimeWxk:
> Most of the APIs take in a `vehicle_name` parameter, which should be the name specified on the settings
> Not passing the parameter uses the first vehicle

Thanks your reply. But I think, may be you can't understand my question.

rajat2004:
Well, then maybe you should have explained it better
See the documentation - https://microsoft.github.io/AirSim/multi_vehicle/

LittleTimeWxk:
> Well, then maybe you should have explained it better
> See the documentation - https://microsoft.github.io/AirSim/multi_vehicle/

OK! Thank you for your reminder.

LittleTimeWxk:
> Well, then maybe you should have explained it better
> See the documentation - https://microsoft.github.io/AirSim/multi_vehicle/

Hi! I have read the documentation, but that's not what i want. I want get the status of all the car one-time. 
Now my plan is:
step1: I can get the name of all control car, and put it to a list;
step2: I can get the status of the corresponding vehicle by name list, and put status to a new list;
step3: I will post the state list to a ros point.

My problem is I don‘t know how to get all the names of the car in C++ backend in the first step. Because There are many things I don’t understand in the code project.
Could you give me some advice and other ideas?

rajat2004:
Well, you're writing the settings file, so you would already know what are the names of the vehicles. An advanced step could be to use a JSON parser and parse the settings yourself. 
This does bring up a possibility of adding a new API to give a list of the vehicles, which could be useful especially if and when https://github.com/microsoft/AirSim/pull/2390 gets in (needs update and lots of testing though)
If you're using ROS, have a look at https://microsoft.github.io/AirSim/airsim_ros_pkgs/ also, it might contan what you need

LittleTimeWxk:
[...]

PR #2936, opened 2020-08-14
Add API for list of vehicle names

rajat2004:
Settings -
```json
{
    "SettingsVersion": 1.2,
    "SimMode": "Car",

    "Vehicles": {
        "Car1": {
          "VehicleType": "PhysXCar"
        },
        "Car2": {
          "VehicleType": "PhysXCar",
          "X": 0, "Y": 10, "Z": 0
        }
    }
}
```
API -
```
>>> client.listVehicles()
['Car1', 'Car2']
>>> 
```

Came to mind in #2921, just wrote it since was straightforward to implement

The keys contain an empty string, coming from it being used to refer to the default vehicle, see [ApiProvider.hpp](https://github.com/microsoft/AirSim/blob/master/AirLib/include/api/ApiProvider.hpp#L67) being called from [SimModeBase.cpp](https://github.com/microsoft/AirSim/blob/master/Unreal/Plugins/AirSim/Source/SimMode/SimModeBase.cpp#L589)
A simple way is to just remove the empty string ourselves, which is being done right now.
However, the API could also mean - return a list of _valid vehicle names which can be passed to APIs_, in which case empty string is valid. But it could also introduce more complexity to the client code, such as removing the empty string if data of each vehicle is required only once, and having an empty string will duplicate calls for one vehicle.
I'm leaning towards removing the empty string, and will add that behaviour in the PR. But do comment if the second behaviour makes more sense. 

Could be particularly useful if #2390 goes in (which I hope to update in the next few days)
Not much of an use-case otherwise, just allows the user to avoid parsing the settings for getting the vehicle names

### DIFF
--- a/AirLib/include/api/RpcLibClientBase.hpp
+++ b/AirLib/include/api/RpcLibClientBase.hpp
@@ -129,6 +129,7 @@ namespace airlib
         bool isRecording();
 
         void simSetWind(const Vector3r& wind) const;
+        vector<string> listVehicles();
 
         std::string getSettingsString() const;
 
--- a/AirLib/include/api/WorldSimApiBase.hpp
+++ b/AirLib/include/api/WorldSimApiBase.hpp
@@ -82,6 +82,7 @@ namespace airlib
         virtual bool isRecording() const = 0;
 
         virtual void setWind(const Vector3r& wind) const = 0;
+        virtual vector<string> listVehicles() const = 0;
 
         virtual std::string getSettingsString() const = 0;
     };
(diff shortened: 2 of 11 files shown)


rajat2004:
Rebased on master, conflict in every file (ouch), didn't go well with the Wind PR
But seems like both Travis and Azure Pipelines have stopped, would be good to have that looked at, don't want to go back to when master didn't compile on different platforms (which I myself caused a few times :sweat_smile: )
Ping @saihv @madratman

zimmy87:
Hey @rajat2004, can you resolve the current conflicts to get this PR ready for merging?

zimmy87:
Tested Unreal & Unity builds, API worked as expected in both builds, so I am moving ahead with merging this.

PR #2942, opened 2020-08-17
Update Docker scripts, info

rajat2004:
Docker version >= 19.03 natively supports Nvidia GPUs, and there's no need to run using `nvidia-docker2` which has been deprecated. Update the script to use the correct command as required, also update some links for installation

This currently uses OpenGL, which is deprecated
Also using `-- headless` causes an error -
```
[2020.08.19-04.58.16:458][  0]LogHAL: Warning: Splash screen image not found.
[2020.08.19-04.58.16:461][  0]LogSlate: New Slate User Created.  User Index 0, Is Virtual User: 0
[2020.08.19-04.58.16:461][  0]LogSlate: Slate User Registered.  User Index 0, Is Virtual User: 0
[2020.08.19-04.58.16:462][  0]LogInit: Using SDL_WINDOW_OPENGL
Signal 11 caught.
Malloc Size=65538 LargeMemoryPoolOffset=65554 
CommonUnixCrashHandler: Signal=11
Malloc Size=65535 LargeMemoryPoolOffset=131119 
Malloc Size=107920 LargeMemoryPoolOffset=239056 
Failed to find symbol file, expected location:
"/home/rajat/AirSim/docker/Blocks/Blocks/Binaries/Linux/Blocks.sym"
```

### DIFF
--- a/docker/download_blocks_env_binary.sh
+++ b/docker/download_blocks_env_binary.sh
@@ -1,4 +1,9 @@
-sudo apt-get install unzip
-wget -c https://github.com/Microsoft/AirSim/releases/download/v1.2.0Linux/Blocks.zip
+#!/bin/bash
+
+if ! which unzip; then
+    sudo apt-get install unzip
+fi
+
+wget -c https://github.com/microsoft/AirSim/releases/download/v1.4.0-linux/Blocks.zip
 unzip Blocks.zip
 rm Blocks.zip
--- a/docker/download_blocks_env_binary.sh
+++ b/docker/download_blocks_env_binary.sh
@@ -1,9 +1,10 @@
 #!/bin/bash
+set -x
 
 if ! which unzip; then
     sudo apt-get install unzip
 fi
 
-wget -c https://github.com/microsoft/AirSim/releases/download/v1.4.0-linux/Blocks.zip
-unzip Blocks.zip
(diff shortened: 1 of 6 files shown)


docker/download_blocks_env_binary.sh
```diff
@@ -1,4 +1,9 @@
-sudo apt-get install unzip
-wget -c https://github.com/Microsoft/AirSim/releases/download/v1.2.0Linux/Blocks.zip
+#!/bin/sh
+
+if ! which unzip; then
+    sudo apt-get install unzip
+fi
+
+wget -c https://github.com/microsoft/AirSim/releases/download/v1.3.1-linux/Blocks.zip
```
rajat2004, line 7: It'll be better to make it such that it fetches the latest Blocks release

madratman:
In 4.25, the headless + vulkan error is fixed via the `-RenderOffScreen` arg.

rajat2004:
Thanks for the info, will test it out soon

docker/download_blocks_env_binary.sh
jonyMarino: Hi Rajat! We can add a tag for this, but personally, I prefer to have an image with fixed versions we know that work.

jonyMarino:
No. The changes seem right. Let's be prepare for support in case someone finds something.
Thanks for all your work Rajat!

PR #2953, opened 2020-08-20
[Travis] Add Ubuntu 20.04, OSX XCode 11.5 jobs

rajat2004:
No documentation changes until tested, currently Focal build fails due to a Python package `trollius` not working on Python 3.8, used in ROS somewhere

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,10 @@ jobs:
       os: linux
       dist: bionic
 
+    - name: Ubuntu 20.04
+      os: linux
+      dist: focal
+
     - name: Ubuntu 16.04
       os: linux
       dist: xenial
@@ -16,9 +20,13 @@ jobs:
         - MSBUILD_PATH="c:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\MSBuild\16.0\Bin"
         - PATH_FIX="C:\Windows\System32"
 
-    - name: MacOS
+    - name: XCode 11.4.1
       os: osx
       osx_image: xcode11.4
+
+    - name: XCode 11.5
+      os: osx
+      osx_image: xcode11.5
 
 
 before_install:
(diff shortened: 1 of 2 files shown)


saihv:
catkin-tools in Noetic should be using asyncio instead of trollius as it natively works with Python 3. 

https://github.com/catkin/catkin_tools/pull/574

Do you have any error logs relating to the trollius errors?

rajat2004:
Ahh yes, sorry, forgot to put a link to my Travis build, here's the currently running one - https://travis-ci.com/github/rajat2004/AirSim/builds/187740417
And the build before that - https://travis-ci.com/github/rajat2004/AirSim/builds/180634888
Not sure why Travis isn't appearing here though

I didn't dig into this at the time since got busy with other things, will try to look at them more now

rajat2004:
Yeah, so the problem is still [there](https://travis-ci.com/github/rajat2004/AirSim/jobs/394020940#L6152), seems like there hasn't been a new release of `catkin_tools` with the `trollius` fix, should it be installed from the git repo instead?

saihv:
The latest travis build is still pulling catkin_tools 0.6.1, the async fix is only present in 0.7.0+.
image

 Could you try this?
`sudo apt-get install python3-osrf-pycommon python3-catkin-tools`

rajat2004:
@saihv I thought I had tried out installing from `apt` earlier, but didn't work. Just tried it out again - https://travis-ci.com/github/rajat2004/AirSim/builds/187934390
Works with 20.04, but breaks 18.04, 16.04. I'll try a bit again to get things working by shifting apt<->pip in places, but I think the easiest way might be to just run the appropriate command for 20.04 separately

saihv:
This looks good. I guess we'll keep an eye out and unify the scripts once the fixed catkin_tools is released. Thanks for looking into this!

rajat2004:
Great, thanks for merging!
Documentation can be updated after some testing has been done on Ubuntu 20.04
Could you perhaps have a look at why Travis isn't appearing in the CI checks?

PR #2956, opened 2020-08-20
Added angle wrapping when calculating angular velocity

cthoma20-arc:
This pull requests adds angle wrapping when calculating angular velocity. Previously, the angular velocity was calculated by converting the current rotation into euler angles, subtracting the previous rotation euler angles, and then calculating the angular velocity from this difference. The problem with this approach is that if the previous angle is close to 180 degrees, the current angle could wrap around back to -180 degrees, so the difference is very large instead of small. To fix this, the difference is normalized after subtraction, so the smallest angle is used.

I created this pull request because this problem was occurring when using AirSim with ArduRover.

### DIFF
--- a/AirLib/include/common/VectorMath.hpp
+++ b/AirLib/include/common/VectorMath.hpp
@@ -4,8 +4,10 @@
 #ifndef air_VectorMath_hpp
 #define air_VectorMath_hpp
 
+#include <cmath>
 #include "common/common_utils/Utils.hpp"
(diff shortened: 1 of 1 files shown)


zimmy87 (approved):
Tested this locally and it works for me.

zimmy87:
hey @cthoma20-arc is there an existing issue that's tracking the original bug being fixed here?

PR #2963, opened 2020-08-23
Fix simPlotPoints causing crash randomly

### LINKED ISSUE #2959: Bug: Using python's simPlotPoints (not too) many times sometimes crashes the Unreal Engine
When running the below line, it works ok.
self.client.simPlotPoints(points=[airsim.Vector3r(x_val=float(0), y_val=float(0), z_val=float(0))],
                                          color_rgba=[1.0, 0.0, 0.0, 1.0], size=1, duration=-1.0, is_persistent=True)
But if I run it multiple times, -sometimes- an Exception is thrown: "UE4Editor-Win64-DebugGame.exe has triggered a breakpoint."
in Unreal Projects\MyProject\Plugins\AirSim\Source\WorldSimApi.cpp line 321
After: 2639, 2846, 6117, and so on calls....
Notice: I run a constant, non-random code before this, always the same, but for some weird reason, it crashes, and also notice that the parameters I'm sending are very simple, and are almost the same as the basic usage example of simPlotPoints.
Annotation 2020-08-22 095752

rajat2004:
Running DrawDebugPoint on GameThread seems to fix the problem

Script - 
```
import airsim

client = airsim.VehicleClient()

points = 10000
for i in range(points):
    client.simPlotPoints(points=[airsim.Vector3r(x_val=float(0), y_val=float(0), z_val=float(0))],
                        color_rgba=[1.0, 0.0, 0.0, 1.0], size=10, duration=-1, is_persistent=True)
```

Without running on game thread specifically, it runs generally for 100 or even 1k calls, but always crashes when using 10k calls.
If this does fix the problem, then will probably need to update all the plot APIs

Thanks to @TomAvrech for reporting the issue and testing!
Fixes #2959

TODO: 
- [x] Make similar changes for all APIs
- [ ] Add some example scripts for plotting APIs (Probably as a future PR)

A question arose as to why something like this isn't required for DrawDebugPoints in Lidar, Distance Sensor (Maybe cause `Tick` runs in game thread)

### DIFF
--- a/Unreal/Plugins/AirSim/Source/WorldSimApi.cpp
+++ b/Unreal/Plugins/AirSim/Source/WorldSimApi.cpp
@@ -310,80 +310,112 @@ std::unique_ptr<std::vector<std::string>> WorldSimApi::swapTextures(const std::s
 //----------- Plotting APIs ----------/
 void WorldSimApi::simFlushPersistentMarkers()
 {
-    FlushPersistentDebugLines(simmode_->GetWorld());
+    UAirBlueprintLib::RunCommandOnGameThread([this]() {
+        FlushPersistentDebugLines(simmode_->GetWorld());
+    }, true);
 }
 
 void WorldSimApi::simPlotPoints(const std::vector<Vector3r>& points, const std::vector<float>& color_rgba, float size, float duration, bool is_persistent)
 {
-    FLinearColor color {color_rgba[0], color_rgba[1], color_rgba[2], color_rgba[3]};
-    for (const auto& point : points)
-    {
-        DrawDebugPoint(simmode_->GetWorld(), simmode_->getGlobalNedTransform().fromGlobalNed(point), size, color.ToFColor(true), is_persistent, duration);
-    }
+    FColor color = FLinearColor{color_rgba[0], color_rgba[1], color_rgba[2], color_rgba[3]}.ToFColor(true);
+
+    UAirBlueprintLib::RunCommandOnGameThread([this, &points, &color, size, duration, is_persistent]() {
+        for (const auto& point : points) {
+            DrawDebugPoint(simmode_->GetWorld(),
+                    simmode_->getGlobalNedTransform().fromGlobalNed(point),
+                    size, color, is_persistent, duration);
+        }
+    }, true);
 }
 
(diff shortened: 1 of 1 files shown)


madratman:
Good catch, I think this should fix it. 
This is the correct reasoning for lidar drawdebugpoints to not crash - (Maybe cause Tick runs in game thread). Lidar's drawdebugpoint is called in every tick in simmodeworldbase.

rajat2004:
Great, thanks for the review @madratman 
If you want, this PR can be merged as is I think, did want to add some examples and documentation but haven't started on them right now, probably as another PR

rajat2004:
I've added some more cleanups like creating the `color` object only once rather than every time, passing simple data-types like `bool`, `float` by value, style fixes

Unreal/Plugins/AirSim/Source/WorldSimApi.cpp
```diff
@@ -310,80 +310,112 @@ std::unique_ptr<std::vector<std::string>> WorldSimApi::swapTextures(const std::s
...
-    FLinearColor color {color_rgba[0], color_rgba[1], color_rgba[2], color_rgba[3]};
-    for (int idx = 0; idx < points.size(); idx += 2)
-    {
-        DrawDebugLine(simmode_->GetWorld(), simmode_->getGlobalNedTransform().fromGlobalNed(points[idx]), simmode_->getGlobalNedTransform().fromGlobalNed(points[idx+1]), color.ToFColor(true), is_persistent, duration, 0, thickness);
-    }
+    UAirBlueprintLib::RunCommandOnGameThread([this, &points, &color, thickness, duration, is_persistent]() {
+        for (int idx = 0; idx < points.size()-1; idx += 2) {
+            DrawDebugLine(simmode_->GetWorld(),
+                simmode_->getGlobalNedTransform().fromGlobalNed(points[idx]),
+                simmode_->getGlobalNedTransform().fromGlobalNed(points[idx+1]),
```
rajat2004, line 355: There is a slight behavioural change here, index limit is decreased since earlier, it could crash when odd number of points are given
This fixes that, the last point will be ignored if odd no. of points are passed

rajat2004:
Thanks for merging! Will try to add some scripts and docs as a future PR

PR #2975, opened 2020-08-26
Don't delete files when using update_to_git.sh

rajat2004:
`update_to_git.sh` currently deletes all the files in `Airlib/src` since they are removed in [`build.sh`](https://github.com/microsoft/AirSim/blob/master/build.sh#L114)
This does have a caveat where it won't delete the files deliberately deleted like in `include`, need to use a better way

### DIFF
--- a/Unreal/Environments/Blocks/update_to_git.sh
+++ b/Unreal/Environments/Blocks/update_to_git.sh
@@ -8,8 +8,8 @@ pushd "$SCRIPT_DIR" >/dev/null
 set -e
 set -x
 
-rsync -a  --exclude 'temp' --delete Plugins/AirSim ../../Plugins/
-rsync -a  --exclude 'temp' --delete Plugins/AirSim/Source/AirLib ../../../
+rsync -a --exclude 'temp' --delete Plugins/AirSim ../../Plugins/
+rsync -a --exclude 'temp' Plugins/AirSim/Source/AirLib ../../../
 
 popd >/dev/null
 set +x


madratman:
Hey, perhaps I'm missing something here. Can you share how are you using this script?
The following doesn't seem to lead in a loss of any files (wrt current master)
- setup.sh
- build.sh
- Unreal/Environments/Blocks/update_from_git.sh
- Unreal/Environments/Blocks/update_to_git.sh

I don't understand what you mean by "build.sh removing AirLib/src"?

rajat2004:
Then probably I'm not using it correctly then, is it required to use it with `update_from_git.sh`? `update_to_git.sh` copies the `AirLib/src` folder as well, so it's not deleted. [L114 in `build.sh`](https://github.com/microsoft/AirSim/blob/master/build.sh#L114) removes the `src` folder after copying to the `Plugins` directory since it's not needed
Without using that - 
```
$ ./update_to_git.sh      
+ rsync -a --exclude temp --delete Plugins/AirSim ../../Plugins/
+ rsync -a --exclude temp --delete Plugins/AirSim/Source/AirLib ../../../
+ popd
+ set +x

rajat at rajat-G5-5587 in ~/AirSim/Unreal/Environments/Blocks (test●●)
$ gws
 D ../../../AirLib/src/api/RpcLibClientBase.cpp
 D ../../../AirLib/src/api/RpcLibServerBase.cpp
 D ../../../AirLib/src/common/common_utils/FileSystem.cpp
 D ../../../AirLib/src/safety/ObstacleMap.cpp
 D ../../../AirLib/src/safety/SafetyEval.cpp
 D ../../../AirLib/src/vehicles/car/api/CarRpcLibClient.cpp
 D ../../../AirLib/src/vehicles/car/api/CarRpcLibServer.cpp
 D ../../../AirLib/src/vehicles/multirotor/api/MultirotorApiBase.cpp
 D ../../../AirLib/src/vehicles/multirotor/api/MultirotorRpcLibClient.cpp
 D ../../../AirLib/src/vehicles/multirotor/api/MultirotorRpcLibServer.cpp
```

madratman:
hmm, in general, `update_to_git.sh` is not needed to be used anywhere in the normal dev workflow. 
I think the original authors added it with the intent being that this script does the opposite of what update_from_git did. 
The intent of update_from_git is to copy updated AirSim plugin with latest AirLib code to Blocks env.
Now, if the dev makes changes to AirLib directly inside this copied plugin code (Unreal/Env/Blocks/Plugins/AirSim/AirLib) while developing in Unreal, and then wants to commit them, they can use update_to_git to copy the changed plugin back to root/AirLib. 

Coming to build.sh, that `rm -rf` lines slightly problematic now. Looking at history, it comes from this commit - https://github.com/microsoft/AirSim/commit/639b2130288c02f0591ef5217ed869ba5b111421. cc @saihv 
I think this change was made to speed up compilation times (CppCompileWithRpc v/s HeaderOnlyWithRpc). 
I think, removing the `rm -rf` line from build.sh won't make a difference, as Unreal will use the precompiled lib and headers anyway. 

Moreover, I think whatever rm -rf accomplishes is moot anyway, coz root/AirLib/ is copied  to Blocks project in L13 of update_from_git.sh (`rsync -a  --exclude 'temp' --delete ../../../AirLib Plugins/AirSim/Source/`)

@saihv  / @rajat2004  can you review the above. 
If this makes sense, I think (1) this PR can be closed, and (2) rm -rf AirLib/src line can be removed from build.sh

rajat2004:
Yup, agree that `update_to_git.sh` & `update_from_git.sh` are not used in my workflow atleast. I was kinda iterating over packaging, and using `build.sh` caused the entire Engine to be rebuilt, I think since it deletes the Binaries folder also. So was looking for a way to make changes and not have to rebuild, and tried out `update_to_git.sh` in the way. That it deleted the .cpp files was quite surprising, so just opened the PR. Still haven't figured out the part of packaging, but that's separate.

Closing this and removing the `rm -rf` line seems okay to me, shouldn't cause any problems

PR #2996, opened 2020-09-01
Sim run console command

ironclownfish:
This allows the client to execute a command in Unreal's native console, via an API.

- Has access to the countless built-in commands such as "stat unit", "stat fps", "open [map]", adjust any config settings, etc. etc.
- Allows the user to create their own APIs much more easily, by adding a custom event to the level blueprint, and then calling the console command "ce MyEventName [args]". This will be very useful internally for bespoke APIs; they're faster to implement and won't clutter the AirSim interface.


rajat2004:
Just noticed that [ConsoleCommand](https://docs.unrealengine.com/en-US/API/Runtime/Engine/GameFramework/APlayerController/ConsoleCommand/index.html) returns a FString, which I guess contains the output of the command if any. Could we perhaps return that instead of just a boolean to indicate whether the command was run?
Still, a way would be needed to indicate whether the command was executed, maybe returning something like `Failed`?

ironclownfish:
ConsoleCommand always returns the empty string. The last line of the function in the engine source is actually `return TEXT("");`

rajat2004:
Oh, wasn't aware of that, thanks!
Nothing's mentioned in the docs, very strange actually, maybe they kept it that way for some backwards compatibility?

ironclownfish:
It is very strange that the documentation for the function doesn't explain what the return value is.

ironclownfish:
Accidentally closed and reopened.

madratman:
lgtm. It would be great to have a docstring, and an example script as well!

PR #3000, opened 2020-09-03
Remove NameMode when spawning actors

saihv:
Temporary fix to preserve backwards compatibility with UE vers < 4.25


rajat2004:
Thanks for this. How will this affect the API, I mean will it still work or is there some behavioural change?

PR #3023, opened 2020-09-16
Add distortion model and distortion params to ros wrapper

m1baldwin:
CameraInfo ros messages have a "distortion_model" field, along with a "D" vector. Currently, there does not seem to be a way to specify this in the capture settings. Also, this is not filled out when generating a camera Info message for a given camera in the airsim_ros_wrapper.cpp. 

This PR adds default distortion fields in the camera info message generated in airsim_ros_wrappers.cpp. It is assuming the camera is ideal and using a plumb_bob model with zero distortion. This allows the video from a camera with a "Scene" image type to be used by software which requires distortion params, like [aurco_detect](http://wiki.ros.org/aruco_detect)


m1baldwin:
Needs to use updated getDistortion() api from PR #3039 .

PR #3039, opened 2020-09-25
Programmable camera distortion

saihv:
Introduces programmable camera distortion through a distortion material and a material parameter collection hosting the Brown distortion coefficients (3 radial, 2 tangential). Coefficients can be changed dynamically through the API.

Currently only works for the scene (RGB) camera.


m1baldwin:
@saihv Will there be a separate PR to also allow the ROS wrapper for airsim to retrieve the distortion parameters of the camera and publish them in the 'D' field of the sensor_msgs/Image message that it publishes per-camera?

ironclownfish (approved):
This looks good to me! There are a couple unnecessary line deletes/adds here and there, but in my opinion we don't need to worry so much about  a small amount of such things.
Cool feature!

saihv:
@m1baldwin I just added a getDistortionParams() API to retrieve the current distortion params of a camera. Perhaps you can reference this function from the ROS side as part of your PR to populate the distortion related fields?

PythonClient/airsim/client.py
```diff
@@ -482,6 +482,45 @@ def simGetCameraInfo(self, camera_name, vehicle_name = ''):
...
+    def simGetDistortionParams(self, camera_name, vehicle_name = ''):
+        """
+        Set camera distortion parameters
+
+        Args:
+            camera_name (str): Name of the camera, for backwards compatibility, ID numbers such as 0,1,etc. can also be used
+            distortion_params (dict): Dictionary of distortion param names and corresponding values
+                                        {"K1": 0.0, "K2": 0.0, "K3": 0.0, "P1": 0.0, "P2": 0.0}
+            vehicle_name (str, optional): Vehicle which the camera is associated with
+        """
```
rajat2004, line 494: Description needs to be fixed here

saihv: Fixed.

Unreal/Plugins/AirSim/Source/SimMode/SimModeBase.cpp
```diff
@@ -201,7 +200,7 @@ void ASimModeBase::initializeTimeOfDay()
         sky_sphere_ = sky_spheres[0];
         static const FName sun_prop_name(TEXT("Directional light actor"));
         auto* p = sky_sphere_class_->FindPropertyByName(sun_prop_name);
-        UObjectProperty* sun_prop = Cast<UObjectProperty>(p);
+        FObjectProperty* sun_prop = Cast<UObjectProperty>(p);
```
rajat2004, line 203: This change breaks UE 4.24, which is okay IMO, just need to update docs to 4.25
It still produces warnings on 4.25, so it should be fixed completely like
 `FObjectProperty* sun_prop = CastField<FObjectProperty>(p)`

rajat2004:
Also, this would normally be caught by Travis, but needs some non-implemented methods in Unity/PawnSimApi so as to not break Unity compilation
https://travis-ci.com/github/rajat2004/AirSim/builds/187945054

m1baldwin:
> @m1baldwin I just added a getDistortionParams() API to retrieve the current distortion params of a camera. Perhaps you can reference this function from the ROS side as part of your PR to populate the distortion related fields?

@saihv sure, I think this is a better approach to use the new api rather than how I have it hard coded currently. I need to update the PR.

m1baldwin:
Is it fine to use “plumb bob” model as the distortion model string here?

Unreal/Plugins/AirSim/Source/SimMode/SimModeBase.cpp
saihv: Good catch! I will revert this, because officially we still wish to support older UE versions.

saihv:
@m1baldwin Yes, plumb bob is appropriate.

Unreal/Plugins/AirSim/Source/SimMode/SimModeBase.cpp
rajat2004: Sure, that's also good, however I think the latest PRs like the segmentation one and even this has asset modifications which were done on 4.25 I assume, and wouldn't work in 4.24. Need to test all this though
While trying to maintain compatibility with 4.24, it might be good to update the docs to recommend 4.25, since some features could stop working on 4.24 and also raise support questions. Currently [docs](https://microsoft.github.io/AirSim/build_windows/) say 4.22 is supported, which I'm pretty sure is not the case

m1baldwin:
@saihv What is the difference between the simGetDistortion and getDistortion methods you have added?  And, would the ros  wrapper need to select between the two?

saihv:
@m1baldwin Can you doublecheck if you're using the right Python client? (i.e., the one in this branch and not another pre-installed version)

rajat2004:
@m1baldwin Which version of UE are you using? I think you'll have to use the version used by @saihv (probably 4.25.3) or later, since earlier versions will most likely fail to load the assets
As for the Python API, you'll have to use the files in this branch, You could create a file in the `multirotor/`, etc folders and use `import setup_path`. Another way which I use is to install airsim using `pip install -e .` from `Pythonclient` folder, that'll use the local package. Useful if changing branches, etc. since no need to use `setup_path` for quick tests

m1baldwin:
@rajat2004 I had upgraded to UE4.25.3 for the purpose of testing this PR with the ros wrapper. Everything seems to run and build fine. For the python client, might be that I am not using the changes in the local, so I will check that. That will be the first thing I check. I merged this PR into my PR I had been testing to bring them to the same place, I had made minimal changes so what I am testing now seems to be mostly this PR. 

Assuming this isn't an issue though for the cpp codes, I wonder why I am getting a seg fault? 

Also, I have noticed that if I remove the simGetDistortion API from the wrapper, and then run the wrapper as is, seemingly randomly the simulation is breaking. In VScode, it is throwing a debug breakpoint at line 494 in the enableCapture function in PIPCamera (https://github.com/saihv/AirSim/blob/PR/distortion/Unreal/Plugins/AirSim/Source/PIPCamera.cpp#L494). This breakout only happens when I run something which is continuously calling the simGetImages API (ros_wrapper in this case)

rajat2004:
@m1baldwin Could you perhaps try out #3064? It might be causing the problem, it reverts the PR #2881, I'm not sure why this didn't appear when I tested with 10-15 min runs on my system, and with other testers as well. But seems to be same as [here](https://github.com/microsoft/AirSim/pull/2713#discussion_r485268601), so reverted the PR

m1baldwin:
@rajat2004 sure, I could test with that next. Does it work with ue4.25.3?

rajat2004:
Yeah, it only has C++ code change, no asset modifications so works with any UE version

m1baldwin:
@rajat2004 okay. It is weird, because I thought I had been using the latest master anyway before this, and I wasn't seeing this failure either. It was only when I tried this PR that the issue popped up.

m1baldwin:
@rajat2004 Just to confirm, PR #3064 does fix the issue for me. The weird thing is I went back and double checked what I had been running previously. All last week, I was running the latest master which included your original PR to improve the image capture rate. For some reason after testing this PR, I started seeing the failure I described above.

I tried resetting everything to what I had last week, but now I see the failure still. Testing the PR you recommended works, but now of course my frame rate is low again. I am going to need to spend some more time testing as I will need the higher fps for my work.

tool3210 (approved):
It looks good.

jonyMarino:
Hi @saihv. We are getting this from Unity build:
```
C:\Users\USER\source\repos\AirSim\Unity\AirLibWrapper\AirsimWrapper\Source\PawnSimApi.cpp(231): error C4716: 'PawnSimApi::getDistortionParams': must return a value [C:\Users\USER\source\repos\AirSim\Unity\AirLi
bWrapper\AirsimWrapper\AirsimWrapper.vcxproj]
```

saihv:
@jonyMarino Fixed

jonyMarino:
Thank you, Sai! Great contribution! I had fun testing it.

PR #3053, opened 2020-10-01
Segmentation palette modifications

saihv:
This PR: 

1. Explicitly adds black (0, 0, 0) and white (255, 255, 255) as part of the segmentation palette - to make masks easier to specify and process when there is a single object class of interest. These RGBs map to indices 0 and 255 in the palette, remaining RGBs are unchanged from their previous versions.
2. Fixes the naming of seg palette.


ykhedar:
While using the AirSim plugin from the master branch, the segmentation does not work with UE4.24.3. The following warning is thrown in the Output Logs of the UE:

`LogLinker: Warning: Asset 'C:/UUsers/user/Desktop/UE_Project/Plugins/AirSim/Content/Blueprints/BP_PIPCamera.uasset' has been saved with empty engine version. The asset will be loaded but may be incompatible.`

`LogLinker: Warning: Asset 'C:/Users/user/Desktop/UE_Project/Plugins/AirSim/Content/HUDAssets/SegmentationMaterial.uasset' has been saved with engine version newer than current and therefore can't be loaded. CurrEngineVersion: 4.24.3-11590370+++UE4+Release-4.24 AssetEngineVersion: 4.25.0-13144385+++UE4+Release-4.25`

Probably the assets were developed under UE 4.25.3 and pushed to the master branch?

PR #3064, opened 2020-10-07
Revert "Disable cameras after fetching images, projection matrix (#2881)"

rajat2004:
Running the Image API continuously seems to trigger exceptions

This reverts commit ba4be152b5f06279261150aada9af510e6e1b431.

### DIFF
--- a/Unreal/Plugins/AirSim/Source/PIPCamera.cpp
+++ b/Unreal/Plugins/AirSim/Source/PIPCamera.cpp
@@ -170,9 +170,6 @@ msr::airlib::ProjectionMatrix APIPCamera::getProjectionMatrix(const APIPCamera::
     else
         mat.setTo(Utils::nan<float>());
 
-    // Disable camera after our work is done
-    const_cast<APIPCamera*>(this)->setCameraTypeEnabled(image_type, false);
-
     return mat;
 }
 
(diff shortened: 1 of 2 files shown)


saihv:
Any thoughts on why this might be causing image capture issues? Could it be because the image API is accessing the camera again just as it's being disabled? (Most Unreal operations might be bound to the scene tick rate)

rajat2004:
Not sure, most likely along what you're thinking, it could be marking it as "Disabled", but some stuff happens later on in the next tick and sometimes things don't work out well. [Comment here](https://github.com/microsoft/AirSim/blob/master/Unreal/Plugins/AirSim/Source/PIPCamera.cpp#L457) indicates that this does cause crashes
I'm not well-versed enough in UE to say anything for sure though

rajat2004:
I've changed it from a draft PR since it has been tested to fix the issue by at least 1 user, though more testing would be good
Should we instead add a `simDisableCamera` API as was done in the original PR https://github.com/microsoft/AirSim/pull/2465?

m1baldwin:
Just wondering, is the net effect of the original PR improving the frame rate to toggle the camera device off in between image requests? I suppose this would mean the camera is being activated/deactivated too rapidly, which the comment @rajat2004 pointed out in line 457 might be causing unreal to crash due to unnecessary calls to Activate()?

saihv:
I was able to repro both the issue and the fix from this PR. Merging this as an immediate fix, we can continue discussion about how best to re-implement the disable feature.

rajat2004:
Great, thanks for testing and merging!
@m1baldwin Yeah, not exactly sure of the reason, but seems like enabling and disabling cameras isn't a completely synchronous task, and activating the camera while not completely deactivated could be causing the crash. This was more apparent in #3018 due to the higher FPS

PR #3067, opened 2020-10-07
[Docs] Added install instructions for Ubuntu 20.04

ash-xyz:
Added install instructions for Ubuntu 20.04 using @rajat2004's #2953 for reference.

I should specify that a work around for [catkin tool's](https://github.com/catkin/catkin_tools) is mentioned in the docs until it reaches 0.7.0


jonyMarino:
Thanks!

PR #3100, opened 2020-10-27
[RFC] linux: llvm: specify llvm lib dir

maquefel:
Request for comments.

In case if we building with llvm other then system - pay respects to
LLVM_DIR to specify libfs++ location.

We can have other version (not suitable) of llvm as our system package.


zimmy87:
Does this PR add a dependency on the llvm package? As best I can tell, that seems to be what the [find_package](https://cmake.org/cmake/help/latest/command/find_package.html) command is doing. Unless there is some functionality that requires this dependency, I think it would be best to make this optional if we want to merge this into master.

PR #3102, opened 2020-10-27
Added continueForFrames

niekvleeuwen:
This pull request makes it possible to run the simulation for a specified number of frames, similar to `simContinueForTime`.

### DIFF
--- a/PythonClient/airsim/client.py
+++ b/PythonClient/airsim/client.py
@@ -110,6 +110,15 @@ class VehicleClient:
             seconds (float): Time to run the simulation for
         """
         self.client.call('simContinueForTime', seconds)
+    
+    def simContinueForFrames(self, frames):
+        """
+        Continue the simulation for the specified number of frames
+
+        Args:
+            frames (float): Frames to run the simulation for
+        """
+        self.client.call('simContinueForFrames', frames)
 
     def getHomeGeoPoint(self, vehicle_name = ''):
         """
--- a/PythonClient/airsim/client.py
+++ b/PythonClient/airsim/client.py
@@ -113,7 +113,7 @@ class VehicleClient:
     
     def simContinueForFrames(self, frames):
         """
-        Continue the simulation for the specified number of frames
+        Continue (or resume if paused) the simulation for the specified number of frames, after which the simulation will be paused.
 
         Args:
             frames (float): Frames to run the simulation for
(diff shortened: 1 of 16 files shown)


Unreal/Plugins/AirSim/Source/SimMode/SimModeWorldBase.cpp
```diff
@@ -100,6 +100,23 @@ void ASimModeWorldBase::continueForTime(double seconds)
     UGameplayStatics::SetGamePaused(this->GetWorld(), true);
 }
 
+void ASimModeWorldBase::continueForFrames(uint32_t frames)
+{
+    if(physics_world_->isPaused())
```
jonyMarino, line 105: What you want to test is the opposite. If it is **not** paused, then...

PythonClient/airsim/client.py
```diff
@@ -110,6 +110,15 @@ def simContinueForTime(self, seconds):
             seconds (float): Time to run the simulation for
         """
         self.client.call('simContinueForTime', seconds)
+    
+    def simContinueForFrames(self, frames):
+        """
+        Continue the simulation for the specified number of frames
```
jonyMarino, line 116: Can you add the expected behavior according to the state (paused or not) of the simulation?

jonyMarino (commented):
Thanks, @niekvleeuwen, for contributing to AirSim! Look at the review, please!

Unreal/Plugins/AirSim/Source/SimMode/SimModeWorldBase.cpp
niekvleeuwen: If `simContinueForFrames` is called, I check if the simulation is paused. If this is the case the simulation must be resumed so that `continueForFrames` can stop the simulation after the number of desired frames. 
When the simulation is not paused when `simContinueForFrames` is called, the simulation should **not** be paused until the desired number of frames has passed.

PythonClient/airsim/client.py
niekvleeuwen: I added an explanation about the expected behavior according to the state of the simulation in d2a2d0632d42bd47f5de3e38c7fb122340cc6978.

niekvleeuwen: Is the added explanation sufficient or are there still improvements that can be made?

zimmy87: this looks to be addressed in the latest version of this PR, so I'm resolving this conversation

zimmy87:
Thank you for contributing this PR @niekvleeuwen, I tested this locally and the new API appears to work as intended so I went ahead and merged this PR.

PR #3120, opened 2020-11-02
fix error type of first parameter of simGetImage

kemen209:
The first parameter of simGetImage should be of type string not integer.


zimmy87:
Thank you @kemen209 for the contribution!

PR #3125, opened 2020-11-04
set 'if sentences' with right condition

kemen209:
(1) The following condition check code in MavLinkMultirotorApi.hpp :
```
if (mavcon != connection_) {
```
are date back to commit a9f444367bb7fcfac44231068b93a0bc7aa2f7de(are valid at that time), but now become obsolete(the condition will always been true, it compared two pointers from two sources). It should be removed.

(2) On the other hand, mav_vehicle_->getConnection() will return nullptr if [connection_info_.control_ip_address == "" ]

And the fix cover these two issues here.

### DIFF
--- a/AirLib/include/vehicles/multirotor/firmwares/mavlink/MavLinkMultirotorApi.hpp
+++ b/AirLib/include/vehicles/multirotor/firmwares/mavlink/MavLinkMultirotorApi.hpp
@@ -427,7 +427,7 @@ public: //methods
 
         // listen to the other mavlink connection also
         auto mavcon = mav_vehicle_->getConnection();
-        if (mavcon != connection_) {
+        if (mavcon != nullptr) {
             mavlinkcom::MavLinkTelemetry gcs;
             mavcon->getTelemetry(gcs);
 
@@ -878,7 +878,10 @@ private: //methods
                 logviewer_out_proxy_ = nullptr;
             }
             else if (mav_vehicle_ != nullptr) {
-                mav_vehicle_->getConnection()->startLoggingSendMessage(std::make_shared<MavLinkLogViewerLog>(logviewer_out_proxy_));
+                auto mavcon = mav_vehicle_->getConnection();
+                if (mavcon != nullptr) {
+                    mavcon->startLoggingSendMessage(std::make_shared<MavLinkLogViewerLog>(logviewer_out_proxy_));
+                }
             }
         }
         return logviewer_proxy_ != nullptr;
@@ -923,7 +926,7 @@ private: //methods
         connection_->join(connection);
 
         auto mavcon = mav_vehicle_->getConnection();
-        if (mavcon != connection_) {
+        if (mavcon != nullptr) {
             mavcon->join(connection);
         }
     }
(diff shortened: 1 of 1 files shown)


AirLib/include/vehicles/multirotor/firmwares/mavlink/MavLinkMultirotorApi.hpp
```diff
@@ -923,7 +926,7 @@ class MavLinkMultirotorApi : public MultirotorApiBase
         connection_->join(connection);
 
         auto mavcon = mav_vehicle_->getConnection();
-        if (mavcon != connection_) {
```
zimmy87, line 926: Has this been tested with trying to connect to PX4 over a serial connection? On https://github.com/microsoft/AirSim/blob/master/AirLib/include/vehicles/multirotor/firmwares/mavlink/MavLinkMultirotorApi.hpp#L1126 there's a call to `mav_vehicle_->connect(connection_);` If this call happens before the check here in `createProxy()` you could inadvertently end up calling join() on the same connection twice.

zimmy87 (changes requested):
I'd like to know if this PR has been tested using a serial connection to make sure it is safe to remove the `mavcon != connection_` check

AirLib/include/vehicles/multirotor/firmwares/mavlink/MavLinkMultirotorApi.hpp
```diff
@@ -1064,15 +1067,12 @@ class MavLinkMultirotorApi : public MultirotorApiBase
...
+        if (mavcon != nullptr) {
             mavcon->subscribe([=](std::shared_ptr<mavlinkcom::MavLinkConnection> connection, const mavlinkcom::MavLinkMessage& msg) {
                 unused(connection);
                 processMavMessages(msg);
                 });
         }
-        else {
-            mav_vehicle_->connect(connection_);
-        }
```
zimmy87, line 1075: Why was this else block removed?

AirLib/include/vehicles/multirotor/firmwares/mavlink/MavLinkMultirotorApi.hpp
kemen209: createProxy() will only be called from connectToLogViewer() and connectToQGC(), and these two will be called after createMavConnection(). So i think this will not cause any problem.

```
        createMavConnection(connection_info_);
        if (mav_vehicle_ != nullptr) {
            connectToLogViewer();
            connectToQGC();
        }
```

kemen209: createProxy() will only be called from connectToLogViewer() and connectToQGC(), and these two will be called after createMavConnection(). So i think this will not cause any problem.

```
        createMavConnection(connection_info_);
        if (mav_vehicle_ != nullptr) {
            connectToLogViewer();
            connectToQGC();
        }
```

AirLib/include/vehicles/multirotor/firmwares/mavlink/MavLinkMultirotorApi.hpp
kemen209: After some re-thinking, i think i made a mistake on this one(treat mav_vehicle_ as mavcon). This else block should not been removed.

kemen209:
@zimmy87 After done some rechecking and re-thinking, i think i made a mistake on the assumption that "mavcon" and "connection_" are the same object. In fact, there are NOT!  I shall drop this merge request. Sorry for wasting your time.

PR #3133, opened 2020-11-10
Clarified multifile pre-compiled binary download instructions

luisquintanilla:
Some environment precompiled binaries span multiple files. It can be confusing as to whether they are different environments or environments spanning multiple files. I added some instructions to clarify this point and make sure to inform users to download both binaries before starting an environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,8 @@ For complete list of changes, view our [Changelog](docs/CHANGELOG.md)
 
 ## How to Get It
 
+[![Build Status](https://travis-ci.org/Microsoft/AirSim.svg?branch=master)](https://travis-ci.org/Microsoft/AirSim)
+
 ### Windows
 * [Download binaries](https://github.com/Microsoft/AirSim/releases)
 * [Build it](https://microsoft.github.io/AirSim/build_windows)
@@ -44,7 +46,7 @@ For complete list of changes, view our [Changelog](docs/CHANGELOG.md)
 ### macOS
 * [Build it](https://microsoft.github.io/AirSim/build_linux)
 
-[![Build Status](https://travis-ci.org/Microsoft/AirSim.svg?branch=master)](https://travis-ci.org/Microsoft/AirSim)
+Some pre-compiled environment binaries may include multiple files (i.e. City.zip.001, City.zip.002). Make sure to download both files before starting the environment.
 
 ## How to Use It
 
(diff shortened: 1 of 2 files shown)


jonyMarino:
@luisquintanilla, Thanks! Want to add download and concatenate, please?

luisquintanilla:
Sure @jonyMarino. To clarify do you mean instead of *Make sure to **download** both files before starting the environment.*, it should say *Make sure to **download and concatenate** both files before starting the environment.*? If so, I don't recall concatenating anything. Unzipping and running one of the files automatically ran the environment successfully. Though both subdirectories were both in the same root directory, so I could see this potentially failing if they're in separate directories.

jonyMarino:
That's rear. Maybe you are using a smart unzipper? Can you try again, please?.

jonyMarino:
I think you are using 7zip. So it depends which "unzipper" you are using. So, can you clarify, please?

luisquintanilla:
@jonyMarino that is correct. I used 7zip.

README.md
```diff
@@ -44,7 +46,7 @@ For complete list of changes, view our [Changelog](docs/CHANGELOG.md)
 ### macOS
 * [Build it](https://microsoft.github.io/AirSim/build_linux)
 
-[![Build Status](https://travis-ci.org/Microsoft/AirSim.svg?branch=master)](https://travis-ci.org/Microsoft/AirSim)
+Some pre-compiled environment binaries may include multiple files (i.e. City.zip.001, City.zip.002). Make sure to download both files before starting the environment.
```
jonyMarino, line 49: I don't think this is a place to talk about that.

docs/use_precompiled.md
```diff
@@ -4,7 +4,7 @@ You can simply download precompiled binaries and run to get started immediately.
 
 ### Unreal Engine
 
-**Windows, Linux**: Download the binaries for the environment of your choice from the [latest release](https://github.com/Microsoft/AirSim/releases).
+**Windows, Linux**: Download the binaries for the environment of your choice from the [latest release](https://github.com/Microsoft/AirSim/releases). Some pre-compiled environment binaries may include multiple files (i.e. City.zip.001, City.zip.002). Make sure to download both files before starting the environment.
```
jonyMarino, line 7: Add: "You may need to concatenate both files to unzip them."

jonyMarino (changes requested):

docs/use_precompiled.md
luisquintanilla: Done

jonyMarino:
Great! Thanks!

jonyMarino (approved):

PR #3134, opened 2020-11-11
fix issue of unit conversion

kemen209:
meter to centimeter should multiple by 100


rajat2004 (approved):
Thanks for catching this! Here's the Mavlink definition - https://mavlink.io/en/messages/common.html#DISTANCE_SENSOR

rajat2004 (commented):
Just an additional comment which came to my mind, it might be good to also change the argument names in the [method definition](https://github.com/microsoft/AirSim/blob/master/AirLib/include/vehicles/multirotor/firmwares/mavlink/MavLinkMultirotorApi.hpp#L1398) to have `_cm` at the end to make it clear to the reader what is expected

jonyMarino:
Thanks for your contributions, @kemen209! keep it up!

PR #3146, opened 2020-11-17
Fix C4244 warnings in build.cmd and issue with manually placing pawns

zimmy87:
This PR contains two bug fixes. The first is for C4244 warnings that get treated as errors, thus breaking the build.cmd script. The other is an issue where removing a PlayerStart and manually placing a build can cause a crash and also incorrectly places the camera (previously it was defaulting the camera to the position of the camera in the Editor, instead of placing the camera next to the pawn)


AirLib/include/api/VehicleSimApiBase.hpp
```diff
@@ -68,6 +68,8 @@ class VehicleSimApiBase : public msr::airlib::UpdatableObject {
     //use pointer here because of derived classes for VehicleSetting
     const AirSimSettings::VehicleSetting* getVehicleSetting() const
     {
+        if (getVehicleName().find("BP_FlyingPawn") == 0)
```
jonyMarino, line 71: We should change this for any vehicle name

jonyMarino:
@zimmy87, please resolve the current conflict with master. Good luck!

rajat2004:
This issue with warnings could have been avoided if the CI was active I think
I guess with the recent changes with Travis it hasn't been working, no idea about Azure Pipelines though, that's in a private organization or something if I'm not wrong, so @jonyMarino if possible, could you have a look at it?
I'm also thinking of setting up a GitHub Actions pipeline for AirSim, haven't actually used one before so will be fun to learn and try it out. Initially the pipeline could just be for doing Linux AirLib build, then maybe add Windows, OSX & ROS wrapper also

jonyMarino:
Hi @rajat2004! Yes, you are totally right. We want to improve our CI tools. Why would you prefer GitHub Actions instead of Travis and Azure Pipelines?

rajat2004:
It's not that I prefer one over the other, Travis & Azure are already present so it would be pretty easy to have them up and running again. Azure specifically would be great since it does the entire build with UE included
Regarding Travis, there was a [recent pricing announcement](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing), and the maintainers would have to decide whether to use Travis or something else. GitHub Actions seems a pretty good alternative to me, ArduPilot also recently added GitHub Actions for doing CI

rajat2004:
I've opened https://github.com/microsoft/AirSim/pull/3180 which adds Ubuntu & Windows builds, and it did pick up the compilation error on Windows. The checks aren't being run on the PR, most likely needs some enabling by the maintainers, or maybe will work after getting merged. See the PR on my fork (https://github.com/rajat2004/AirSim/pull/26) or the branch itself which will show the checks being executed

Issue #2575, opened 2020-04-14
PX4 HIL OriginGeopoint Issue (Airsim 1.2.2)


731avenger:
I am attempting to run what seems like a very simple scenario. I need define Airsim and the included PX4 HIL sim with a real origin/home position. After reading through the documentation their is two approaches one can take:
One is to use the params setting which define the home position in the SIL sim. https://microsoft.github.io/AirSim/px4_sitl/#setting-gps-origin
```
    "Vehicles": {
        "Drone1": {
            "VehicleType": "PX4Multirotor",
            "VehicleSysID": 3,
            "SerialPort": "COM16",
            "UseSerial": true,
            "params": {
               "LPE_LAT": 33.33313889,
               "LPE_LON": -84.2666667
                }
        }
    }
```
However, this approach seems to lead to very mixed results with the PX4 origin being set all over the globe rather inconstantly. In the SIL sim, this method works fine but as soon as you introduced the HIL sim, this method seems to no longer work on PX4 1.10.1 running on the Mini Pixhawk 4 (FMU5).

The second approach is to use the OrginGeopoint option. 
```
	"OriginGeopoint": {
        "Latitude": 33.33313889,
        "Longitude": -84.2666667,
        "Altitude": 20
    },
    "Vehicles": {
        "Drone1": {
            "VehicleType": "PX4Multirotor",
            "VehicleSysID": 3,
            "SerialPort": "COM16",
            "UseSerial": true,
			"Z": 180
        }
    }
```
This method at least got me in the right zip code. However, when I define "33.33313889, -84.2666667" I expect to see that in Qground Control, but QGC is showing a Lat of 33.3487423 and a long -84.240305 which is ~4Km to the northeast of my desired point. 

I am running the pre-built plains map on Airsim 1.2.2. I have attempted this with a number of pixhawk 4s and Mini 4s running PX4 1.10, 1.10.1, 1.9 and 1.11. I am running Airsim on Windows 10 64-bit with the latest  updates.

rajat2004:
The `params` section in `settings.json` was added after 1.2.2 release, and therefore doesn't have any effect on the Home position
As to the difference in the positions from `OriginGeoPoint`, I think the PlayerStart in the env is causing the change. The PlayerStart in UE also has its own offset, which is being added to the Origin set in `settings.json`. Don't think this can be changed since it's a binary, but an easy fix could be to observe the difference and then compensate for that in `settings.json` to get the desired coordinates

[...]

PR #3147, opened 2020-11-17
Fixed OriginGeopoint GPS error

### LINKED ISSUE #2588: Clarification on the OriginGeopoint
There seems to be conflicting data on what the "OriginGeopoint" is. In the documentation it's referenced as the location of the PlayerStart while in code it's commented as the coordinates of Unreal level at the coordinates 0,0,0.

Code:
https://github.com/microsoft/AirSim/blob/e24a11d5985c1d6f573735f28a3844bc634559db/AirLib/include/common/AirSimSettings.hpp#L350

Documentation:
https://microsoft.github.io/AirSim/settings/#origingeopoint

Which leads into my larger question of what would need to be modified in-code, or otherwise to correspond an arbitrary point in Unreal to a specified lat-lon-alt?

Thanks!

Edit: 
From what I've tested it seems to be the lat/lon/alt of the Unreal level at 0,0,0, but would still like some confirmation on this, in which case we should update the documentation for now :)

ahmed-elsaharti:
Fixes #2588, #2575
Currently, "OriginGeopoint" in settings.json is the Lat/Lon location of the world coordinate system origin of the Unreal level  (0,0,0) while it is described in the documentation as the Lat/Lon location of the PlayerStart/Vehicle.

This results in an offset in Lat/Lon between the OriginGeopoint specified and that of the vehicle (reported by the API or by QGC in HITL or SITL):
image
With PlayerStart at 
image, we get:
image

This PR fixes the issue by using the PlayerStart location to set a new origin for the world coordinate system. Doing so moves the world frame to the PlayerStart point without moving the Player or any of the objects in the scene. We then re-grab the 'new' player start position (which is now 0,0,0 in relation to the new world origin). 
This results in the OriginGeopoint falling right where the Player/Vehicle is (since the player is at the world origin now) while reporting the correct PlayerStart location in reference to the level origin.

After the fix, for the same OriginGeopoint and PlayerStart position as above we get:
image

(Thanks to @jonyMarino for his help with this)

### DIFF
--- a/Unreal/Plugins/AirSim/Source/SimMode/SimModeBase.cpp
+++ b/Unreal/Plugins/AirSim/Source/SimMode/SimModeBase.cpp
@@ -92,8 +92,15 @@ void ASimModeBase::BeginPlay()
     //get player start
     //this must be done from within actor otherwise we don't get player start
     APlayerController* player_controller = this->GetWorld()->GetFirstPlayerController();
+    // Grab player location
     FTransform player_start_transform = player_controller->GetViewTarget()->GetActorTransform();
-    global_ned_transform_.reset(new NedTransform(player_start_transform, 
+    FVector player_loc = player_start_transform.GetLocation();
+    // Move the world origin to the player's location (this moves the coordinate system and adds
+    // a corresponding offset to all positions to compensate for the shift)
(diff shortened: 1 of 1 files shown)


jonyMarino (approved):

jonyMarino:
Thank you, Ahmed! Keep going!

Issue #3010, opened 2020-09-08
ClockSpeed issues with PX4 SITL


clintlombard:
When changing the `ClockSpeed` parameter in `settings.json` to anything faster or slower than normal speed, flight becomes unstable with PX4 in SITL. I even tried setting PX4's `PX4_SIM_SPEED_FACTOR` to no avail.

This seems to be a long standing issue https://discuss.px4.io/t/px4-sitl-airsim-lockstep-and-timeouts/17380

Python: 3.6.9
AirSim Plugin: commit bd8ee6b 
AirSim Python API: 1.2.8
UE: 4.25.3
PX4: v1.10.1 and v1.10.2 (v1.11.0 doesn't yet work with AirSIm)
OS: Ubuntu 18.04 (docker container)

m1baldwin:
@clintlombard Just wondering, when you try running v1.11.0 of the px4 release, what errors do you get? I am getting accel timeout errors. When I try running with the v1.10.*, it works fine though.

clintlombard:
No errors on `v1.11.0` it just doesn't seem to connect. Might be the updates to mavlink api. In `v1.10.*` under the normal clock speed (1.0) I get the accel timeouts and things still work. However, if you decrease the clock speed to say 0.25, then the PX4 EKF GPS fusion just times out (so basically just dead reckoning).

m1baldwin:
Interesting, I can at least get it to connect with px4 version 1.11.0. 

When I have tried running at a speed factor of less than 1 with px4. I have also seen issues. For me, the issues I saw is that the drone was no longer able to hit waypoints accurately, i.e. it would usually completely overshoot them. I couldn't find a fix and haven't looked into why this might be happening.

Is decreasing the clock speed also decreasing the rate the sensor plugin in airsim for the gps is producing data? Could be that the px4 sitl stack is still running at full speed, however the airsim clock is reduced per the spec in the settings file. Maybe there is some signaling missing here to properly configure the px4 sitl to run at a matching clock rate?

Just a thought, there is an environmental variable that can be set to configure the px4 sim speed. Maybe this also needs to be set before running px4 sitl? It can be set with: 
export PX4_SIM_SPEED_FACTOR <R>
https://dev.px4.io/v1.10/en/simulation/gazebo.html#change-simulation-speed

clintlombard:
I'll have to give `v1.11.0` another try.

Well technically things should be running in lockstep so I would assume PX4 would just wait on AirSim.

[...]

PR #3158, opened 2020-11-20
unified the simulation time 

kemen209:
make it work under steppable-clock. 

Maybe related to #3010.

### DIFF
--- a/AirLib/include/vehicles/multirotor/firmwares/mavlink/ArduCopterSoloApi.hpp
+++ b/AirLib/include/vehicles/multirotor/firmwares/mavlink/ArduCopterSoloApi.hpp
@@ -28,7 +28,7 @@ public:
             const auto& imu_output = getImuData("");
 
             SensorMessage packet;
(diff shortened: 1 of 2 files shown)


zimmy87:
Hi @kemen209, how was this PR tested?

kemen209:
Hi @zimmy87, i try the following setting on my local machine:
```
setting.json

"ClockType": "SteppableClock",
"ClockSpeed": 1.0,
```

And i use some debugging trick to slow down the simu clock, and it worked as i expected: simu clock only move one step, but the wall clock may passed hundreds  of seconds.

Besides this, no other tests, like auto test involved.  

From my understanding, this bug only a problem when use 'SteppableClock' with slow or faster simu clock.

zimmy87:
Thank you for submitting this PR @kemen209! I tested this locally and it appears to work as intended, so I am merging this. This also fixed #3010 for me, so I will refer people to this fix in that issue.

PR #3163, opened 2020-11-21
pixels_as_float not working in Unity

ykeuter:
This seems to be a typo.


jonyMarino:
Hi @ykeuter! Thanks for catching this and contributing!

PR #3194, opened 2020-12-07
Update airsim ros pkgs docs

### LINKED ISSUE #3095: AirSim with ROS WSL2 Connection Error
I've been trying to connect my WSL2 ROS instance to AirSim on Windows. The settings.json I am using is `front_stereo_and_center_mono.json` from `airsim_tutorial_pkgs`. When I run the command `roslaunch airsim_ros_pkgs airsim_node.launch output:=screen host:=$WSL_HOST_IP` it gets connected, but starts outputting this error:

```
Exception raised by the API:
rpclib: function 'simGetGroundTruthEnvironment' (called with 1 arg(s)) threw an exception. The exception contained this information: Vehicle Sim-API for 'SimpleFlight' is not available. This could either because this is not a simulation or this vehicle does not exist.
```

Currently, I am using Ubuntu 20.04 in WSL2 and have confirmed that the ros packages work on another linux computer running only Ubuntu 20.04.

Has anyone ran into this issue and resolved it?

jonyMarino:
Thanks to the collaborators involved in the discussion in #3095!

### DIFF
--- a/docs/airsim_ros_pkgs.md
+++ b/docs/airsim_ros_pkgs.md
@@ -253,7 +253,7 @@ From within Windows, the WSL distribution's files are located at (type in window
    `\\wsl$\Ubuntu-18.04`
    
 #### How to run Airsim on Windows and ROS wrapper on WSL
-You need to have the same setting.json content for the simulator and WSL (e.g., in each Documents\Airsim folder), and due to a bug, the drone needs to have the name "SimpleFlight" in the settings.  
+You need to have the same setting.json content for the simulator and WSL (e.g., in each Documents\Airsim folder).
 For WSL 1 execute:  
 `export WSL_HOST_IP=127.0.0.1`  
 and for WSL 2:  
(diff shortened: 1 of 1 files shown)


docs/airsim_ros_pkgs.md
```diff
@@ -198,17 +199,71 @@ Throttle, brake, steering and gear selections for control. Both automatic and ma
...
+
+
+From within Windows, the WSL distribution's files are located at (type in windows Explorer address bar):
+
+   `\\wsl$\<distribution name>`
+   e.g.
+   `\\wsl$\Ubuntu-18.04`
+   
+#### How to run Airsim on Windows and ROS wrapper on WSL
+You need to have the same setting.json content for the simulator and WSL (e.g., in each Documents\Airsim folder), and due to a bug, the drone needs to have the name "SimpleFlight" in the settings.
```
rajat2004, line 256: Just to clarify, the name of the vehicle must be `SimpleFlight`, but the autopilot can be PX4 or ArduPilot as well? This is applicable for both cars and drones?
Pretty strange bug, I guess will be good to fix it now itself when it has been found if not too difficult

jonyMarino: Hi Rajat! Thanks for the review! I didn't try with autopilot, either with cars. I think it is a good idea to clarify that there in the docs. We don't have time right now to search for the cause of this. I'm going to ask in the community if someone is interested.

jonyMarino: I have tested and there is not such an issue. I'm going to change this and merge. Thanks!

PR #3196, opened 2020-12-08
Add ApiServerPort to available AirSim settings

wangwwno1:
[The original docs](https://airsim-fork.readthedocs.io/en/latest/settings.html#available-settings-and-their-defaults) omit the "ApiServerPort" option - an option enable the user to modify the default rpc port (41451) used by airsim.

This PR will fix it.

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -37,6 +37,7 @@ Below are complete list of settings available along with their default values. I
   "ClockType": "",
   "ClockSpeed": 1,
   "LocalHostIp": "127.0.0.1",
+  "ApiServerPort": 41451,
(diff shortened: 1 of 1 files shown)


rajat2004 (commented):
Thanks for this! Along with this, I feel adding a small description of the parameter would be very useful

wangwwno1:
> Thanks for this! Along with this, I feel adding a small description of the parameter would be very useful

Extra description added below the descriptiton of LocalHostIp Setting.

jonyMarino:
@wangwwno1 
Thanks for contributing!

PR #3202, opened 2020-12-10
Event camera simulation

saihv:
This PR introduces a simple event camera simulation in Python, using numba for performance. 

The event simulator uses two consecutive RGB images (converted to grayscale), and computes "events" based on the change in log luminance between the images. These events are reported as a stream of bytes, following this format:

`<x> <y> <timestamp> <pol>`

x and y are the pixel locations of the event firing, timestamp is the global timestamp in microseconds and pol is either +1/-1 depending on whether the brightness increased or decreased. Along with this bytestream, an accumulation of events over a 2D frame is also constructed, known as an 'event image' that visualizes +1 events as red and -1 as blue pixels. 

image

There are quite a few parameters that can be tuned to achieve a level of visual fidelity/performance. The main factors would be the resolution of the camera and the log luminance threshold (TOL) that determines whether or not a detected change counts as an event. There is also currently a max limit on the number of events generated per pair of images, which can also be tuned. 

- [X] Event simulator backend
- [X] Event image visualization
- [x] Return list of bytes for each new image obtained
- [x] Save events to a file if needed

### DIFF
--- a/PythonClient/eventcamera_sim/event_simulator.py
+++ b/PythonClient/eventcamera_sim/event_simulator.py
@@ -114,8 +114,6 @@ class EventSimulator:
         self.event_count = 0
         self.spikes = np.zeros((self.npix))
 
-    # One of the trick here for good performance is to avoid allocating
-    # memory here because we will be calling it very often
     def image_callback(self, new_image, new_time):
         if self.last_image is None:
             self.init(new_image, new_time)
@@ -125,7 +123,6 @@ class EventSimulator:
         assert new_image.shape == self.resolution
         new_image = new_image.reshape(-1)  # Free operation
 
-        # Copy is faster than reallocating memory
         np.copyto(self.current_image, new_image)
 
         delta_time = new_time - self.last_time
@@ -158,5 +155,3 @@ class EventSimulator:
         result.sort(order=["timestamp"], axis=0)
 
         return self.spikes, result
-
-        # TODO sort events or insert them into a heap
--- a/PythonClient/eventcamera_sim/event_simulator.py
+++ b/PythonClient/eventcamera_sim/event_simulator.py
@@ -35,7 +35,7 @@ def esim(
     n_pix_row,
 ):
     count = 0
-    maxSpikes = int(delta_time / (refractory_period_ns * 1e-3))
+    max_spikes = int(delta_time / (refractory_period_ns * 1e-3))
     for x in prange(x_end):
         itdt = np.log(current_image[x])
         it = np.log(previous_image[x])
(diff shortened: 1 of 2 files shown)


PythonClient/eventcamera_sim/test_event_sim.py
```diff
@@ -0,0 +1,110 @@
...
+            np.fromstring(response[0].image_data_uint8, dtype=np.uint8),
+            event_generator.rgb_image_shape,
+        )
+
+        img = cv2.cvtColor(img, cv2.COLOR_BGR2GRAY).astype(np.float32)
+        img = cv2.add(img, 0.001)
+
+        ts = time.time_ns()
+        ts_delta = (ts - event_generator.start_ts) * 1e-3
+        # start = time.time()
```
zimmy87, line 99: is this line necessary? it looks like a commented out debug line

PythonClient/eventcamera_sim/test_event_sim.py
```diff
@@ -0,0 +1,110 @@
...
+        event_img, events = event_generator.ev_sim.image_callback(img, ts_delta)
+        # print(f"event gen time: {time.time() - start}")
```
zimmy87, line 101: same as https://github.com/microsoft/AirSim/pull/3202/files#diff-aca498f945b6f7ea12be1442d344a97ab196a87357e2a430e07b0dfac1c33a24R99

PythonClient/eventcamera_sim/event_simulator.py
```diff
@@ -0,0 +1,157 @@
...
+        itdt = np.log(current_image[x])
+        it = np.log(previous_image[x])
+        deltaL = itdt - it
+
+        if np.abs(deltaL) < TOL:
+            continue
+
+        pol = np.sign(deltaL)
+
+        crossUpdate = pol * TOL
```
zimmy87, line 49: according to https://www.python.org/dev/peps/pep-0008/#function-and-variable-names local variables should be lower_case_with_underscores

zimmy87 (changes requested):
Most comments are style/debugging-related. Currently blocked from testing due to the last comment though.

PythonClient/eventcamera_sim/event_simulator.py
saihv: Fixed

PythonClient/eventcamera_sim/test_event_sim.py
saihv: removed

PythonClient/eventcamera_sim/test_event_sim.py
saihv: removed

PythonClient/eventcamera_sim/test_event_sim.py
```diff
@@ -0,0 +1,110 @@
...
+
+        if events is not None and events.shape[0] > 0:
+            if event_generator.save:
```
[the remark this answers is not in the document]
saihv: Fixed

zimmy87 (approved):
looks good to me

JamesYang110043:
> Hello, I want to use event camera in Airsim. when I run <test_event_sim.py> , the error comes as followes and the <events.pkl> is blank. (Python = 3.9 AirSim = 1.5 UE=4.26) **ERRORS** Initialized event camera simulator with sensor size: (64, 64) D:\RL\AirSimv1.5\PythonClient\eventcamera_sim\test_event_sim.py:91: DeprecationWarning: The binary mode of fromstring is deprecated, as it behaves surprisingly on unicode inputs. Use frombuffer instead np.fromstring(response[0].image_data_uint8, dtype=np.uint8), event_generator.rgb_image_shape,) D:\RL\ANACONDA\envs\ASclient\lib\site-packages\numba\core\typed_passes.py:329: NumbaPerformanceWarning: The keyword argument 'parallel=True' was specified but no transformation for parallel execution was possible.
> 
> To find out why, try turning on parallel diagnostics, see https://numba.readthedocs.io/en/stable/user/parallel.html#diagnostics for help.
> 
> File "event_simulator.py", line 24: @njit(parallel=True) def esim( ^
> 
> warnings.warn(errors.NumbaPerformanceWarning(msg, D:\RL\AirSimv1.5\PythonClient\eventcamera_sim\test_event_sim.py:91: DeprecationWarning: The binary mode of fromstring is deprecated, as it behaves surprisingly on unicode inputs. Use frombuffer instead np.fromstring(response[0].image_data_uint8, dtype=np.uint8), event_generator.rgb_image_shape,)

I have a same problem here, and I got the same error log.
Maybe "Namba" lead to this error. But I don't have any idea.
Did you solved it? Thx

PR #3180, opened 2020-12-03
Add Github Actions CI Checks

rajat2004:
Currently adds Ubuntu, Windows builds

TODO:

- [x] OSX
- [x] Unity build
- [x] Ubuntu ROS Wrapper compilation

Workflow to publish the GitHub pages can also be added, so the maintainers no longer need to do that manually 

See https://github.com/rajat2004/AirSim/pull/26 on my fork which shows the checks

PR #3209, opened 2020-12-12
Voxel grid construction

saihv:
### About
This PR introduces a feature that constructs ground truth voxel grids of the world directly from Unreal Engine. The voxel grids are stored in a [binvox format](https://www.patrickmin.com/binvox/binvox.html) which can then be converted by the user into an octomap .bt or any other relevant, desired format. Subsequently, these voxel grids/octomaps can be used within mapping/planning.

The logic for constructing the voxel grid is in WorldSimApi.cpp->createVoxelGrid(). For now, the assumption is that the voxel grid is a cube - and the API call is:

```
createVoxelGrid(self, position, x, y, z, res, of)

position (Vector3r): Global position around which voxel grid is centered in m
x, y, z (float): Size of each voxel grid dimension in m
res (float): Resolution of voxel grid in m
of (str): Name of output file to save voxel grid as
```

### How Has This Been Tested?
Simple test script:
```
import airsim
c = airsim.VehicleClient()
center = a.Vector3r(0, 0, 0)
c.createVoxelGrid(center, 100, 100, 100, 0.5, "map.binvox")
```

[viewvox](https://www.patrickmin.com/viewvox/) is a nifty little utility to then visualize the created binvox file. Similarly, `binvox2bt` can convert the binvox to an octomap file.

### Example voxel grid in Blocks:
image

### DIFF
--- a/PythonClient/airsim/client.py
+++ b/PythonClient/airsim/client.py
@@ -790,7 +790,7 @@ class VehicleClient:
         """
         self.client.call('simSetWind', wind)
 
-    def createVoxelGrid(self, position, x, y, z, res, of):
+    def simCreateVoxelGrid(self, position, x, y, z, res, of):
         """
         Construct and save a binvox-formatted voxel grid of environment
 
--- a/PythonClient/airsim/client.py
+++ b/PythonClient/airsim/client.py
@@ -790,6 +790,19 @@ class VehicleClient:
         """
         self.client.call('simSetWind', wind)
 
+    def createVoxelGrid(self, position, x, y, z, res, of):
+        """
+        Construct and save a binvox-formatted voxel grid of environment
+
+        Args:
+            position (Vector3r): Position around which voxel grid is centered in m
+            x, y, z (float): Size of each voxel grid dimension in m
+            res (float): Resolution of voxel grid in m
+            of (str): Name of output file to save voxel grid as
+
+        """
+        self.client.call('simCreateVoxelGrid', position, x, y, z, res, of)
+
 # -----------------------------------  Multirotor APIs ---------------------------------------------
 class MultirotorClient(VehicleClient, object):
     def __init__(self, ip = "", port = 41451, timeout_value = 3600):
--- a/PythonClient/airsim/client.py
+++ b/PythonClient/airsim/client.py
@@ -801,7 +801,7 @@ class VehicleClient:
             of (str): Name of output file to save voxel grid as
 
         """
-        self.client.call('simCreateVoxelGrid', position, x, y, z, res, of)
+        return self.client.call('simCreateVoxelGrid', position, x, y, z, res, of)
 
 # -----------------------------------  Multirotor APIs ---------------------------------------------
 class MultirotorClient(VehicleClient, object):
--- a/Unreal/Plugins/AirSim/Source/WorldSimApi.cpp
+++ b/Unreal/Plugins/AirSim/Source/WorldSimApi.cpp
@@ -161,6 +161,79 @@ AActor* WorldSimApi::createNewActor(const FActorSpawnParameters& spawn_params, c
     return NewActor;
 }
 
+void WorldSimApi::createVoxelGrid(const Vector3r& position, const int& x_size, const int& y_size, const int& z_size, const float& res, const std::string& output_file)
+{
+    int ncells_x = x_size / res;
+    int ncells_y = y_size / res;
+    int ncells_z = z_size / res;
+    
+    voxel_grid_.resize(ncells_x * ncells_y * ncells_z);
+
+    int ctr = 0;
+    float scale = 1 / res;
+    float scale_cm = scale * 100;
+    FCollisionQueryParams params;
+    params.bFindInitialOverlaps = true;
+    params.bTraceComplex = false;
+    params.TraceTag = "";
+    auto position_in_UE_frame = simmode_->getGlobalNedTransform().fromGlobalNed(position);
+    for (float i = 0; i < ncells_x; i++) {
+        for (float k = 0; k < ncells_z; k++) {
+            for (float j = 0; j < ncells_y; j++) {
+                int idx = i + ncells_x * (k + ncells_z * j);
+                FVector position = FVector((i - ncells_x /2) * scale_cm, (j - ncells_y /2) * scale_cm, (k - ncells_z /2) * scale_cm) + position_in_UE_frame;
+                voxel_grid_[idx] = (unsigned int)simmode_->GetWorld()->OverlapBlockingTestByChannel(position, FQuat::Identity, ECollisionChannel::ECC_Pawn, FCollisionShape::MakeBox(FVector(res/50)), params);
+
+            }
+        }
+    }
+    
+    std::ofstream* output = new std::ofstream(output_file, std::ios::out | std::ios::binary);
+    if (!output->good())
+    {
+        std::cerr << "Error: Could not open output file " << output << "!" << std::endl;
+        exit(1);
+    }
+
+    // Write the binvox file using run-length encoding
+    // where each pair of bytes is of the format (run value, run length)
+    *output << "#binvox 1\n";
+    *output << "dim " << ncells_x << " " << ncells_z << " " << ncells_y << "\n";
+    *output << "translate " << int(-x_size/2) << " " << int(-y_size / 2) << " " << int(-z_size / 2) << "\n";
+    *output << "scale " << scale << "\n";
+    *output << "data\n";
+    bool run_value = voxel_grid_[0];
+    unsigned int run_length = 0;
+    for (size_t i = 0; i < voxel_grid_.size(); ++i)
+    {
+        if (voxel_grid_[i] == run_value)
+        {
+            // This is a run (repeated bit value)
+            run_length++;
+            if (run_length == 255)
+            {
+                *output << static_cast<char>(run_value);
+                *output << static_cast<char>(run_length);
+                run_length = 0;
+            }
+        }
+        else
+        {
+            // End of a run
+            *output << static_cast<char>(run_value);
+            *output << static_cast<char>(run_length);
+            run_value = voxel_grid_[i];
+            run_length = 1;
+        }
+    }
+    if (run_length > 0)
+    {
+        *output << static_cast<char>(run_value);
+        *output << static_cast<char>(run_length);
+    }
+    output->close();
+}
+
 bool WorldSimApi::isPaused() const
 {
     return simmode_->isPaused();
--- a/Unreal/Plugins/AirSim/Source/WorldSimApi.cpp
+++ b/Unreal/Plugins/AirSim/Source/WorldSimApi.cpp
@@ -170,8 +170,7 @@ void WorldSimApi::createVoxelGrid(const Vector3r& position, const int& x_size, c
     voxel_grid_.resize(ncells_x * ncells_y * ncells_z);
 
     int ctr = 0;
-    float scale = res;
-    float scale_cm = scale * 100;
+    float scale_cm = res * 100;
     FCollisionQueryParams params;
     params.bFindInitialOverlaps = true;
     params.bTraceComplex = false;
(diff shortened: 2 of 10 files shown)


Unreal/Plugins/AirSim/Source/WorldSimApi.cpp
```diff
@@ -161,6 +161,79 @@ AActor* WorldSimApi::createNewActor(const FActorSpawnParameters& spawn_params, c
     return NewActor;
 }
 
+void WorldSimApi::createVoxelGrid(const Vector3r& position, const int& x_size, const int& y_size, const int& z_size, const float& res, const std::string& output_file)
+{
+    int ncells_x = x_size / res;
+    int ncells_y = y_size / res;
+    int ncells_z = z_size / res;
+    
+    voxel_grid_.resize(ncells_x * ncells_y * ncells_z);
```
rajat2004, line 170: `voxel_grid_` doesn't seem to have been declared anywhere

Unreal/Plugins/AirSim/Source/WorldSimApi.cpp
```diff
@@ -161,6 +161,79 @@ AActor* WorldSimApi::createNewActor(const FActorSpawnParameters& spawn_params, c
...
+            for (float j = 0; j < ncells_y; j++) {
+                int idx = i + ncells_x * (k + ncells_z * j);
+                FVector position = FVector((i - ncells_x /2) * scale_cm, (j - ncells_y /2) * scale_cm, (k - ncells_z /2) * scale_cm) + position_in_UE_frame;
+                voxel_grid_[idx] = (unsigned int)simmode_->GetWorld()->OverlapBlockingTestByChannel(position, FQuat::Identity, ECollisionChannel::ECC_Pawn, FCollisionShape::MakeBox(FVector(scale_cm /2)), params);
+
+            }
+        }
+    }
+    
+    std::ofstream* output = new std::ofstream(output_file, std::ios::out | std::ios::binary);
```
rajat2004, line 191: I think we can just do `std::ofstream output(output_file, std::ios::out | std::ios::binary)` and use the object later on directly, no need to `new` and dereference and all, also not sure but `new` can give memory leaks if not deleted in this case

rajat2004 (commented):
Wow, looks pretty awesome

Unreal/Plugins/AirSim/Source/WorldSimApi.cpp
saihv: Whoops, not sure how that line went missing from the commit. Fixed

Unreal/Plugins/AirSim/Source/WorldSimApi.cpp
saihv: You're right, safer this way too. Fixed

saihv:
@jonyMarino @zimmy87 Any ideas why the azure CI pipelines are not getting triggered for PRs?

rajat2004:
Yup, fixing the CI would be a big help, there's an open PR to add Gtihub Actions also, to replace Travis - https://github.com/microsoft/AirSim/pull/3180

Unreal/Plugins/AirSim/Source/WorldSimApi.cpp
```diff
@@ -161,6 +161,77 @@ AActor* WorldSimApi::createNewActor(const FActorSpawnParameters& spawn_params, c
...
+                FVector position = FVector((i - ncells_x /2) * scale_cm, (j - ncells_y /2) * scale_cm, (k - ncells_z /2) * scale_cm) + position_in_UE_frame;
+                voxel_grid_[idx] = simmode_->GetWorld()->OverlapBlockingTestByChannel(position, FQuat::Identity, ECollisionChannel::ECC_Pawn, FCollisionShape::MakeBox(FVector(scale_cm /2)), params);
+            }
+        }
+    }
+    
+    std::ofstream output(output_file, std::ios::out | std::ios::binary);
+    if (!output.good())
+    {
+        std::cerr << "Error: Could not open output file " << &output << "!" << std::endl;
```
rajat2004, line 192: Shouldn't it be `output_file` here, right now it would just print a memory address

Unreal/Plugins/AirSim/Source/WorldSimApi.cpp
```diff
@@ -161,6 +161,77 @@ AActor* WorldSimApi::createNewActor(const FActorSpawnParameters& spawn_params, c
...
+        exit(1);
```
rajat2004, line 193: Shouldn't it be `output_file` above in the error message, right now it would just print a memory address
Also, just a thought, do we want to exit if it fails to open the file, or just return `false`?

saihv: Good catch - that was a copy paste artifact I forgot to fix :-) 

I modified the function to return the status, and also display a UE message instead if the file cannot be opened.

Unreal/Plugins/AirSim/Source/WorldSimApi.cpp
```diff
@@ -161,6 +161,80 @@ AActor* WorldSimApi::createNewActor(const FActorSpawnParameters& spawn_params, c
...
+bool WorldSimApi::createVoxelGrid(const Vector3r& position, const int& x_size, const int& y_size, const int& z_size, const float& res, const std::string& output_file)
+{
+    bool success = false;
+    int ncells_x = x_size / res;
+    int ncells_y = y_size / res;
+    int ncells_z = z_size / res;
+    
+    voxel_grid_.resize(ncells_x * ncells_y * ncells_z);
+
+    int ctr = 0;
```
rajat2004, line 173: Unused variable

Unreal/Plugins/AirSim/Source/WorldSimApi.cpp
```diff
@@ -161,6 +161,80 @@ AActor* WorldSimApi::createNewActor(const FActorSpawnParameters& spawn_params, c
...
+    {
+        UE_LOG(LogTemp, Error, TEXT("Could not open output file for voxel grid!"));
+        return success;
+    }
+
+    // Write the binvox file using run-length encoding
+    // where each pair of bytes is of the format (run value, run length)
+    output << "#binvox 1\n";
+    output << "dim " << ncells_x << " " << ncells_z << " " << ncells_y << "\n";
+    output << "translate " << int(-x_size/2) << " " << int(-y_size / 2) << " " << int(-z_size/2) << "\n";
```
rajat2004, line 201: `x_size` and others are integers, so casting shouldn't be required here

Unreal/Plugins/AirSim/Source/WorldSimApi.cpp
```diff
@@ -161,6 +161,80 @@ AActor* WorldSimApi::createNewActor(const FActorSpawnParameters& spawn_params, c
...
+    output << "scale " << 1/x_size << "\n";
```
rajat2004, line 202: `x_size` is an integer here, so the output will always be 0 for `x_size > 1`, maybe it was meant to be `1.0f/x_size`

rajat2004 (commented):
Just some small nitpicks

Unreal/Plugins/AirSim/Source/WorldSimApi.cpp
saihv: Fixed

Unreal/Plugins/AirSim/Source/WorldSimApi.cpp
saihv: Fixed. This was a weird parameter, according to the binvox documentation, scale is important to tell how much to scale the grid by to fit in a 1x1x1 unit cube, but viewvox displays the voxel grid regardless of what value. Probably should keep an eye on this.

rajat2004:
Seems like Azure Pipelines is active now, from the [latest commit checks](https://github.com/microsoft/AirSim/runs/1553617586)
Though Windows build has failed

saihv:
It's the Windows container initialization that's failing during CI.

@jonyMarino Just a heads up
```
2020-12-14T22:53:40.8901202Z 
##[section]Starting: Initialize containers
2020-12-14T22:53:41.0464982Z 
##[command]C:\Program Files\Docker\docker.EXE version --format '{{.Server.APIVersion}}'
2020-12-14T22:53:42.6313274Z error during connect: Get http://%2F%2F.%2Fpipe%2Fdocker_engine/v1.40/version: open //./pipe/docker_engine: The system cannot find the file specified. In the default daemon configuration on Windows, the docker client must be run elevated to connect. This error may also indicate that the docker daemon is not running.
2020-12-14T22:53:42.6315842Z '
2020-12-14T22:53:42.8171413Z 
##[error]Exit code 1 returned from process: file name 'C:\Program Files\Docker\docker.EXE', arguments 'version --format '{{.Server.APIVersion}}''.
2020-12-14T22:53:42.8190394Z 
##[section]Finishing: Initialize containers
```

jonyMarino:
Hi @saihv, I do not know if you get a report with the CI results.

saihv:
@jonyMarino Thanks for re-enabling CI - will fix the Unity build shortly.

jonyMarino:
Thank @rajat2004. This came from his work with GitHub Actions (#3180). We still see what happens with Azure pipelines.

zimmy87 (changes requested):
Tested this locally and it works for me. Only comment is function naming-related.

PythonClient/airsim/client.py
```diff
@@ -790,6 +790,19 @@ def simSetWind(self, wind):
         """
         self.client.call('simSetWind', wind)
 
+    def createVoxelGrid(self, position, x, y, z, res, of):
```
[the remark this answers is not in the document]
saihv: You're right. Fixed, thanks for the catch

zimmy87 (approved):

PythonClient/airsim/client.py
```diff
@@ -790,6 +790,19 @@ def simSetWind(self, wind):
         """
         self.client.call('simSetWind', wind)
 
+    def simCreateVoxelGrid(self, position, x, y, z, res, of):
+        """
+        Construct and save a binvox-formatted voxel grid of environment
+
+        Args:
+            position (Vector3r): Position around which voxel grid is centered in m
+            x, y, z (float): Size of each voxel grid dimension in m
```
rajat2004, line 799: Should be `int`, not `float`, from the method arguments defined in C++

rajat2004 (commented):
A bit late now, just pointing it out so that it doesn't get forgotten, maybe will add it in a new PR (along with some doc updates)

GuangyaoSHI:
> ### About
> 
> This PR introduces a feature that constructs ground truth voxel grids of the world directly from Unreal Engine. The voxel grids are stored in a [binvox format](https://www.patrickmin.com/binvox/binvox.html) which can then be converted by the user into an octomap .bt or any other relevant, desired format. Subsequently, these voxel grids/octomaps can be used within mapping/planning.
> 
> The logic for constructing the voxel grid is in WorldSimApi.cpp->createVoxelGrid(). For now, the assumption is that the voxel grid is a cube - and the API call is:
> 
> ```
> createVoxelGrid(self, position, x, y, z, res, of)
> 
> position (Vector3r): Global position around which voxel grid is centered in m
> x, y, z (float): Size of each voxel grid dimension in m
> res (float): Resolution of voxel grid in m
> of (str): Name of output file to save voxel grid as
> ```
> 
> ### How Has This Been Tested?
> 
> Simple test script:
> 
> ```
> import airsim
> c = airsim.VehicleClient()
> center = a.Vector3r(0, 0, 0)
> c.createVoxelGrid(center, 100, 100, 100, 0.5, "map.binvox")
> ```
> 
> [viewvox](https://www.patrickmin.com/viewvox/) is a nifty little utility to then visualize the created binvox file. Similarly, `binvox2bt` can convert the binvox to an octomap file.
> ### Example voxel grid in Blocks:
> 
> image

Thank you for your work. I went over the C++ code. I think here x, y, z are not the the size of one voxel but the overall size of all voxels because I notice that you use x/res to compute number of cells.

zhudelong:
Very nice work! Can the voxel grid also model the objects that do not have an entity inside the world model, e.g., leaves?

Leon-Redfield:
Hi! I'm trying to get the global map in the format of PointCloud, and I tried to use this api simcreatevoxelGrid() but it seems that it doesn't work well for irregular objects such as trees with leaves. Is there any api or plan to add an api which models the whole part of the objects?

Anonymous-N:
hi! if there is anyway to get the semangtic of voxel?

Anonymous-N:
hi! Is there any way to get the semantic information of a voxel？thank you

kingchou007:
I was wondering how can I get the position of the block.

kingchou007:
> I was wondering how can I get the position of the block.

such as the position information of obstacles(x, y)

PR #3233, opened 2020-12-18
Rename DepthPlanner to DepthPlanar

### LINKED ISSUE #1839: Rename DepthPlanner to DepthPlanar
It seems obvious that the intended name of the DepthPlanner image type was actually DepthPlanar, and the documentation on this actually corroborates this theory very strongly, quote:

> For `ImageType = DepthPlanner`, you get depth in camera plan, i.e., all points that are in plan parallel to camera have same depth.

This is of course supposed to be

> For `ImageType = DepthPlanar`, you get depth in camera **plane**, i.e., all points that are **in-plane** parallel to camera have same depth.

saihv:
### About
Fixes a longstanding typo : DepthPlanner in camera types should be DepthPlanar.

### How Has This Been Tested?
Fix is mainly about renaming lines of code/assets. Sim should run normally.


rajat2004:
This seems to be pretty straightforward fix, I'm just a bit concerned about how it'll break existing scripts of users, a note should probably be added in the docs regarding this rename. Got a more eye-catching message, an exception could also be raised in the Python scripts when `DepthPlanner` is used, might be too much though

ahmed-elsaharti:
> Got a more eye-catching message, an exception could also be raised in the Python scripts when `DepthPlanner` is used,

@rajat2004 this is a really good idea. It'll at least guarantee that users who use this camera type find out about the change even if they do not read the docs.

rajat2004:
My previous comment is quite strange, really sorry for that, shouldn't post when sleepy :sweat_smile: 
What I was saying is that an exception will be thrown on the usage of `DepthPlanner`, should we raise a custom exception also? Though I'm not aware of any way to do that currently.
Another possibility is to actually go in the other direction and continue to allow usage of `DepthPlanner`, for some backwards compatibility. Not very much in favour of this though myself.

ahmed-elsaharti:
This is probably doable by defining `__getattribute__` within the `ImageType` class. However, it needs the outdated py scripts to have been creating instances of the class (`A = ImageType().Dep...`) rather than just copying the class definition (`A = ImageType.Dep...`). This probably isn't the case though so this might do more harm than good..

saihv:
Good call about backward compatibility issues. This is pretty hacky, but would probably serve the purpose.

```
class ImageType:
    Scene = 0
    DepthPlanar = 1
    DepthPerspective = 2
    DepthVis = 3
    DisparityNormalized = 4
    Segmentation = 5
    SurfaceNormals = 6
    Infrared = 7
    DepthPlanner = print("DepthPlanner has been (correctly) renamed to DepthPlanar. Please use ImageType.DepthPlanar instead."); raise AttributeError
```

ahmed-elsaharti:
@saihv testing this resulted in `print("DepthPlanner has........"); raise AttributeError` being executed on importing airsim which resulted in the error being raised right away.

saihv:
Ouch, yeah, you're right - it was an issue with my testing. That's to be expected as Python does instantiate classes in the background at import time. I guess another way to do it is a bit more convoluted, with a metaclass over image type:

```
class _ImageType(type):
    @property
    def Scene(cls):
        return 0
    # add other types here with corresponding numbers

    def __getattr__(self, key):
        if key == 'DepthPlanner':
            print("DepthPlanner has been (correctly) renamed to DepthPlanar. Please use ImageType.DepthPlanar instead.")

class ImageType(metaclass=_ImageType):
    pass
```

ahmed-elsaharti:
@saihv Ah! Nice workaround! This works perfectly and should point out the change to users using older scripts.

zimmy87:
Hey @saihv, are you planning on adding the metaclass over image type that raises an error upon usage of DepthPlanner? I don't see it in any of the commits for this PR.

saihv:
@zimmy87 Sorry, somehow forgot to push that commit. Could you try with the latest one?

jonyMarino:
Thanks, Sai for your contribution!
And thanks to @rajat2004, @ahmed-elsaharti and @zimmy87 for the reviews and ideas!

PR #3239, opened 2020-12-22
Documentation for voxel grid and event sim functionality

saihv:
#### About
Some documentation on the internal working, as well as how to use the corresponding API of the new features: voxel grid construction and event camera simulation.


docs/event_sim.md
```diff
@@ -0,0 +1,50 @@
...
+ev_sim = EventSimulator(W, H)
+```
+
+The actual computation of the events is triggered through a callback function, which is executed every time a new RGB image is obtained. The first time this function is called, due to the lack of a 'previous' image, it acts as an initialization of the event sim. 
+
+```
+event_img, events = ev_sim.image_callback(img, ts_delta)
+```
+The callback returns an event image as a one dimensional array of +1/-1 values, thus indicating only whether events were seen at each pixel, but not the timing/number of events. This one dimensional array can be converted into the red/blue event image as seen in the function `convert_event_img_rgb`. `events` is a numpy array of events, each of format `<x> <y> <timestamp> <pol>`.
+
```
jonyMarino, line 30: Sorry, maybe a little bit late. Is it correct to call it a callback? I see a common synchronous function there. Am I missing something?

docs/event_sim.md
```diff
@@ -0,0 +1,50 @@
...
+2. The log luminance threshold `TOL` that determines whether or not a detected change counts as an event.
+
+Note: There is also currently a max limit on the number of events generated per pair of images, which can also be tuned.
+
+
+#### Algorithm
+The working of the event simulator loosely follows this set of operations:
+1. Take the difference between the log intensities of the current and previous frames. 
+2. Iterating over all pixels, calculate the polarity for each each pixel based on a threshold of change in log intensity. 
+3. Determine the number of events to be fired per pixel, based on extent of intensity change over the threshold. Let $N_{max}$ be the maximum number of events that can occur at a single pixel, then the total number of firings to be simulated at pixel location $u$ would be $N_e(u) = min(N_{max}, \frac{\Delta L(u)}{TOL})$.
```
jonyMarino, line 45: How can I test if the formulas are going to be rendered well? Running mkdocs?

jonyMarino (changes requested):
Minor things.
I still need to review the voxel grid doc.
Thanks for this contribution!

docs/voxel_grid.md
```diff
@@ -0,0 +1,55 @@
...
+            voxel_grid_[idx] = simmode_->GetWorld()->OverlapBlockingTestByChannel(position, FQuat::Identity, ECollisionChannel::ECC_Pawn, FCollisionShape::MakeBox(FVector(scale_cm /2)), params);
+        }
+    }
+}
+```
+
+The occupancy of the map is calculated iteratively over all discretized cells, which can make it an intensive operation depending on the resolution of the cells, and the total size of the area being measured. If the user's map of interest does not change much, it is possible to run the voxel grid operation once on this map, and save the voxel grid and reuse it. For performance, or with dynamic environments, we recommend running the voxel grid generation for a small area around the robot; and subsequently use it for local planning purposes.
+
+
+The voxel grids are stored in the binvox format which can then be converted by the user into an octomap .bt or any other relevant, desired format. Subsequently, these voxel grids/octomaps can be used within mapping/planning. [viewvox](https://www.patrickmin.com/viewvox/) is a nifty little utility to then visualize the created binvox file. Similarly, `binvox2bt` can convert the binvox to an octomap file.
```
zimmy87, line 37: [viewvox] should be capitalized since it's the first word of a sentence

docs/voxel_grid.md
```diff
@@ -0,0 +1,55 @@
+AirSim provides a feature that constructs ground truth voxel grids of the world directly from Unreal Engine. A voxel grid is a representation of the occupancy of a given world/map, by discretizing into cells of a certain size; and recording a voxel if that particular location is occupied. 
+
+The logic for constructing the voxel grid is in WorldSimApi.cpp->createVoxelGrid(). For now, the assumption is that the voxel grid is a cube - and the API call is of the structure:
+
+```
+createVoxelGrid(self, position, x, y, z, res, of)
```
zimmy87, line 6: The API that the end user will end up calling is simCreateVoxelGrid(), correct? I feel we should list the external name of the API rather than the internal method name.

docs/voxel_grid.md
```diff
@@ -0,0 +1,55 @@
...
+
+```
+import airsim
+c = airsim.VehicleClient()
+center = airsim.Vector3r(0, 0, 0)
+output_path = "~/map.binvox"
+c.simCreateVoxelGrid(center, 100, 100, 100, 0.5, output_path)
+```
+
+And visualized through `viewvox ~/map.binvox`.
```
zimmy87, line 55: When I try to run this example locally, I get the following error: "Error opening [\~/map.binvox]". Does "\~/" evaluate to a valid path in windows? I can't seem to find the outputted map.binvox file anywhere in my users/ folder, which seems like it would be the equivalent location in windows

zimmy87 (changes requested):
the voxel grid section looks good to me apart from a few nitpick items

docs/voxel_grid.md
saihv: This should match whatever was commanded in `output_path` when calling simCreateVoxelGrid. You can see in Line 51 I was using `~/.map.binvox` as my output path, but you can use whichever path.

zimmy87: I double-checked what output_path was set to in the example script and confirmed it was the same as the script above: `output_path = "~/map.binvox"`. I'm still getting the same error when I run `viewvox ~/map.binvox`. On Ubuntu, ~/ evaluates to the user's home directory, but I don't think it does the same thing in Windows. When I try to navigate to the path with `cd ~/` in Windows, I get an error: `The system cannot find the path specified.`

saihv: @zimmy87 Sure, I can make that change. This documentation was just supposed to be illustrative of how to use the function, it was originally written under the assumption that the user would change it as appropriate. And yes, `~` does not exist in Windows.

docs/event_sim.md
```diff
@@ -0,0 +1,50 @@
+AirSim provides a Python-based event camera simulator, aimed at performance and ability to run in real-time along with the sim.
+
+#### Event cameras
+An event camera is a special vision sensor that measures changes in logarithmic brightness and only reports 'events', i.e., a set of bytes every time the absolute change exceeds a certain threshold. The event camera only reports the timestamp of the measurement, pixel location and the polarity: which is a +1/-1 based on whether the logarithmic brightness increased or decreased. Most event cameras have a temporal resolution of the order of microseconds, making them significantly faster than RGB sensors, and also demonstrate a high dynamic range and low motion blur. More details about event cameras can be found in [this tutorial from RPG-UZH](http://rpg.ifi.uzh.ch/docs/scaramuzza/Tutorial_on_Event_Cameras_Scaramuzza.pdf)
```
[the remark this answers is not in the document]
saihv: Clarified.

docs/event_sim.md
saihv: Clarified (it's just the way the function is behaving similar to)

docs/voxel_grid.md
saihv: Fixed.

docs/voxel_grid.md
saihv: Fixed.

docs/event_sim.md
saihv: This I am not entirely sure about. Markdown preview works fine for me in VS code. Are there any other docs that have used latex formulae?

docs/event_sim.md
```diff
@@ -0,0 +1,50 @@
...
+
+`<x> <y> <timestamp> <pol>`
+
+x and y are the pixel locations of the event firing, timestamp is the global timestamp in microseconds and pol is either +1/-1 depending on whether the brightness increased or decreased. Along with this bytestream, an accumulation of events over a 2D frame is also constructed, known as an 'event image' that visualizes +1 events as red and -1 as blue pixels. An example event image is shown below:
+
+![image](images/event_sim.png)
+
+#### Usage
+An example script to run the event simulator alongside AirSim is located at https://github.com/microsoft/AirSim/blob/master/PythonClient/eventcamera_sim/test_event_sim.py. The implementation of the actual event simulation, written in Python and numba, is at https://github.com/microsoft/AirSim/blob/master/PythonClient/eventcamera_sim/event_simulator.py. The event simulator is initialized as follows, with the arguments controlling the resolution of the camera.
```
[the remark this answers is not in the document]
saihv: Done.

docs/voxel_grid.md
saihv: Fixed.

rajat2004:
The files need to be added to `mkdocs.yml` as well for them to appear on the site

Issue #2419, opened 2020-02-14
Question: Energy level of the multirotor


charbelmansour:
Hello,
I find that there exist these sensors in Airsim:
-Camera
-IMU
-Magnetometer
-GPS
-Barometer
-Distance sensor
-Lidar
Is there a sensor related to the energy of a multirotor? I mean if there exists a sensor indicating how much energy remains in the multirotor.
If yes, can you tell me please how it works, I mean by what factor does the energy level of the multirotor decrease based on the traveled distance.
Best Regards,
Thank you for your help.

rajat2004:
Currently, there's no battery on energy consumption calculation in AirSim. https://github.com/microsoft/AirSim/issues/672 could be helpful for implementing such a feature

aiaarif:
You can take a look at this repo. They claim to have implemented the energy model. https://github.com/srivatsankrishnan/AirSim-1

charbelmansour:
okay thank you I will check the site.

charbelmansour:
Can you tell me please where I can find rpc-base-reset-plus-energy branch in this repository
https://github.com/srivatsankrishnan/AirSim-1 @aiaarif

aiaarif:
You can switch to different branch by clicking branch icon and selecting the appropriate one. Github should have a nice tutorial on that.

charbelmansour:
Hello @aiaarif 

I have cloned the project and run it on unreal engine. The good news that it works fine.

However, we are running this simulation on Window and the drone is not flying.

Can you please advise us how we can make the drone fly. We have imported the Hello Drone folder but still the drone is not flying.

Thank you for your help,

Charbel

aiaarif:
Please refer to the documentation of Airsim. They have a step by step by guide for that.

charbelmansour:
Hello, I cloned this project: https://github.com/srivatsankrishnan/AirSim-1
I want to get information about the energy of the drone. In this repository, they have implemented a method getTripStats that returns information about the trip including the energy consumed.
I am able to control the drone using python commands.
I am trying to run this code:

import airsim

import numpy as np
import os
import tempfile
import pprint

client = airsim.MultirotorClient(ip="127.0.0.1")
client.confirmConnection()
client.enableApiControl(True)
client.armDisarm(True)

res = client.getTripStats()
print (res.flight_time)


I'm getting the following error:
[...]

Issue #672, opened 2017-12-09
Is there a method to find out about drone's energy consumption?


git-moeen:
I'd like to know if it's possible to check the energy consumption of the drone? (after it covers a certain path, let's say). I'd like to use this information for path planning.

saihv:
Just a very high level idea: I believe the drones in AirSim are modeled with the MT2212 motors and the GWS 9X5 props in mind. There should also be a thrust calculation somewhere in the code. You should be able to relate the commanded throttle to the thrust, and subsequently to motor current through the thrust vs throttle and current vs thrust curves from the motor datasheets. Of course, this would be approximate and only for the motors.

sytelus:
I would suggest to look in to [this introduction, page 3](https://www.researchgate.net/file.PostFileLoader.html?id=57795c0ff7b67efa2d7d0171&assetKey=AS%3A379835361841153%401467571215713). The power consumed by a rotor is given by (torque * angular velocity * constant). These quantities are available in [this function](https://github.com/Microsoft/AirSim/blob/master/AirLib/include/vehicles/multirotor/Rotor.hpp#L115) so you can easily compute power consumed per rotor.  Then in [this function](https://github.com/Microsoft/AirSim/blob/master/AirLib/include/vehicles/multirotor/MultiRotor.hpp#L75) you can sum total power consumed by all rotors + add some fixed amount for onboard electronics.

RVBldr:
Internal VSTF 7415

kaioforte:
Hi, I was trying to calculate battery consumption of the drone, following the tip given by @sytelus . I could not find a way to access in code the info required to the calculations (torque, angular velocity) present in this [file](https://github.com/Microsoft/AirSim/blob/master/AirLib/include/vehicles/multirotor/Rotor.hpp#L115) using the API or any other method.

Do you have a solution for this problem?

jonyMarino:
Closed due to inactivity. Doc update pending

PR #3242, opened 2020-12-23
Rotor output API

ahmed-elsaharti:
Potentially helps with : #672 #2419   

## About
This PR adds an API that returns each rotor's speed, torque, and thrust. 
This is potentially useful for the future development of energy/battery level simulation within AirSim using @sytelus 's method outlined [here](https://github.com/microsoft/AirSim/issues/672#issuecomment-350904172)

~~The method used here is **super** crude and uses [a global variable](https://github.com/ahmed-elsaharti/AirSim/blob/03e044fe36eded643a54155e5fb0ed7b142bc8c8/AirLib/include/common/Common.hpp#L18-L24) that gets updated with the rotor parameters with every `updateRenderedState` while waiting for the `getRotorStates` API call. This probably is not the most elegant way of doing this.
**ie:**~~ Any testing and/or suggestions on how this could be improved are very welcome
Also, this is set up to work for quadrotors but scaling the struct up to take up any number of rotors is easy.

## How Has This Been Tested?
This was tested with the Blocks environment on UE 4.24 with Simpleflight and PX4 SITL. No issues so far.

## Screenshots:
On the python side here's the basic structure:
getRotorStates() (type = ~~RotorStateAPI~~ RotorStates) contains -> ~~rotors_ rotors (type = ~~RotorSTTVector~~ RotorVector) contains ->~~ rotors (type = list of rotors[0]-rotors[3] dicts) contains -> ['speed'], ['thrust'] and ['torque_scaler'] as floats

`print(client.getRotorStates())`:
image

`print(client.getRotorStates().rotors_.rotor[0])`:
image

### DIFF
--- a/AirLib/include/api/RpcLibAdapatorsBase.hpp
+++ b/AirLib/include/api/RpcLibAdapatorsBase.hpp
@@ -64,37 +64,7 @@ public:
             return d;
         }
     };
-    struct RotorVector {
-        std::vector<RotorParameters> rotor;
 
-        MSGPACK_DEFINE_MAP(rotor);
-
-        RotorVector()
-        {}
-
-        RotorVector(const msr::airlib::RotorVector& s)
-        {
-            rotor.clear();
-            for (unsigned int i = 0; i < s.rotor.size(); i++)
-            {
-                rotor.push_back(RotorParameters());
-                rotor[i] = s.rotor[i];
-            }
-        }
-
-        msr::airlib::RotorVector to() const
-        {
-            msr::airlib::RotorVector d;
-            for (unsigned int i = 0; i < rotor.size(); i++)
-            {
-                d.rotor.push_back(msr::airlib::RotorParameters());
-                d.rotor[i].speed = rotor[i].speed;
-                d.rotor[i].thrust = rotor[i].thrust;
-                d.rotor[i].torque_scaler = rotor[i].torque_scaler;
-            }
-            return d;
-        }
-    };
     struct Vector3r {
         msr::airlib::real_T x_val = 0, y_val = 0, z_val = 0;
         MSGPACK_DEFINE_MAP(x_val, y_val, z_val);
--- a/AirLib/include/api/RpcLibAdapatorsBase.hpp
+++ b/AirLib/include/api/RpcLibAdapatorsBase.hpp
@@ -36,6 +36,64 @@ public:
             d.push_back(TDest(s.at(i)));
     }
 
+    struct RotorParameters {
+        msr::airlib::real_T thrust;
+        msr::airlib::real_T torque_scaler;
+        msr::airlib::real_T speed;
+
+        MSGPACK_DEFINE_MAP(thrust, torque_scaler, speed);
+
+        RotorParameters()
+        {}
+
+        RotorParameters(const msr::airlib::RotorParameters& s)
+        {
+            thrust = s.thrust;
+            torque_scaler = s.torque_scaler;
+            speed = s.speed;
+        }
+
+        msr::airlib::RotorParameters to() const
+        {
+            msr::airlib::RotorParameters d;
+
+            d.thrust = thrust;
+            d.torque_scaler = torque_scaler;
+            d.speed = speed;
+
+            return d;
+        }
+    };
+    struct RotorVector {
+        std::vector<RotorParameters> rotor;
+
+        MSGPACK_DEFINE_MAP(rotor);
+
+        RotorVector()
+        {}
+
+        RotorVector(const msr::airlib::RotorVector& s)
+        {
+            rotor.clear();
+            for (unsigned int i = 0; i < s.rotor.size(); i++)
+            {
+                rotor.push_back(RotorParameters());
+                rotor[i] = s.rotor[i];
+            }
+        }
+
+        msr::airlib::RotorVector to() const
+        {
+            msr::airlib::RotorVector d;
+            for (unsigned int i = 0; i < d.rotor.size(); i++)
+            {
+                d.rotor[i].speed = rotor[i].speed;
+                d.rotor[i].thrust = rotor[i].thrust;
+                d.rotor[i].torque_scaler = rotor[i].torque_scaler;
+            }
+            return d;
+        }
+    };
     struct Vector3r {
         msr::airlib::real_T x_val = 0, y_val = 0, z_val = 0;
         MSGPACK_DEFINE_MAP(x_val, y_val, z_val);
--- a/AirLib/include/api/RpcLibAdapatorsBase.hpp
+++ b/AirLib/include/api/RpcLibAdapatorsBase.hpp
@@ -36,29 +36,6 @@ public:
             d.push_back(TDest(s.at(i)));
     }
 
-    struct RotorParameters {
-        msr::airlib::real_T thrust;
-        msr::airlib::real_T torque_scaler;
-        msr::airlib::real_T speed;
-
-        MSGPACK_DEFINE_MAP(thrust, torque_scaler, speed);
-
-        RotorParameters()
-        {}
-
-        RotorParameters(const msr::airlib::RotorParameters& s)
-        {
-            thrust = s.thrust;
-            torque_scaler = s.torque_scaler;
-            speed = s.speed;
-        }
-
-        msr::airlib::RotorParameters to() const
-        {
-            return msr::airlib::RotorParameters(thrust, torque_scaler, speed);
-        }
-    };
-
     struct Vector3r {
         msr::airlib::real_T x_val = 0, y_val = 0, z_val = 0;
         MSGPACK_DEFINE_MAP(x_val, y_val, z_val);
--- a/AirLib/include/api/RpcLibAdapatorsBase.hpp
+++ b/AirLib/include/api/RpcLibAdapatorsBase.hpp
@@ -55,13 +55,7 @@ public:
 
         msr::airlib::RotorParameters to() const
         {
-            msr::airlib::RotorParameters d;
-
-            d.thrust = thrust;
-            d.torque_scaler = torque_scaler;
-            d.speed = speed;
-
-            return d;
+            return msr::airlib::RotorParameters(thrust, torque_scaler, speed);
         }
     };
 
--- a/AirLib/include/api/RpcLibAdapatorsBase.hpp
+++ b/AirLib/include/api/RpcLibAdapatorsBase.hpp
@@ -85,8 +85,9 @@ public:
         msr::airlib::RotorVector to() const
         {
             msr::airlib::RotorVector d;
-            for (unsigned int i = 0; i < d.rotor.size(); i++)
+            for (unsigned int i = 0; i < rotor.size(); i++)
             {
+                d.rotor.push_back(msr::airlib::RotorParameters());
                 d.rotor[i].speed = rotor[i].speed;
                 d.rotor[i].thrust = rotor[i].thrust;
                 d.rotor[i].torque_scaler = rotor[i].torque_scaler;
--- a/AirLib/include/common/CommonStructs.hpp
+++ b/AirLib/include/common/CommonStructs.hpp
@@ -16,6 +16,11 @@ struct RotorParameters {
 
     RotorParameters()
     {}
+
+    RotorParameters(const real_T& thrust_val, const real_T& torque_scaler_val, const real_T& speed_val)
+        : thrust(thrust_val), torque_scaler(torque_scaler_val), speed(speed_val)
+    {
+    }
 };
 
 //velocity
--- a/AirLib/include/common/CommonStructs.hpp
+++ b/AirLib/include/common/CommonStructs.hpp
@@ -8,20 +8,6 @@
 #include <ostream>
 
 namespace msr { namespace airlib {
-//rotor feedback
-struct RotorParameters {
-    real_T thrust = 0;
-    real_T torque_scaler = 0;
-    real_T speed = 0;
-
-    RotorParameters()
-    {}
-
-    RotorParameters(const real_T& thrust_val, const real_T& torque_scaler_val, const real_T& speed_val)
-        : thrust(thrust_val), torque_scaler(torque_scaler_val), speed(speed_val)
-    {
-    }
-};
 
 //velocity
 struct Twist {
--- a/AirLib/include/common/CommonStructs.hpp
+++ b/AirLib/include/common/CommonStructs.hpp
@@ -8,6 +8,21 @@
 #include <ostream>
 
 namespace msr { namespace airlib {
+//rotor feedback
+struct RotorParameters {
+    real_T thrust = 0;
+    real_T torque_scaler = 0;
+    real_T speed = 0;
+
+    RotorParameters()
+    {}
+};
+struct RotorVector {
+    std::vector<RotorParameters> rotor;
+
+    RotorVector()
+    {}
+};
 
 //velocity
 struct Twist {
--- a/AirLib/include/common/CommonStructs.hpp
+++ b/AirLib/include/common/CommonStructs.hpp
@@ -17,12 +17,6 @@ struct RotorParameters {
     RotorParameters()
     {}
 };
-struct RotorVector {
-    std::vector<RotorParameters> rotor;
-
-    RotorVector()
-    {}
-};
 
 //velocity
 struct Twist {
--- a/AirLib/include/vehicles/multirotor/api/MultirotorApiBase.hpp
+++ b/AirLib/include/vehicles/multirotor/api/MultirotorApiBase.hpp
@@ -149,7 +149,7 @@ public: //these APIs uses above low level APIs
     }
 
     /******************* rotors' states setter ********************/
-    void setRotorStates(RotorStates& rotor_states)
+    void setRotorStates(const RotorStates& rotor_states)
     {
         rotor_states_ = rotor_states;
     }
--- a/AirLib/include/vehicles/multirotor/api/MultirotorApiBase.hpp
+++ b/AirLib/include/vehicles/multirotor/api/MultirotorApiBase.hpp
@@ -123,6 +123,11 @@ public: //these APIs uses above low level APIs
         float obs_avoidance_vel, const Vector3r& origin, float xy_length, float max_z, float min_z);
 
     /************************* high level status APIs *********************************/
+    RotorStates getRotorStates() const
+    {
+        return rotor_states_;
+    }
+
     MultirotorState getMultirotorState() const
     {
         MultirotorState state;
@@ -141,6 +146,12 @@ public: //these APIs uses above low level APIs
     virtual void cancelLastTask() override
     {
         token_.cancel();
+    }
+
+    /******************* rotors' states setter ********************/
+    void setRotorStates(RotorStates rotor_states)
+    {
+        rotor_states_ = rotor_states;
     }
 
 protected: //utility methods
@@ -341,6 +352,7 @@ private: //variables
     float landing_vel_ = 0.2f; //velocity to use for landing
     float approx_zero_vel_ = 0.05f;
     float approx_zero_angular_vel_ = 0.01f;
+    RotorStates rotor_states_;
 };
 
 }} //namespace
--- a/AirLib/include/vehicles/multirotor/api/MultirotorApiBase.hpp
+++ b/AirLib/include/vehicles/multirotor/api/MultirotorApiBase.hpp
@@ -149,7 +149,7 @@ public: //these APIs uses above low level APIs
     }
 
     /******************* rotors' states setter ********************/
-    void setRotorStates(RotorStates rotor_states)
+    void setRotorStates(RotorStates& rotor_states)
     {
         rotor_states_ = rotor_states;
     }
--- a/AirLib/include/vehicles/multirotor/api/MultirotorCommon.hpp
+++ b/AirLib/include/vehicles/multirotor/api/MultirotorCommon.hpp
@@ -29,6 +29,13 @@ struct RotorParameters {
         : thrust(thrust_val), torque_scaler(torque_scaler_val), speed(speed_val)
     {
     }
+
+    void update(const real_T& thrust_val, const real_T& torque_scaler_val, const real_T& speed_val)
+    {
+        thrust = thrust_val;
+        torque_scaler = torque_scaler_val;
+        speed = speed_val;
+    }
 };
 
 struct RotorStates {
--- a/AirLib/include/vehicles/multirotor/api/MultirotorCommon.hpp
+++ b/AirLib/include/vehicles/multirotor/api/MultirotorCommon.hpp
@@ -16,7 +16,18 @@ enum class LandedState : uint {
     Landed = 0,
     Flying = 1
 };
+// Struct for rotor state API
+struct RotorStates {
+    RotorVector rotors;
+    uint64_t timestamp;
 
+    RotorStates()
+    {}
+    RotorStates(const RotorVector& rotors_val, uint64_t timestamp_val)
+        : rotors(rotors_val), timestamp(timestamp_val)
+    {
+    }
+};
 //Yaw mode specifies if yaw should be set as angle or angular velocity around the center of drone
 struct YawMode {
     bool is_rate = true;
--- a/AirLib/include/vehicles/multirotor/api/MultirotorCommon.hpp
+++ b/AirLib/include/vehicles/multirotor/api/MultirotorCommon.hpp
@@ -18,12 +18,12 @@ enum class LandedState : uint {
 };
 // Struct for rotor state API
 struct RotorStates {
-    RotorVector rotors;
+    std::vector<RotorParameters> rotors;
     uint64_t timestamp;
 
     RotorStates()
     {}
-    RotorStates(const RotorVector& rotors_val, uint64_t timestamp_val)
+    RotorStates(const std::vector<RotorParameters>& rotors_val, uint64_t timestamp_val)
         : rotors(rotors_val), timestamp(timestamp_val)
     {
     }
--- a/AirLib/include/vehicles/multirotor/api/MultirotorCommon.hpp
+++ b/AirLib/include/vehicles/multirotor/api/MultirotorCommon.hpp
@@ -16,7 +16,21 @@ enum class LandedState : uint {
     Landed = 0,
     Flying = 1
 };
-// Struct for rotor state API
+// Structs for rotor state API
+struct RotorParameters {
+    real_T thrust = 0;
+    real_T torque_scaler = 0;
+    real_T speed = 0;
+
+    RotorParameters()
+    {}
+
+    RotorParameters(const real_T& thrust_val, const real_T& torque_scaler_val, const real_T& speed_val)
+        : thrust(thrust_val), torque_scaler(torque_scaler_val), speed(speed_val)
+    {
+    }
+};
+
 struct RotorStates {
     std::vector<RotorParameters> rotors;
     uint64_t timestamp;
@@ -28,6 +42,7 @@ struct RotorStates {
     {
     }
 };
+
 //Yaw mode specifies if yaw should be set as angle or angular velocity around the center of drone
 struct YawMode {
     bool is_rate = true;
--- a/AirLib/include/vehicles/multirotor/api/MultirotorRpcLibAdapators.hpp
+++ b/AirLib/include/vehicles/multirotor/api/MultirotorRpcLibAdapators.hpp
@@ -39,6 +39,27 @@ public:
         }
     };
 
+    struct RotorStates {
+        RotorVector rotors;
+        uint64_t timestamp;
+
+        MSGPACK_DEFINE_MAP(rotors, timestamp);
+
+        RotorStates()
+        {}
+
+        RotorStates(const msr::airlib::RotorStates& s)
+        {
+            rotors = s.rotors;
+            timestamp = s.timestamp;
+        }
+
+        msr::airlib::RotorStates to() const
+        {
+            return msr::airlib::RotorStates(rotors.to(), timestamp);
+        }
+    };
+
     struct MultirotorState {
         CollisionInfo collision;
         KinematicsState kinematics_estimated;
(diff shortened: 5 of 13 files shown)


AirLib/include/api/RpcLibAdapatorsBase.hpp
```diff
@@ -36,6 +36,62 @@ class RpcLibAdapatorsBase {
             d.push_back(TDest(s.at(i)));
     }
 
+    struct RotorSTTParameters {
```
zimmy87, line 39: Does STT here stand for "speed, thrust, torque_scalar"? If so, this strikes me running counter to the recommended ISO standard [here](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#nl5-avoid-encoding-type-information-in-names); i.e. I don't feel it's necessary to include information about a struct's members in its name.

AirLib/include/api/RpcLibAdapatorsBase.hpp
```diff
@@ -36,6 +36,62 @@ class RpcLibAdapatorsBase {
...
+            msr::airlib::RotorSTTParameters d;
+
+            d.thrust = thrust;
+            d.torque_scaler = torque_scaler;
+            d.speed = speed;
+
+            return d;
+        }
+    };
+    struct RotorSTTVector {
```
zimmy87, line 67: same comment as with RotorSTTParameters about possibly removing "STT" from the name

AirLib/src/vehicles/multirotor/api/MultirotorApiBase.cpp
```diff
@@ -11,6 +11,8 @@
 #include <iostream>
 #include <fstream>
 
+// Global var for rotor state API
+RotorParameters rotorparamsGlobal[4];
```
zimmy87, line 15: this can be removed if the recommended changes for replacing this global with a private variable are implemented elsewhere

zimmy87 (changes requested):
Hi @ahmed-elsaharti, thank you for creating this PR! I have a few comments that describe an alternative to using the global variable in your code. Most of my other comments deal with naming conventions and style guidelines. Please feel free to ask any questions if you'd like extra clarification about any of my comments.

AirLib/include/common/CommonStructs.hpp
```diff
@@ -8,6 +8,21 @@
 #include <ostream>
 
 namespace msr { namespace airlib {
+//rotor feedback
+struct RotorSTTParameters {
+        real_T thrust = 0;
```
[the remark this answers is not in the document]
ahmed-elsaharti: Fixed in 3ee3cad

AirLib/include/common/CommonStructs.hpp
```diff
@@ -8,6 +8,21 @@
...
+        real_T torque_scaler = 0;
+        real_T speed = 0;
+
+        RotorSTTParameters()
+        {}
+    };
+struct RotorSTTVector {
+    RotorSTTParameters rotor[4];
+
+        RotorSTTVector()
```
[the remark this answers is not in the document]
ahmed-elsaharti: Fixed in 3ee3cad

AirLib/include/vehicles/multirotor/api/MultirotorCommon.hpp
```diff
@@ -16,7 +16,18 @@ enum class LandedState : uint {
     Landed = 0,
     Flying = 1
 };
+// Struct for rotor state API
+struct RotorStateAPI {
+    RotorSTTVector rotors_;
```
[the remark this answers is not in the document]
ahmed-elsaharti: Fixed in 3ee3cad

AirLib/include/vehicles/multirotor/api/MultirotorRpcLibAdapators.hpp
```diff
@@ -39,6 +39,27 @@ class MultirotorRpcLibAdapators : public RpcLibAdapatorsBase {
         }
     };
 
+    struct RotorStateAPI {
+        RotorSTTVector rotors_;
```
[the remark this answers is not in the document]
ahmed-elsaharti: Fixed in 3ee3cad

PythonClient/airsim/types.py
```diff
@@ -217,6 +217,8 @@ def nanPose():
     def containsNan(self):
         return (self.position.containsNan() or self.orientation.containsNan())
 
+class RotorSTTVector(MsgpackMixin):
+    Parameters = []
```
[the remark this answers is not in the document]
ahmed-elsaharti: Fixed in 3ee3cad

PythonClient/airsim/types.py
```diff
@@ -359,6 +361,10 @@ class MultirotorState(MsgpackMixin):
     ready_message = ""
     can_arm = False
 
+class RotorStateAPI(MsgpackMixin):
+    timestamp = np.uint64(0)
+    rotors_ = RotorSTTVector()
```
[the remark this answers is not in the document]
ahmed-elsaharti: Fixed in 3ee3cad

AirLib/include/vehicles/multirotor/api/MultirotorCommon.hpp
```diff
@@ -16,7 +16,18 @@ enum class LandedState : uint {
     Landed = 0,
     Flying = 1
 };
+// Struct for rotor state API
+struct RotorStateAPI {
```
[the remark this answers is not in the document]
ahmed-elsaharti: Fixed in 3ee3cad

AirLib/include/vehicles/multirotor/api/MultirotorRpcLibAdapators.hpp
```diff
@@ -39,6 +39,27 @@ class MultirotorRpcLibAdapators : public RpcLibAdapatorsBase {
         }
     };
 
+    struct RotorStateAPI {
```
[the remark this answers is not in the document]
ahmed-elsaharti: Fixed in 3ee3cad

PythonClient/airsim/client.py
```diff
@@ -1231,7 +1231,17 @@ def getMultirotorState(self, vehicle_name = ''):
         """
         return MultirotorState.from_msgpack(self.client.call('getMultirotorState', vehicle_name))
     getMultirotorState.__annotations__ = {'return': MultirotorState}
+    # query rotor states
+    def getRotorStates(self, vehicle_name = ''):
+        """
+        Args:
+            vehicle_name (str, optional): Vehicle to get the rotor state of
```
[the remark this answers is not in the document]
ahmed-elsaharti: Added more elaboration on what the API is returning. Please let me know if more/less detail is required

Unreal/Plugins/AirSim/Source/Vehicles/Multirotor/MultirotorPawnSimApi.cpp
```diff
@@ -74,6 +74,10 @@ void MultirotorPawnSimApi::updateRenderedState(float dt)
     //update rotor poses
     for (unsigned int i = 0; i < rotor_count_; ++i) {
         const auto& rotor_output = multirotor_physics_body_->getRotorOutput(i);
+        // update global rotor variable
+        rotorparamsGlobal[i].speed = rotor_output.speed;
+        rotorparamsGlobal[i].thrust = rotor_output.thrust;
+        rotorparamsGlobal[i].torque_scaler = rotor_output.torque_scaler;
```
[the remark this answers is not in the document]
ahmed-elsaharti: Thanks for the elaborate instructions David :D. I implemented the methodology you described in e67c2b2.

AirLib/include/common/Common.hpp
```diff
@@ -15,6 +15,14 @@
...
 #include "VectorMath.hpp"
 
+// Global var for rotor state API
+struct RotorParameters {
+    float speed;
+    float torque_scaler;
+    float thrust;
+};
+extern RotorParameters rotorparamsGlobal[4];
+
```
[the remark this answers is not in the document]
ahmed-elsaharti: Global variable and struct have been removed 👍

AirLib/include/api/RpcLibAdapatorsBase.hpp
```diff
@@ -36,6 +36,62 @@ class RpcLibAdapatorsBase {
...
+        RotorSTTParameters rotor[4];
```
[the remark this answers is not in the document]
ahmed-elsaharti: Thats a very valid point! 
To get this to work without memory access violations I had to clear the vector (might not be necessary but I did it as a precaution) and fill it with the rotor params in both [MultirotorPawnSimApi.cpp](https://github.com/microsoft/AirSim/blob/83e3a6dc1901deb5f10cb22e326316039dabd527/Unreal/Plugins/AirSim/Source/Vehicles/Multirotor/MultirotorPawnSimApi.cpp#L75-L79) and [RpcLibAdapatorsBase.hpp](https://github.com/microsoft/AirSim/blob/83e3a6dc1901deb5f10cb22e326316039dabd527/AirLib/include/api/RpcLibAdapatorsBase.hpp#L77-L80). I'm pretty sure there's a more elegant way of doing this though. Let me know if this needs changing

AirLib/include/api/RpcLibAdapatorsBase.hpp
ahmed-elsaharti: Yes it does. Named it this way just for me to remember the purpose of the struct later. This was fixed in 3ee3cad anyway

ahmed-elsaharti:
Hi @zimmy87, Thank YOU for the EXTREMELY helpful comments!
I fixed all the naming issues, got rid of the global variable and generalized the size of 'rotors'.
Please let me know if any other modifications are needed.

zimmy87:
Hi @ahmed-elsaharti, 2 of the checks are failing for this PR due to a unrelated break in the Unity build from another PR. The fix for this build break was committed earlier today, can you rebase this PR to pick up the fix and rerun all the checks?

ahmed-elsaharti:
@zimmy87 Yup, that PR fixed the Unity build and the checks have passed 👍

AirLib/include/api/RpcLibAdapatorsBase.hpp
```diff
@@ -36,6 +36,64 @@ class RpcLibAdapatorsBase {
...
+            {
+                rotor.push_back(RotorParameters());
+                rotor[i] = s.rotor[i];
+            }
+        }
+
+        msr::airlib::RotorVector to() const
+        {
+            msr::airlib::RotorVector d;
+            for (unsigned int i = 0; i < d.rotor.size(); i++)
```
zimmy87, line 88: shouldn't this for loop look like this:
```
for (unsigned int i = 0; i < rotor.size(); i++)
```
I think as is, d.rotor.size() will always evaluate to 0. Also you need to push a new RotorParameters() on to d.rotor inside the for loop.

zimmy87 (changes requested):
This PR looks good for the most part, I just have 1 remaining comment.

AirLib/include/api/RpcLibAdapatorsBase.hpp
ahmed-elsaharti: That actually makes complete sense.
Let me know if the latest commit does what you had in mind.

AirLib/include/vehicles/multirotor/api/MultirotorApiBase.hpp
```diff
@@ -143,6 +148,12 @@ class MultirotorApiBase : public VehicleApiBase {
         token_.cancel();
     }
 
+    /******************* rotors' states setter ********************/
+    void setRotorStates(RotorStates rotor_states)
```
rajat2004, line 152: Pass by ref here?

AirLib/include/api/RpcLibAdapatorsBase.hpp
```diff
@@ -36,6 +36,65 @@ class RpcLibAdapatorsBase {
...
+        {}
+
+        RotorVector(const msr::airlib::RotorVector& s)
+        {
+            rotor.clear();
+            for (unsigned int i = 0; i < s.rotor.size(); i++)
+            {
+                rotor.push_back(RotorParameters());
+                rotor[i] = s.rotor[i];
+            }
```
rajat2004, line 82: This can be replaced by something like (variable names can be improved ig) - 
```
for(const auto& r : s.rotor)
    rotor.push_back(RotorParameters(r));
```
Looks cleaner and avoids copy constructors

AirLib/include/api/RpcLibAdapatorsBase.hpp
```diff
@@ -36,6 +36,65 @@ class RpcLibAdapatorsBase {
...
+        msr::airlib::RotorVector to() const
+        {
+            msr::airlib::RotorVector d;
+            for (unsigned int i = 0; i < rotor.size(); i++)
+            {
+                d.rotor.push_back(msr::airlib::RotorParameters());
+                d.rotor[i].speed = rotor[i].speed;
+                d.rotor[i].thrust = rotor[i].thrust;
+                d.rotor[i].torque_scaler = rotor[i].torque_scaler;
+            }
```
rajat2004, line 94: Same, something like -
```
for(const auto& r : rotor)
    d.rotor.push_back(r.to());
```
Copying each value individually will just cause more pain if the struct needs to be modified later on

Unreal/Plugins/AirSim/Source/Vehicles/Multirotor/MultirotorPawnSimApi.cpp
```diff
@@ -68,8 +68,14 @@ void MultirotorPawnSimApi::updateRenderedState(float dt)
...
 
     //update rotor poses
+    rotor_states_.rotors.rotor.clear();
     for (unsigned int i = 0; i < rotor_count_; ++i) {
         const auto& rotor_output = multirotor_physics_body_->getRotorOutput(i);
+        // update private rotor variable
+        rotor_states_.rotors.rotor.push_back(RotorParameters());
+        rotor_states_.rotors.rotor[i].speed = rotor_output.speed;
+        rotor_states_.rotors.rotor[i].thrust = rotor_output.thrust;
+        rotor_states_.rotors.rotor[i].torque_scaler = rotor_output.torque_scaler;
```
rajat2004, line 78: Adding the parameterized constructor will allow cleaning up this part as well

AirLib/include/common/CommonStructs.hpp
```diff
@@ -8,6 +8,21 @@
...
+struct RotorParameters {
+    real_T thrust = 0;
+    real_T torque_scaler = 0;
+    real_T speed = 0;
+
+    RotorParameters()
+    {}
+};
+struct RotorVector {
+    std::vector<RotorParameters> rotor;
```
rajat2004, line 21: Instead of just `rotor`, maybe something like `rotorv`? This line seemed pretty confusing in `MultirotorPawnSimApi` - `rotor_states_.rotors.rotor[i].speed`. TBH, I don't really see the point of having a separate `RotorVector` struct, but that's just my opinion and depends on the maintainers

PythonClient/airsim/types.py
```diff
@@ -359,6 +361,10 @@ class MultirotorState(MsgpackMixin):
     ready_message = ""
     can_arm = False
 
+class RotorStates(MsgpackMixin):
+    timestamp = np.uint64(0)
+    rotors = RotorVector()
```
rajat2004, line 366: This could be moved to near the `RotorVector` def

Unreal/Plugins/AirSim/Source/Vehicles/Multirotor/MultirotorPawnSimApi.cpp
```diff
@@ -68,8 +68,14 @@ void MultirotorPawnSimApi::updateRenderedState(float dt)
     collision_response = multirotor_physics_body_->getCollisionResponseInfo();
 
     //update rotor poses
+    rotor_states_.rotors.rotor.clear();
     for (unsigned int i = 0; i < rotor_count_; ++i) {
         const auto& rotor_output = multirotor_physics_body_->getRotorOutput(i);
+        // update private rotor variable
+        rotor_states_.rotors.rotor.push_back(RotorParameters());
```
rajat2004, line 75: Instead of continuously clearing and pushing back new objects, can't there instead be like a `set` method or something which would just update the values

PythonClient/airsim/types.py
ahmed-elsaharti: Mainly placed right below Multirotor state while RotorVector is placed at line 220 to stay consistent with (what I think is) the format of the types.py file. Where classes at the top contain classes like quaternion, etc, etc and the ones at the bottom contain classes with variables of the type of these top classes.

AirLib/include/common/CommonStructs.hpp
ahmed-elsaharti: Honestly although it looks odd, I did it with rotors.rotor[i] just to keep everything separate as is the case with most of AirSim's code. Just so that if the code needs to be changed or is looked at later by someone else they'd know that its a parent struct containing vectors of rotor parameters.
This also would help make things easier to understand/keep-track-of on the end user's side imo.

Unreal/Plugins/AirSim/Source/Vehicles/Multirotor/MultirotorPawnSimApi.cpp
ahmed-elsaharti: Similar 'big picture' concept as [this](https://github.com/microsoft/AirSim/pull/3242#discussion_r570335117), although this takes up 4 lines and could probably be cleaner, I think keeping it simple makes this easy to look at and understand later for those (like me) who haven't dived deeply into AirSim's structure.

zimmy87: While making the setting of these values more explicit can make it easier to understand for someone who is just looking at this file, it also makes the code slightly harder to maintain. For example, if a new field was added to RotorParameters and we wanted to keep it updated along with all the other fields, we would also have to know to change this file to ensure the update happens and we wouldn't necessarily catch that until we notice a bug at runtime. Whereas when updating a parameterized constructor for a new field, leaving this file as is will cause a compile error, forcing the developer to keep this file updated to match the RotorParameters definition. Additionally, I think most contributors will either recognize what the parameterized constructor is doing, or they have access to an IDE with auto-lookup so it would be rather straight-forward to lookup the definition of the constructor if they are confused. In this case, I would prefer to take the slight hit to readability as a trade-off for extra maintainability.

AirLib/include/api/RpcLibAdapatorsBase.hpp
zimmy87: if RotorVector is removed, I still think this is a good change to make in msr::airlib_rpclib::MultirotorRpcLibAdapators::RotorStates

AirLib/include/api/RpcLibAdapatorsBase.hpp
zimmy87: same as above, I think this will end up applying to msr::airlib_rpclib::MultirotorRpcLibAdapators::RotorStates

AirLib/src/vehicles/multirotor/api/MultirotorApiBase.cpp
```diff
@@ -849,6 +849,5 @@ float MultirotorApiBase::getObsAvoidanceVelocity(float risk_dist, float max_obs_
     unused(risk_dist);
     return max_obs_avoidance_vel;
 }
-
```
zimmy87, line 852: Is there any particular reason for removing the whitespace here? Seems like this might be left-over from a previous change.

zimmy87 (changes requested):
I agree with most of @rajat2004's comments and think it would be prudent to include them before merging

AirLib/include/common/CommonStructs.hpp
ahmed-elsaharti: Got rid of RotorVector 👍

PythonClient/airsim/types.py
ahmed-elsaharti: Got rid of RotorVector 👍

AirLib/include/api/RpcLibAdapatorsBase.hpp
ahmed-elsaharti: Removed RotorVector and applied these changes to msr::airlib_rpclib::MultirotorRpcLibAdapators::RotorStates

AirLib/include/api/RpcLibAdapatorsBase.hpp
ahmed-elsaharti: Removed RotorVector and applied these changes to msr::airlib_rpclib::MultirotorRpcLibAdapators::RotorStates

AirLib/src/vehicles/multirotor/api/MultirotorApiBase.cpp
ahmed-elsaharti: Fixed

AirLib/include/vehicles/multirotor/api/MultirotorApiBase.hpp
ahmed-elsaharti: Fixed

AirLib/include/common/CommonStructs.hpp
```diff
@@ -8,6 +8,21 @@
...
 
 namespace msr { namespace airlib {
+//rotor feedback
+struct RotorParameters {
+    real_T thrust = 0;
+    real_T torque_scaler = 0;
+    real_T speed = 0;
+
+    RotorParameters()
+    {}
```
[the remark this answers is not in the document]
ahmed-elsaharti: Valid point. An initializer list has been added to RotorParameters

Unreal/Plugins/AirSim/Source/Vehicles/Multirotor/MultirotorPawnSimApi.cpp
ahmed-elsaharti: Added an initializer list to RotorParameters and replaced these with [the if condition below](https://github.com/microsoft/AirSim/pull/3242#discussion_r580833347) and some description of what's happening

Unreal/Plugins/AirSim/Source/Vehicles/Multirotor/MultirotorPawnSimApi.cpp
ahmed-elsaharti: Fixed

ahmed-elsaharti:
@rajat2004 @zimmy87 Thank you guys for the useful comments!
I can't for the life of me remember why I added RotorVector initially but I do agree, it is 100% redundant at this point.
I adjusted everything to:
image
rather than:
image
and added an initializer list to `RotorParameters` to keep things clean when populating/updating the rotor states.
One thing I know for sure is that I should stop looking at PR reviews after midnight. Everything made much more sense when I re-read the comments a few days later 😅😅

Anyway, the PR should be good to go now that the Github checks have run successfully. Let me know if any other modifications are needed.

Unreal/Plugins/AirSim/Source/Vehicles/Multirotor/MultirotorPawnSimApi.cpp
```diff
@@ -70,6 +70,13 @@ void MultirotorPawnSimApi::updateRenderedState(float dt)
     //update rotor poses
     for (unsigned int i = 0; i < rotor_count_; ++i) {
         const auto& rotor_output = multirotor_physics_body_->getRotorOutput(i);
+        // update private rotor variable
+        if (rotor_states_.rotors.size() < i + 1)
+            //add a new element to the vector if it doesnt contain a rotor/element of index 'i' yet
+            rotor_states_.rotors.push_back(RotorParameters(rotor_output.speed, rotor_output.thrust, rotor_output.torque_scaler));
+        else
+            //if the vector has already been populated previously, just update the values
+            rotor_states_.rotors[i] = RotorParameters(rotor_output.speed, rotor_output.thrust, rotor_output.torque_scaler);
```
rajat2004, line 79: This is still creating and replacing a new object, a more efficient way could be something like `rotor_states_.rotors[i].update(rotor_output.speed, rotor_output.thrust, rotor_output.torque_scaler)`, where `update` or `setValues` would just update the values.

Just my 2 cents, its up to the PR author and the maintainers as to which is better

PythonClient/airsim/client.py
```diff
@@ -1282,7 +1282,20 @@ def getMultirotorState(self, vehicle_name = ''):
...
+    def getRotorStates(self, vehicle_name = ''):
+        """
+        Used to obtain the current state of all a multirotor's rotors. The state includes the speeds,
+        thrusts and torques for all rotors.
+
+        Args:
+            vehicle_name (str, optional): Vehicle to get the rotor state of
 
+        Returns:
+            RotorStates: Containing a timestamp and the speed, trust and torque of all rotors.
```
rajat2004, line 1295: Spelling - `thrust`

rajat2004 (commented):
Just some small nitpicks

AirLib/include/vehicles/multirotor/api/MultirotorApiBase.hpp
```diff
@@ -143,6 +148,12 @@ class MultirotorApiBase : public VehicleApiBase {
         token_.cancel();
     }
 
+    /******************* rotors' states setter ********************/
+    void setRotorStates(RotorStates& rotor_states)
```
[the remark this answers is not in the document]
ahmed-elsaharti: Fixed

AirLib/include/vehicles/multirotor/api/MultirotorRpcLibAdapators.hpp
```diff
@@ -39,6 +39,35 @@ class MultirotorRpcLibAdapators : public RpcLibAdapatorsBase {
...
+            }
+            timestamp = s.timestamp;
+        }
+
+        msr::airlib::RotorStates to() const
+        {
+            std::vector<msr::airlib::RotorParameters> d;
+            for (const auto& r : rotors)
+            {
+                d.push_back(msr::airlib::RotorParameters(r.to()));
```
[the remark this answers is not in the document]
ahmed-elsaharti: Fixed

AirLib/include/common/CommonStructs.hpp
```diff
@@ -8,6 +8,20 @@
 #include <ostream>
 
 namespace msr { namespace airlib {
+//rotor feedback
+struct RotorParameters {
```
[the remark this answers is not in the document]
ahmed-elsaharti: Good point. Fixed

AirLib/include/api/RpcLibAdapatorsBase.hpp
```diff
@@ -36,6 +36,29 @@ class RpcLibAdapatorsBase {
             d.push_back(TDest(s.at(i)));
     }
 
+    struct RotorParameters {
```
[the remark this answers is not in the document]
ahmed-elsaharti: Fixed

Unreal/Plugins/AirSim/Source/Vehicles/Multirotor/MultirotorPawnSimApi.cpp
ahmed-elsaharti: I'm honestly fine with both ways. 
@zimmy87 what do you think? as-is or should I add an `update` method to `RotorParameters`.

I've also realized that I made a mistake here with the order of the parameters. That has been fixed in the latest commit

PythonClient/airsim/client.py
ahmed-elsaharti: Good catch!
Fixed

Unreal/Plugins/AirSim/Source/Vehicles/Multirotor/MultirotorPawnSimApi.cpp
ahmed-elsaharti: On the topic of efficiency, I was thinking of having a for loop populate `rotor_states_.rotors` in `MultirotorPawnSimApi::initialize` once instead of having to go through an if condition (the condition for-which would always be false after the first run) with every call of `MultirotorPawnSimApi::updateRenderedState`. Any thoughts?

rajat2004: Yeah, that should be a good way, similar to [this](https://github.com/microsoft/AirSim/blob/0a3f84c92e195f1d4b217afb5dcedfea09bff454/Unreal/Plugins/AirSim/Source/Vehicles/Multirotor/MultirotorPawnSimApi.cpp#L27) should work

ahmed-elsaharti: Oh yeah, using the `assign` modifier is a better idea 👍

zimmy87: I like the idea of using the assign function in `MultirotorPawnSimApi::initialize` to instantiate all the structs and then calling an `update` method defined in `RotorParameters` to set the values in each struct here

zimmy87 (approved):
This PR looks good to me and once the one remaining feedback concerning populating the rotor states is resolved, I think it will be ready for merging

Unreal/Plugins/AirSim/Source/Vehicles/Multirotor/MultirotorPawnSimApi.cpp
ahmed-elsaharti: Done 👍 
Let me know if anything else needs changing

ahmed-elsaharti:
@zimmy87 the last commit (4d8b3c9) should take care of [these changes](https://github.com/microsoft/AirSim/pull/3242#discussion_r585914035). Let me know if any other changes are needed 👍

zimmy87:
Thank you for the contribution @ahmed-elsaharti!

condeLucas:
Hi @ahmed-elsaharti

Thank you for your contribution. 

Using @sytelus 's method to have the energy/battery consumption, we have only to multiply **Thrust** * **torque_scaler** * **speed** for each rotor and sum then during the simulation time?

Do you know the units of these measurements?

Thank you so much for your attention.

PR #3247, opened 2020-12-25
Docs: Add plugin for tip, note blocks

rajat2004:
Fixes: 
- Enables plugin for note, tip blocks in Mkdocs
- Adds tip, note blocks in a few files   

## About
Note, Tip blocks are common in rst files which are converted to HTML by Sphinx, these are more noticable than normal text. Adding this plugin allows similar usage in Mkdocs. Example -

image

Came to my mind after seeing https://github.com/microsoft/AirSim/discussions/3245

Only added in a few files right now

## How Has This Been Tested?

`mkdocs serve` from the root directory

image


rajat2004:
@jonyMarino @zimmy87 FYI, the Azure Windows pipeline seems to be broken again

jonyMarino:
Thanks, Rajat! I didn't know how to add that. You read my mind!

PR #3250, opened 2020-12-28
Adding Sensor Simulation (GPS + IMU)  and EKF support

rohitmohan7:
Fixes: #    

### DIFF
--- a/AirLib/include/sensors/gps/GpsBase.hpp
+++ b/AirLib/include/sensors/gps/GpsBase.hpp
@@ -7,7 +7,7 @@
 
 #include "sensors/SensorBase.hpp"
 #include "common/CommonStructs.hpp"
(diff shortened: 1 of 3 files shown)


AirLib/include/sensors/gps/GpsBase.hpp
```diff
@@ -123,6 +123,7 @@ class GpsBase  : public SensorBase {
 
 private: 
     Output output_;
+    std::shared_ptr<Ekf> _ekf;
```
jonyMarino, line 126: Can you put the ekf outside the sensor? I don't want this PR to modify existing sensors and controllers

jonyMarino:
I think this will create a new controller that would be selected through the settings. Am I correct?

Issue #3258, opened 2020-12-29
Segmentation images not appearing correctly from Blocks environment


uzgit:
### What's your question?

Why do the Segmentation images that I am extracting from the Blocks sample environment appear just as normal (darker) RGB images? I am very new to AirSim/Unreal so I kind of assume I am doing something naive or overlooking something simple.

### Include context on what you are trying to achieve

I am trying to collect Scene and Segmentation images from an edited version of the sample environment Blocks. The scene pictures are extracted as expected, but the Segmentation images appear just as slightly different RGB images. I am trying to extract Segmentation images where each mesh/object is given a single, solid color in order to create a mask so that I can train a model to detect certain objects within the scene.

#### Context details

OS: Ubuntu 20.04
Python: 3.8.5
SITL: ArduCopter (sim_vehicle.py)

### Include details of what you already did to find answers
I am using the following Python script to extract images from AirSim:
```
#!/usr/bin/python3

import sys
sys.path.insert(0, "/home/joshua/git/Airsim/PythonClient") # use the included AirSim Python client
import airsim
import numpy

client = airsim.MultirotorClient()

# try to exclude all objects from the segmentation image as per the documentation
result = client.simSetSegmentationObjectID("[\w]*", -1, True)
print("set all object ids to -1: {}".format(result))

# set the ID of the symbol cube to 99 instead of -1 to include it in the segmentation image
result = client.simSetSegmentationObjectID("symbol_cube", 99)
symbol_cube_id = client.simGetSegmentationObjectID("symbol_cube")
print("set symbol_cube id: {} -> {}".format(result, symbol_cube_id))

# verify IDs for all objects are -1 unless that object is the symbol cube
scene_objects = client.simListSceneObjects("[\w]*")
for scene_object in scene_objects:
    scene_object_id = client.simGetSegmentationObjectID(scene_object)
    # print("{} : {}".format(scene_object, scene_object_id))
    if( scene_object != "symbol_cube" ):
        assert scene_object_id == -1

# request Scene and Segmentation images
requests = [airsim.ImageRequest(camera_name=0, image_type=airsim.ImageType.Scene, pixels_as_float=False, compress=False),
            airsim.ImageRequest(camera_name=0, image_type=airsim.ImageType.Segmentation, pixels_as_float=False, compress=False)]
responses = client.simGetImages(requests)
print("response length: {}".format(len(responses)))

# for each returned image
for i in range(len(requests)):

    # make the image as a 1-dimensional numpy array
[...]

PR #3263, opened 2020-12-30
Update UE version, remove duplicated code in install script

rajat2004:
Fixes: 
- Older UE4 version in `install_unreal.sh` script, uodated to 4.25
- Remove duplicated script in `install_run_all.sh`, instead call `install_unreal.sh` to do the installation.

## About

Came to my mind after #3258, wasn't aware before that the script is actualy being used much :sweat_smile: 
Removing duplicated install script makes it easier to update going further. Docs and project update was done in https://github.com/microsoft/AirSim/pull/3227

## How Has This Been Tested?

The script itself hasn't been tested, but wrote 2 small test scripts to demonstrate -

test.sh -
```sh
#!/bin/bash

set -e
set -x

./test2.sh $1

echo "from test"
```

test2.sh -
```sh
#!/bin/bash

echo $1
exit 1
```

### DIFF
--- a/install_run_all.sh
+++ b/install_run_all.sh
@@ -7,31 +7,14 @@ pushd "$SCRIPT_DIR"  >/dev/null
 set -e
 set -x
 
-#confirm unreal install directory
+# Get Unreal install directory
 UnrealDir=$1
 if [[ !(-z "UnrealDir") ]]; then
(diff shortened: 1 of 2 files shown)


jonyMarino:
Thank you Rajat!

PR #3265, opened 2020-12-30
Fix comment of airsim.to_quaternion

alonfaraj:
RPY -> PRY in radians

Fixes: #    

Inaccurate documentation comment


jonyMarino:
Thanks @alonfaraj!

PR #3273, opened 2021-01-05
Only install packages if required on OSX

rajat2004:
Fixes: MacOS Actions CI    

## About

Currently, the MacOS build is giving error - `Error: wget 1.20.3_2 is already installed`. This modifies the script to only install the package if required. Based on [this](https://discourse.brew.sh/t/skip-ignore-brew-install-if-package-is-already-installed/633)

Recommended way is to use [homebrew-bundle](https://github.com/Homebrew/homebrew-bundle), but I haven't ever used it and might be a bit overkill, but let me know if that's the better way to go.

## How Has This Been Tested?

Github Actions CI

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -9,6 +9,9 @@ downloadHighPolySuv=true
 MIN_CMAKE_VERSION=3.10.0
 function version_less_than_equal_to() { test "$(printf '%s\n' "$@" | sort -V | head -n 1)" = "$1"; }
 
+# brew gives error if package is already installed
+function brew_install() { brew list $1 &>/dev/null || brew install $1; }
+
 # Parse command line arguments
 while [[ $# -gt 0 ]]
 do
@@ -63,7 +66,8 @@ if [ "$(uname)" == "Darwin" ]; then # osx
         sudo dseditgroup -o edit -a `whoami` -t user dialout
     fi
 
-    brew install wget coreutils
+    brew_install wget
+    brew_install coreutils
 
     if version_less_than_equal_to $cmake_ver $MIN_CMAKE_VERSION; then
         brew install cmake  # should get cmake 3.8


setup.sh
```diff
@@ -9,6 +9,9 @@ downloadHighPolySuv=true
 MIN_CMAKE_VERSION=3.10.0
 function version_less_than_equal_to() { test "$(printf '%s\n' "$@" | sort -V | head -n 1)" = "$1"; }
 
+# brew gives error if package is already installed
+function brew_install() { brew list $1 &>/dev/null || brew install $1; }
```
zimmy87, line 13: Will this consume errors from brew apart from the packaged already installed error? There might be an alternate way of detecting if a specific package is installed described [here](https://stackoverflow.com/questions/20802320/detect-if-homebrew-package-is-installed)

zimmy87 (commented):
Hi @rajat2004 we have a question that we'd like addressed before approving this PR for merging.

setup.sh
rajat2004: Do you mean using something like `brew ls --versions <formula>`? If yes, then there shouldn't be any difference, according to the [manpage](https://docs.brew.sh/Manpage#list-ls-options-formulacask), `ls` and `list` are same, and `--version` will reduce the output. If that's more preferable I'll make the change, but the end result should remain the same I believe. Also [here's](https://github.com/rajat2004/AirSim/runs/1647721281?check_suite_focus=true#step:3:2160) the output before I put the `&>/dev/null`

If referring to some other method, then could you specify that one? Though others seem to work or not work for some packages

zimmy87: Hey @rajat2004, I looked into the syntax of this bash command more and it appears I misunderstood it's behavior yesterday. I thought the redirecting of the output to /dev/null was consuming any errors thrown by this command, but that's not the case. Consider my question there resolved. And you're right about ls and list being the same, I see now that the code outlined in the stackoverflow link is just a more explicit way of doing the same thing you are here.

PR #3305, opened 2021-01-14
Fixed issue where no new scene is rendered after simContinueForTime

### LINKED ISSUE #3304: Scene doesn't change after calling simContinueForTime with a low value
## Bug report 
- AirSim Version: Latest master
- UE/Unity version: 4.25
- OS Version: Ubuntu 18.04

I am trying to simulate a 90 FPS camera. To achieve this I'm running the simulation for a short period of time and then retrieving an image.
```c++
void update()
{
    // simulate 90 FPS
    client.simContinueForTime(1.0/90);

    // retrieve image
    const std::vector<ImageResponse> &response = client.simGetImages(request);
    
    // process frame
}
```
When comparing the images before and after calling `simContinueForTime`, I noticed that the images are the same.

niekvleeuwen:
Fixes: #3304
When comparing the images before and after calling simContinueForTime, I noticed that the images are the same. This pull request fixes this issue. Please see #3304 for a more detailed explanation.


jonyMarino:
It needs investigation why the MacOS check is failing.

zimmy87:
Tested with pause_continue_drone.py and a custom script for repro'ing #3304; didn't notice a regression with either. MacOS build check failure went away after merging with master so this looks good to me now, moving ahead with merge.

PR #3343, opened 2021-01-25
Fix UnityDemo build

### LINKED ISSUE #3276: Some problem about the Unitydemo/carDemo
https://github.com/microsoft/AirSim/blob/68ae8de7065d2225a1c211f656ec875ee2e622f6/Unity/UnityDemo/Assets/AirSimAssets/Scripts/Vehicles/VehicleCompanion.cs#L197

Hi, I have some question about this function, It seem like we didn't use this function. 
So I try to annotation this part of code.
But when I try to run the  carDemo, the Editor will crash.
From the Log i found

Loaded scene 'Assets/Scenes/CarDemo.unity'
	Deserialize:            27.018 ms
	Integration:            13.212 ms
	Integration of assets:  1.326 ms
	Thread Wait Time:       6.267 ms
	Total Operation Time:   47.824 ms
NullReferenceException: Object reference not set to an instance of an object
  at AirSimUnity.VehicleCompanion.Reset (System.String vehicleName) [0x00025] in C:\AirSim-master\Unity\UnityDemo\Assets\AirSimAssets\Scripts\Vehicles\VehicleCompanion.cs:146 

https://github.com/microsoft/AirSim/blob/68ae8de7065d2225a1c211f656ec875ee2e622f6/Unity/UnityDemo/Assets/AirSimAssets/Scripts/Vehicles/VehicleCompanion.cs#L146

But I didn't found which code use it, except “getfunctionpointerfordelegate”.

Is there anyone else have the same situation like me？

zimmy87:
Fixes: #3276

## About
This change fixes the build break caused by VehicleCompanion.SetDistortionParam referencing a non-existent parameter. This also fixes an issue where the DllImport of InitVehicleManager was missing two parameters that were added to the API in the AirLibWrapper. 

## How Has This Been Tested?
AirLibWrapper was built with this PR applied and it was confirmed that both car and drone mode in the UnityDemo project can launch without issue.


ClarkZha:
This fix solves the build problem in master (on Linux 18.04 with Unity 2019.3.12. )

Unity/UnityDemo/Assets/AirSimAssets/Scripts/Utilities/PInvokeWrapper.cs
```diff
@@ -13,8 +13,9 @@ public static class PInvokeWrapper {
         [DllImport(DLL_NAME)]
         public static extern void InitVehicleManager(IntPtr SetPose, IntPtr GetPose, IntPtr GetCollisionInfo, IntPtr GetRCData,
             IntPtr GetSimImages, IntPtr SetRotorSpeed, IntPtr SetEnableApi, IntPtr SetCarApiControls, IntPtr GetCarState,
-            IntPtr GetCameraInfo, IntPtr SetCameraPose, IntPtr SetCameraFoV, IntPtr SetSegmentationObjectid, IntPtr GetSegmentationObjectId,
-            IntPtr PrintLogMessage, IntPtr GetTransformFromUnity, IntPtr Reset, IntPtr GetVelocity, IntPtr GetRayCastHit, IntPtr Pause);
+            IntPtr GetCameraInfo, IntPtr SetCameraPose, IntPtr SetCameraFoV, IntPtr setDistortionParam, IntPtr getDistortionParams,
+            IntPtr setSegmentationObjectId, IntPtr getSegmentationObjectId, IntPtr PrintLogMessage, IntPtr GetTransformFromUnity,
```
rajat2004, line 17: All the other method names in Unity seem to follow the same convention of uppercase method name for the APIs defined in WorldSimApi, is it possible to do the same here? Otherwise this will always be a spot of confusion later on

zimmy87: Good catch! My latest commit updates this PR to align with the naming conventions of the rest of the Unity project

PR #3351, opened 2021-01-31
Fix compile error for Unity build

niekvleeuwen:
## About
@zimmy87 Follow up on #3102. I noticed that the Unity build of my previous pull request was failing. At the time I commited this feature the Github Actions were not present, so I didn't notice the failing build. This pull request fixes this failing Unity build.

### DIFF
--- a/PythonClient/airsim/client.py
+++ b/PythonClient/airsim/client.py
@@ -116,7 +116,7 @@ class VehicleClient:
         Continue (or resume if paused) the simulation for the specified number of frames, after which the simulation will be paused.
 
         Args:
-            frames (float): Frames to run the simulation for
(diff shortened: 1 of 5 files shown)


rajat2004:
@niekvleeuwen Another thing I noticed, the [docstring](https://github.com/microsoft/AirSim/blob/master/PythonClient/airsim/client.py#L119) mentions the type as `float`, but it should be an integer

niekvleeuwen:
@rajat2004 Thanks for noticing! I updated my branch to fix the incorrect docstring.

zimmy87:
thanks for catching and fixing this @niekvleeuwen!

PR #3401, opened 2021-02-18
Multirotor mass limit clarification & broken link fix

### LINKED ISSUE #3400: Colour pallet link is broken in Docs
The the pallet link can be found in 
https://microsoft.github.io/AirSim/image_apis/#changing-colors-for-object-ids

ahmed-elsaharti:
Fixes: #3400

## About
This PR adds some clarification about where the mass bounds in [MultiRotorParams.hpp](https://github.com/microsoft/AirSim/blob/288c90ca78ce09ed58de8508622cb97071c8f217/AirLib/include/vehicles/multirotor/MultiRotorParams.hpp#L226) come from. 
Where the max limit is actually the mass equivalent to the maximum lift force the rotors+propellers would be able to generate.
image
image
(1.6kg if using the parameters in [RotorParams.hpp](https://github.com/microsoft/AirSim/blob/288c90ca78ce09ed58de8508622cb97071c8f217/AirLib/include/vehicles/multirotor/RotorParams.hpp#L45) on a quadrotor)
and the min limit is the minimum weight equivalent to 50%of the maximum lift power, aka: the lift power at 50% throttle which is the idling throttle (on SimpleFlight). Anything lower than this will lift off as soon as the drone is armed (ie: with the throttle stick untouched)
(0.8kg  if using the parameters in [RotorParams.hpp](https://github.com/microsoft/AirSim/blob/288c90ca78ce09ed58de8508622cb97071c8f217/AirLib/include/vehicles/multirotor/RotorParams.hpp#L45) on a quadrotor)

This PR also fixed a broken link in [the docs](https://microsoft.github.io/AirSim/image_apis/#changing-colors-for-object-ids)

### DIFF
--- a/AirLib/include/vehicles/multirotor/MultiRotorParams.hpp
+++ b/AirLib/include/vehicles/multirotor/MultiRotorParams.hpp
@@ -223,9 +223,10 @@ protected: //static utility functions for derived classes to use
         std::vector<real_T> arm_lengths(params.rotor_count, 0.2275f);
 
         //set up mass
-        //this has to be between max_thrust*rotor_count/10 (1.6kg using default parameters in RotorParams.hpp) and 0.5*max_thrust*rotor_count/10 (0.8kg using default parameters)
+        //this has to be between max_thrust*rotor_count/10 (1.6kg using default parameters in RotorParams.hpp) and (idle throttle percentage)*max_thrust*rotor_count/10 (0.8kg using default parameters and SimpleFlight)
         //any value above the maximum would result in the motors not being able to lift the body even at max thrust,
-        //and any value below the minimum would cause the drone to fly upwards on idling throttle (50% of the max throttle)
+        //and any value below the minimum would cause the drone to fly upwards on idling throttle (50% of the max throttle if using SimpleFlight)
+        //Note that the default idle throttle percentage is 50% if you are using SimpleFlight
         params.mass = 1.0f; 
 
         real_T motor_assembly_weight = 0.055f;  //weight for MT2212 motor for F450 frame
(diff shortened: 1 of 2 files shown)


AirLib/include/vehicles/multirotor/MultiRotorParams.hpp
```diff
@@ -250,7 +254,11 @@ class MultiRotorParams {
         std::vector<real_T> arm_lengths(params.rotor_count, 0.2275f);
 
         //set up mass
-        params.mass = 1.0f; //can be varied from 0.800 to 1.600
+        //this has to be between max_thrust*rotor_count/10 (2.5kg using default parameters in RotorParams.hpp) and 0.5*max_thrust*rotor_count/10 (1.25kg using default parameters)
+        //any value above the maximum would result in the motors not being able to lift the body even at max thrust,
+        //and any value below the minimum would cause the drone to fly upwards on idling throttle (50% of the max throttle)
+        params.mass = 1.0f;
```
rajat2004, line 260: Isn't this below the 1.25 kg for 50% thrust as calculated, should this be updated as well?

ahmed-elsaharti: Actually, since (AFAIK) simple flight does not have a hexacopter mode, the only way this would be used would be through PX4 (or maybe Ardupilot, etc if they support hexacopters) where the idle throttle is set to 20% by default. So 1kg would be within the correct range.

Nonetheless, I'll just replace '50%' with 'idle throttle percentage' to make the description more generalized

jonyMarino:
Thanks, Ahmed!

EdwinMeriaux:
@ahmed-elsaharti 

Hello!
So I understand that the idle is set to 50% of thrust. But where is the 50% set?
Is there any way to change this to say 45%?

ahmed-elsaharti:
Hi @EdwinMeriaux ,
If I remember correctly, the 50% idle thrust (I'm assuming we're talking about the simpleflight flight controller here) in my case was set by the gamepad/joystick connected to the computer via USB. ie: since i was using a USB gamepad with thumbsticks that default to the center, it pushed out a signal of 50% (ie: center). I remember the idle throttle dropping below 50% once no joysticks were connected to the computer.

Idr exactly where the non-joystick idle throttle was set but I'd assume it would be somewhere in the simpleflight directory since that parameter should be controlled by the autopilot code

PR #3421, opened 2021-02-24
Fix missing `import os` in apis.md

evroon:
## About
Fixes the missing `os` import in `apis.md`.


jonyMarino:
Thanks, @evroon for contributing!

PR #3453, opened 2021-03-08
Added Lidar PointCloud Python Example Scripts

euboon1601:
Fixes: #    

## About

This pull request includes some python example scripts created to scan 3D objects/terrain with the built-in AirSim LiDAR module. It can be used for a multiple lidar sensors onboard a single vehicle of any type. There are two main variations of the script: one which is written for the sensors which are configured with 'SensorLocalFrame' as the DataFrame, and the other for 'VehicleInertialFrame' as the DataFrame. they can be differentiated based on the python scripts' naming scheme.
## How Has This Been Tested?
 
These scripts are meant to run parallel to any movement scripts created for vehicle displacements and does not move the vehicle by itself, then saves the terrain scanned as a point cloud in '.asc' format.

For the unmodified scripts, they are written for the presence of two LiDAR modules on a single vehicle in AirSim (See first image for configuration). The line 'lidarTest.execute('Drone1',['LidarSensor1','LidarSensor2'])' accepts two arguments, where in this case, 'Drone1' is the name for the relevant vehicle, and the second argument is an array that contains the LiDAR sensors' names from which you wish to extract data from.

Upon completion of running these scripts, the point clouds of each sensor will be stored as '{Vehicle Name}_{Sensor Name}_pointcloud.asc'.

In the presence of multiple vehicles, just run multiple of these scripts, one for each vehicle, and change the first argument under 'lidarTest.execute()' to the relevant vehicle's name. In the case where there is a different number of sensors you wish to read from, alter the array which is the second argument of 'lidarTest.execute()' to include all the names of the sensors you wish to read  from.

For more info, I will include my GitHub Page with a more detailed README file that explains how to use them: 
https://github.com/euboon1601/AirSimLiDARpointcloud

image
Capture 1
Capture 2


PythonClient/multirotor/drone_lidar_pointcloud.py
```diff
@@ -0,0 +1,84 @@
...
+        self.client.armDisarm(False)
+        self.client.reset()
+
+        self.client.enableApiControl(False)
+        print("Done!\n")
+
+# main
+if __name__ == "__main__":
+    args = sys.argv
+    args.pop(0)
```
rajat2004, line 73: Why is this needed?

PythonClient/multirotor/drone_lidar_pointcloud.py
```diff
@@ -0,0 +1,84 @@
...
+        self.client.confirmConnection()
+        self.client.enableApiControl(True)
+
+    def execute(self):
+        #self.client.reset()
+        print('Scanning Has Started')
+        f = open('data.asc', "w")
+        try:
+            while True:
+                lidarData = self.client.getLidarData();
```
rajat2004, line 30: Maybe a global variable such as `LIDAR_NAME` to easily change the lidar to get data from. It could be an arg to the method also, like different files for different lidars

rajat2004 (commented):
Just some small comments. Apart from this, I have some doubts

- The rotation matrix and all being done to convert from Local NED reference frame to global correct, when using the lidar in `SensorLocalFrame`. Does using the `VehicleInertialFrame` work without requiring all this? That part of the code can be improved, but haven't spent much time on reviewing it for this reason.

- 2 different scripts are possibly not required, especially since the script isn't meant to control and move the vehicle, just collect and store the lidar data. Making the change to `VehicleClient` and removing the arm, disarm stuff should make it vehicle agnostic
- Please add some example settings in the beginning of the script, this will help when using it, especially is a specific frame is used
- It might be good add a few lines in the Lidar doc page as well, along with any images if you prefer

euboon1601:
Hi, thanks for the comments, I really appreciate your input on all this, it has been very helpful, and I'll try to implement what you've suggested as best as I can.

> * The rotation matrix and all being done to convert from Local NED reference frame to global correct, when using the lidar in `SensorLocalFrame`. Does using the `VehicleInertialFrame` work without requiring all this? That part of the code can be improved, but haven't spent much time on reviewing it for this reason.

You're absolutely right on this, I'd initially tested this script using `SensorLocalFrame` as it is more in line with the current project that I'm working on. With that in mind, would you say it's a good idea if I include 2 scripts, one optimized for using the `SensorLocalFrame` frame, and another for `VehicleInertialFrame` frame, and specify their use-case through comments at the top of the script? Or would you say it's better to just have the `VehicleInertialFrame` frame script?
> * 2 different scripts are possibly not required, especially since the script isn't meant to control and move the vehicle, just collect and store the lidar data. Making the change to `VehicleClient` and removing the arm, disarm stuff should make it vehicle agnostic

Again, you're correct on this too, I'll make the necessary changes. But since that's the case, I worry whether I should still include the script under the multirotor or car folders, as I kind of feel it might not be very relevant to those folders anymore.
> * Please add some example settings in the beginning of the script, this will help when using it, especially is a specific frame is used
> * It might be good add a few lines in the Lidar doc page as well, along with any images if you prefer

These are very good suggestions, I'll definitely try to sort them out.

I really appreciate your time in reviewing this request, and I truly apologize for any shortcomings, as this is all very new to me, but I had hoped to contribute some of the my findings that I think would be useful for anyone else that may be interested in it.

PythonClient/multirotor/drone_lidar_pointcloud.py
```diff
@@ -0,0 +1,84 @@
...
+        print("Done!\n")
+
+# main
+if __name__ == "__main__":
+    args = sys.argv
+    args.pop(0)
+
+    arg_parser = argparse.ArgumentParser("Lidar.py makes drone fly and gets Lidar data")
+
+    arg_parser.add_argument('-save-to-disk', type=bool, help="save Lidar data to disk", default=False)
```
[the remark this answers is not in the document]
euboon1601: I'll remove this in a later revision of the script, thanks for pointing it out

PythonClient/multirotor/drone_lidar_pointcloud.py
euboon1601: I'll remove this too, thanks for letting me know of it

rajat2004:
It's a great PR, reviewing the code helps in figuring out the possible bugs and improvements which can be made. And code reviews have been a big factor in improving my own skills, and doing this for other PRs helps me as well as the contributors. And welcome to the world of open-source!

> You're absolutely right on this, I'd initially tested this script using `SensorLocalFrame` as it is more in line with the current project that I'm working on. With that in mind, would you say it's a good idea if I include 2 scripts, one optimized for using the `SensorLocalFrame` frame, and another for `VehicleInertialFrame` frame, and specify their use-case through comments at the top of the script? Or would you say it's better to just have the `VehicleInertialFrame` frame script?

Sure, having 2 different implementations should be good in any case, can be used depending on the requirements. For eg, if the vehicle is using it for some collision avoidance as well during flying, then it's likely that `SensorLocalFrame` is used.

> Again, you're correct on this too, I'll make the necessary changes. But since that's the case, I worry whether I should still include the script under the multirotor or car folders, as I kind of feel it might not be very relevant to those folders anymore.

Should be okay keeping it under the multirotor folder, it can be mentioned in the docs and the script that it can be used for any vehicle, so shouldn't be a problem.

PythonClient/multirotor/drone_lidar_pointcloud.py
rajat2004: See the [getLidarData()](https://microsoft.github.io/AirSim/api_docs/html/#airsim.client.VehicleClient.getLidarData) API, it takes args `lidar_name` and `vehicle_name` which can be used to select the sensor in case of multiple sensors and vehicles. The sensor and vehicle name corresponds to the name set in the settings

euboon1601:
Updates since last PR review:

- Replaced individual Car/Multirotor scripts with Sensor Frame/Vehicle Frame scripts that work for all AirSim supported Vehicles.
- Removed irrelevant portions of the code.
- Added support for multiple LiDAR sensors on-board a single vehicle (which have to be declared as elements of an array when running the .execute function). Scanned data for each sensor will also be saved under their respective .asc files.
- Support for multiple vehicles using a single script run still yet to be implemented, but for now running multiple scripts, one for each vehicle should still work.
- Added some more documentation for hopefully better clarification.

I hope all the changes that have been made are able to address the previously mentioned issues with this pull request, and that this may better serve anyone who is interested about the contents provided here. Do let me know if there is anything that should be improved upon, any feedback is greatly appreciated.

PythonClient/multirotor/sensorframe_lidar_pointcloud.py
```diff
@@ -0,0 +1,73 @@
...
+class LidarTest:
+
+    def __init__(self):
+
+        # connect to the AirSim simulator
+        self.client = airsim.VehicleClient()
+        self.client.confirmConnection()
+        print('Connected!\n')
+
+    def execute(self,vehicleName,lidarNames):
```
rajat2004, line 24: Airsim follows the convention of using snake case for variables, see https://microsoft.github.io/AirSim/coding_guidelines/#naming-conventions
To be consistent with the codebase, best to rename it as `vehicle_name`, `lidar_names`

PythonClient/multirotor/sensorframe_lidar_pointcloud.py
```diff
@@ -0,0 +1,73 @@
...
+        print('Scanning Has Started\n')
+        print('Use Keyboard Interrupt \'CTRL + C\' to Stop Scanning\n')
+        flag1=0
+        try:
+            while True:
+                for lidars in lidarNames:
+                    if flag1 < len(lidarNames):
+                        f = open(vehicleName+'_'+lidars+'_pointcloud.asc','w')
+                    else:
+                        f = open(vehicleName+'_'+lidars+'_pointcloud.asc','a')
```
rajat2004, line 35: Descriptive variable names are preferred, rather than `flag1`. `lidars` is also slightly wrong, better would be `lidar` or `lidar_name`.
From what I understand, this piece of code is used to determine whether to erase the already existing data in the files first by opening in 'write' mode and afterwards in 'append' mode.
A clearer way could be like -

```python
existing_data_cleared = false

try:
  while True:
    for lidar_name in lidar_names:
      filename = f"{vehicle_name}_{lidar_name}_pointcloud.asc"
        if not existing_data_cleared:
            f = open(filename,  'w')
        else:
            f = open(filename, 'a')
            ...

    # After for-loop
    existing_data_cleared = true
```

There might be better ways also, such as storing this in a temp folder. In the beginning, check if the folder already exists and if so, clear it out and then just write in append mode directly. 

The f-string is just my preference, up to you how you want to write. Ignore the inconsistent indentation, Github doesn't allow tabs, just using spaces continuously gets frustrating

PythonClient/multirotor/sensorframe_lidar_pointcloud.py
```diff
@@ -0,0 +1,73 @@
...
+                    lidarData = self.client.getLidarData(lidar_name=lidars,vehicle_name=vehicleName);
+                    [q0,q1,q2,q3] = [lidarData.pose.orientation.w_val,lidarData.pose.orientation.x_val,lidarData.pose.orientation.y_val,lidarData.pose.orientation.z_val]
+                    rotationMatrix = np.array(([1-2*(q2*q2+q3*q3),2*(q1*q2-q3*q0),2*(q1*q3+q2*q0)],
+                                                  [2*(q1*q2+q3*q0),1-2*(q1*q1+q3*q3),2*(q2*q3-q1*q0)],
+                                                  [2*(q1*q3-q2*q0),2*(q2*q3+q1*q0),1-2*(q1*q1+q2*q2)]))
```
rajat2004, line 40: Could you add a reference as to the source of this matrix, I saw https://automaticaddison.com/how-to-convert-a-quaternion-to-a-rotation-matrix/ but the formula is different

PythonClient/multirotor/sensorframe_lidar_pointcloud.py
```diff
@@ -0,0 +1,73 @@
...
+            pass
+            #time.sleep(5)
+            airsim.wait_key('Press any key to reset to original state')
+            self.client.reset()
+            print("Done!\n")
+
+# main
+if __name__ == "__main__":
+    lidarTest = LidarTest()
+    lidarTest.execute('Drone1',['LidarSensor1','LidarSensor2'])
```
rajat2004, line 73: It'll better to add the exact settings used in the script itslef, that helps in testing and usage

rajat2004 (commented):
2 files should be enough, either in the multirotor or car folder. Replicating the exact code in different files isn't very useful, and the script can be ran for any vehicle or users can copy the code to incorporate into their scripts.

Also, you might want to add some details about this in `docs/lidar.md`, it'll be more visible to users in that way

PythonClient/multirotor/sensorframe_lidar_pointcloud.py
euboon1601: To be completely honest, I got it from this Wikipedia page:
https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles
where it was supposedly sourced from Q54 of this Matrix/Quaternion FAQ compilation:
https://j3d.org/matrix_faq/matrfaq_latest.html#Q54

PythonClient/multirotor/sensorframe_lidar_pointcloud.py
euboon1601: Would having a comment that goes something like this be sufficient? In the case of the sensor frame script at least, with the proper indentation of course.

 _Sample settings.json used for this script:
    {
        "SeeDocsAt": "https://github.com/Microsoft/AirSim/blob/master/docs/settings_json.md",
        "SettingsVersion": 1.2,
        "SimMode": "Multirotor",
         "Vehicles": {
            "Drone1": {
                "VehicleType": "SimpleFlight",
                "AutoCreate": true,
                "Sensors": {
                    "LidarSensor1": { 
                        "SensorType": 6,
                        "Enabled" : true,
                        ...
                        "DataFrame": "SensorLocalFrame"
                    },
                    "LidarSensor2": { 
                        "SensorType": 6,
                        "Enabled" : true,
                        ...
                        "DataFrame": "SensorLocalFrame"
                    }
                }
            }
        }
    }_

PythonClient/multirotor/sensorframe_lidar_pointcloud.py
euboon1601: This seems a lot more intuitive, thanks for the suggestion!

PythonClient/multirotor/sensorframe_lidar_pointcloud.py
```diff
@@ -0,0 +1,99 @@
+# Python client example to get Lidar data from a drone, although this script works for any AirSim-supported vehicle
+# This script is for Lidar sensors using 'SensorLocalFrame' as DataFrame under settings.json.
+# Sample settings.json used for this script:
+#    {
+#        "SeeDocsAt": "https://github.com/Microsoft/AirSim/blob/master/docs/settings_json.md",
+#        "SettingsVersion": 1.2,
```
rajat2004, line 6: Ah, it might be better to use multiline comment for this, see [high_res_camera.py](https://github.com/microsoft/AirSim/blob/master/PythonClient/multirotor/high_res_camera.py) or [multi_agent_drone.py](https://github.com/microsoft/AirSim/blob/master/PythonClient/multirotor/multi_agent_drone.py). Main reason I would say is that if someone wants to copy the settings, they'll have to remove the `#` on each line. Some editors can do this in one shot, but better to keep it simple, it'll also be important when viewing the file on GitHub itself

Upto you and the maintainers

PythonClient/multirotor/sensorframe_lidar_pointcloud.py
```diff
@@ -0,0 +1,99 @@
...
+#        }
+#    }
+import setup_path 
+import airsim
+
+import sys
+import math
+import time
+import argparse
+import pprint
```
rajat2004, line 36: Unused imports

PythonClient/multirotor/sensorframe_lidar_pointcloud.py
```diff
@@ -0,0 +1,99 @@
...
+import numpy as np
+
+# Makes the drone fly and get Lidar data
```
rajat2004, line 39: Outdated comment

PythonClient/multirotor/sensorframe_lidar_pointcloud.py
```diff
@@ -0,0 +1,99 @@
...
+        existing_data_cleared = False   #change to true to superimpose new scans onto existing .asc files
+        try:
+            while True:
+                for lidar_name in lidar_names:
+                    filename = f"{vehicle_name}_{lidar_name}_pointcloud.asc"
+                    if not existing_data_cleared:
+                        f = open(filename,'w')
+                    else:
+                        f = open(filename,'a')
+                    lidarData = self.client.getLidarData(lidar_name=lidar_name,vehicle_name=vehicle_name)
```
rajat2004, line 62: Snake case - `lidar_data`

PythonClient/multirotor/sensorframe_lidar_pointcloud.py
```diff
@@ -0,0 +1,99 @@
...
+                    [q0,q1,q2,q3] = [lidarData.pose.orientation.w_val,lidarData.pose.orientation.x_val,lidarData.pose.orientation.y_val,lidarData.pose.orientation.z_val]
```
rajat2004, line 63: No need to create a list here

```python
orientation = lidar_data.pose.orientation
q0, q1, q2, q3 = orientation.w_val, orientation.x_val, orientation.y_val, orientation.z_val
```

PythonClient/multirotor/sensorframe_lidar_pointcloud.py
```diff
@@ -0,0 +1,99 @@
...
+                        xyz_flag = xyz_flag + 1
+                        if (xyz_flag == 3):
+                            [corrected_x,corrected_y,corrected_z] = np.matmul(rotation_matrix,xyz_coordinates)
+                            final_x = corrected_x + lidarData.pose.position.x_val
+                            final_y = corrected_y + lidarData.pose.position.y_val
+                            final_z = corrected_z + lidarData.pose.position.z_val
+                            f.write("%f %f %f %d %d %d \n" % (final_x,final_y,final_z,255,255,0))
+                            xyz_flag=0
+                        else:
+                            pass
```
rajat2004, line 86: Using Python slicing makes this easier to understand and much shorter -

```python
position = lidar_data.pose.position

for i in range(0, len(lidar_data.point_cloud), 3):
    xyz = lidar_data.point_cloud[i:i+3]

    corrected_x, corrected_y, corrected_z = np.matmul(rotation_matrix, np.asarray(xyz))
    final_x = corrected_x + position.x_val
    final_y = corrected_y + position.y_val
    final_z = corrected_z + position.z_val

    f.write("%f %f %f %d %d %d \n" % (final_x,final_y,final_z,255,255,0))
```

PythonClient/multirotor/sensorframe_lidar_pointcloud.py
```diff
@@ -0,0 +1,99 @@
...
+                    f.close()
+                existing_data_cleared = True
+        except KeyboardInterrupt:
+            pass
+            #time.sleep(5)
+            airsim.wait_key('Press any key to reset to original state')
+            self.client.reset()
```
rajat2004, line 93: Since this is supposed to be a data collection script, best to not affect the vehicle state itself. `reset()` will reset the vehicle to its original position, which might not be intended
Also remove the commented out line

rajat2004 (commented):
Really sorry for the piecewise review, was intending to do this yesterday but had some work

Also, you can click 'Resolve Conversation' if you feel the comment has been addressed, it'll make things easier for everyone to see what is outdated and still relevant. I can't do it myself since only maintainers and PR authors can

PythonClient/multirotor/sensorframe_lidar_pointcloud.py
rajat2004: Actually, they're both the same formulas, e.g. its a unit quaternion (`q0*q0 + q1*q1 + q2*q2 + q3*q3 == 1`), so the values actually turn out to be same. Any article could be used, but good to add a reference

PythonClient/multirotor/sensorframe_lidar_pointcloud.py
euboon1601: thanks for pointing it out, I've removed them in the update to the scripts.

PythonClient/multirotor/sensorframe_lidar_pointcloud.py
euboon1601: this has been changed in the current revision, thanks for letting me know of this concern

PythonClient/multirotor/sensorframe_lidar_pointcloud.py
euboon1601: thanks for your input with this, I agree that it makes things a lot simpler, and much neater too!

euboon1601:
I really can't thank you enough for taking your time to do all these reviews, and for providing all the extremely helpful inputs to this PR. This has been a really great experience as someone who is new to this whole situation, and I believe that everything I've learned so far from this will certainly be of great use for me moving forward. I do apologize once again if there was any form of inconvenience caused from having to look through this PR so many times due to all the things I've overlooked.

rajat2004:
No problem at all, it was great to help others improve and consequently myself as well, with another big plus of better code going into the project

zimmy87:
Ran the new scripts locally and the output looks correct to me, so I'm moving ahead with merge

PR #3465, opened 2021-03-15
Fix drone shaking at high altitudes

### LINKED ISSUE #1394: Question: drone starts to shake if the altitude is over 50m
Hi, 
Version: Airsim 1.2.0, binary.
Control: call moveOnPathSync with default setting but assigned waypoints.
Problem: looks fine if the altitude is lower than 50m. However, if I tried to fly along the same path with an altitude over 60m, the drone started to shake. Does anyone have an idea about this abnormal phenomenon?

zimmy87:
Fixes: #1394

## About
Updates MultirotorApiBase.cpp to use the velocity PID with target z velocity 0 when we are close to our goal altitude to fix drone shaking. Thank you @ACLeighner for finding this fix!

## How Has This Been Tested?
A test script was used that calls moveToPositionAsync to send the drone to an altitude of 60m then flies the drone in a square path to observe the shaking described in #1394.


ahmed-elsaharti:
Thank you for this PR @zimmy87. I was literally about to start digging into this issue last night!
Will test ASAP and edit this comment with the results :+1:

zimmy87:
This PR has been replaced and will now be tracked in #3478

PR #3468, opened 2021-03-15
PythonClient: replace time.clock with time.time

dayjaby:
## About
This allows to run the opencv_show.py example even on Ubuntu 20.04 with Python3.8+ as time.clock() got deprecated.

## How Has This Been Tested?
Tested on Ubuntu 20.04


jonyMarino:
Hi David!
Thanks for contributing!

PR #3470, opened 2021-03-16
Avoid using deprecated git clone with basic auth

dayjaby:
GitHub is complaining if you try to clone UnrealEngine, which requires
you to authenticate into GitHub:

```
You recently used a password to access the repository at EpicGames/UnrealEngine with git using git/2.25.1.

Basic authentication using a password to Git is deprecated and will soon no longer work. Visit https://github.blog/2020-12-15-token-authentication-requirements-for-git-operations/ for more information around suggested workarounds and removal dates.
```

Also adding the default install location for UnrealEngine to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ cmake_install.cmake
 CMakeCache.txt
 
 #Unreal Engine
+/UnrealEngine
 StarterContent/
 Saved/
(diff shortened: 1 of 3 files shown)


zimmy87:
Tested this locally and it works for me so I am moving ahead with merging. Thanks for the contribution @dayjaby!

Issue #232, opened 2017-05-24
Get location and bounding box of objects in environment


Geekking:
How can I get the 2D location and 3D location of objects in the scene? For example, when the drone is flying on the road, can I get the 3D box of the cars or pedestrains on the road?

sytelus:
Yes, you can but you will need to dig in to Unreal Engine programming. Basically environment has all these meshes so you can iterate through meshes that interests you and then there are direct methods to get its position and size in world coordinates. If you have never done any Unreal engine programming, its non-trivial, however.

Geekking:
OK, thanks.

cangokalp:
@Geekking Have you figured this out? I wish there was an easy way to get the coordinates and bounding box of the objects in the environment too. I need the same thing

alessandro-montanari:
@Geekking looking for the same thing here.

evroon:
Probably way too late, but for future reference: You can get bounding boxes without modifying the source code using the segmentation outputs.

First, set the object IDs of everything to zero, except the drone (if you want to get bounding boxes around the drone):
```python
client.simSetSegmentationObjectID("[\w]*", 0, True)
client.simSetSegmentationObjectID("Drone[\w]*", 255, True)
```
Save the segmentation images as in the [hello-drone example](https://microsoft.github.io/AirSim/apis/#hello-drone).

Now you can calculate the bounding box simply by checking which rows and columns contain non-zero pixels: (there is probably a more efficient way, but it works)

```python
import numpy as np
import cv2

img = cv2.imread('path_to_image.png')

height, width = img.shape[:2]
img_size = np.array([width, height])
start_x, start_y = -1, -1
end_x, end_y = -1, -1

for y in range(height):
    if np.sum(img[y, :, :]) > 0.0:
        end_y = y

        if start_y == -1:
            start_y = y

for x in range(width):
    if np.sum(img[:, x, :]) > 0.0:
        end_x = x

        if start_x == -1:
            start_x = x

img = cv2.rectangle(img, (start_x, start_y), (end_x, end_y), (255, 0, 0))
cv2.imwrite('frame.png', img)
```

The (very thin, sorry) blue box shows the bounding box in the acquired segmentation image:
![frame](https://user-images.githubusercontent.com/11857441/110245888-79fd4d00-7f65-11eb-9b50-636294dc2162.png)

rancheng:
[...]

PR #3472, opened 2021-03-16
Add object detection capability and python API

alonfaraj:
## About

This PR adds the capability to detect objects with Unreal.
It support setting radius from camera to search for objects and setting object name in wildcard format.
One can control these settings for each camera, image type and vehicle combination separately.
It output relevant information as described in `DetectionInfo`.

Itcurrently support only `ImageType::Scene`, but can be extended by attaching the `DetectionComponent` in BP_PIPCamera to the relevant camera and add corresponding lines in `APIPCamera::PostInitializeComponents`

- Detection APIs implementation:
  - [x] simSetDetectionFilterRadius
  - [x] simAddDetectionFilterMeshName
  - [x] simClearDetectionMeshNames
  - [x] simGetDetections

- Detection struct:
```
class DetectionInfo(MsgpackMixin):
    name = ''
    geoPoint = GeoPoint()
    box2D = Box2D()
    box3D = Box3D()
    relative_pose = Pose()
```

TODO:
- Implement some get API functions
- Add Enable/Disable Detection capability API or from `settings.json`

Most of the detection and object filter code was copied from https://github.com/unrealgt/UnrealGT and changed for my own needs.

Probably much more work to do but I'm using it for a while and though other users might find it useful.

## How Has This Been Tested?

Tested on Blocks and ModularNeighborhood environments by running the detection python script in this PR (Windows).

Example API Call -
```
camera_name = "0"
image_type = airsim.ImageType.Scene

client.simSetDetectionFilterRadius(camera_name, image_type, 80 * 100) # in [cm]
client.simAddDetectionFilterMeshName(camera_name, image_type, "Car_*") 
client.simGetDetections(camera_name, image_type)
client.simClearDetectionMeshNames(camera_name, image_type)
```
Example output -
```
Cylinder: <DetectionInfo> {   'box2D': <Box2D> {   'max': <Vector2r> {   'x_val': 617.025634765625,
    'y_val': 583.5487060546875},
    'min': <Vector2r> {   'x_val': 485.74359130859375,
    'y_val': 438.33465576171875}},
    'box3D': <Box3D> {   'max': <Vector3r> {   'x_val': 4.900000095367432,
    'y_val': 0.7999999523162842,
    'z_val': 0.5199999809265137},
    'min': <Vector3r> {   'x_val': 3.8999998569488525,
    'y_val': -0.19999998807907104,
    'z_val': 1.5199999809265137}},
    'geo_point': <GeoPoint> {   'altitude': 16.979999542236328,
    'latitude': 32.28772183970703,
    'longitude': 34.864785008379876},
    'name': 'Cylinder9_2',
    'relative_pose': <Pose> {   'orientation': <Quaternionr> {   'w_val': 0.9929741621017456,
    'x_val': 0.0038591264747083187,
    'y_val': -0.11333247274160385,
    'z_val': 0.03381215035915375},
    'position': <Vector3r> {   'x_val': 4.400000095367432,
    'y_val': 0.29999998211860657,
    'z_val': 1.0199999809265137}}}
```

- Blocks
Unreal
blocks_ue4
Python
blocks_python

- ModularNeighborhood
Unreal

MN_ue4
Python
MN_python


alonfaraj:
@rajat2004 any idea why checks failed for Unity build?

AirLib/include/api/RpcLibClientBase.hpp
```diff
@@ -15,6 +15,7 @@
 #include "physics/Kinematics.hpp"
 #include "physics/Environment.hpp"
 #include "api/WorldSimApiBase.hpp"
+#include <map>
```
rajat2004, line 18: unused include

AirLib/src/api/RpcLibServerBase.cpp
```diff
@@ -180,6 +180,19 @@ RpcLibServerBase::RpcLibServerBase(ApiProvider* api_provider, const std::string&
...
+	pimpl_->server.bind("simAddDetectionFilterMeshName", [&](const std::string& mesh_name) -> void {
+		getWorldSimApi()->addDetectionFilterMeshName(mesh_name);
+		});
+	pimpl_->server.bind("simSetDetectionFilterRadius", [&](const float radius_cm) -> void {
+		getWorldSimApi()->setDetectionFilterRadius(radius_cm);
+		});
+
+	pimpl_->server.bind("simGetDetections", [&](const std::string& camera_name, ImageCaptureBase::ImageType type, const std::string& vehicle_name) -> vector<RpcLibAdapatorsBase::DetectionInfo> {
+        const auto& response = getVehicleSimApi(vehicle_name)->getDetections(camera_name, type);
+        //return RpcLibAdapatorsBase::DetectionInfo(response);
```
rajat2004, line 192: Can be removed

PythonClient/detection/detection.py
```diff
@@ -0,0 +1,38 @@
...
+while True:
+    rawImage = client.simGetImage(camera_name, image_type)
+    if not rawImage:
+        continue
+    png = cv2.imdecode(airsim.string_to_uint8_array(rawImage), cv2.IMREAD_UNCHANGED)
+    cars = client.simGetDetections(camera_name, image_type)
+    if cars:
+        for car in cars:
+            s = pprint.pformat(car)
+            print("Cylinder: %s" % s)
```
rajat2004, line 29: It should be cylinders here correct?

PythonClient/detection/detection.py
```diff
@@ -0,0 +1,38 @@
...
+
+            cv2.rectangle(png,(car.topLeft_x,car.topLeft_y),(car.bottomRight_x,car.bottomRight_y),(255,0,0),2)
```
rajat2004, line 31: Maybe even display the name of the object above the box? Upto you if it seems useful

Unreal/Plugins/AirSim/Source/AirBlueprintLib.cpp
```diff
@@ -585,6 +587,27 @@ std::vector<msr::airlib::MeshPositionVertexBuffersResponse> UAirBlueprintLib::Ge
...
 
 
+void UAirBlueprintLib::addDetectionFilterMeshName(const std::string& name)
+{
+	for (TObjectIterator<UDetectionComponent> detectionIterator; detectionIterator; ++detectionIterator)
+	{
+		UDetectionComponent* detection = *detectionIterator;
+        if (!detection->ObjectFilter.WildcardMeshNames.Contains(FString(name.c_str())))
+        {
+            detection->ObjectFilter.WildcardMeshNames.Add(FString(name.c_str()));
```
rajat2004, line 597: AirSim uses snake case for variables, might be better to convert `detectionIterator`. You could even shorten it to`itr` also

`FString(name.c_str())` could also be constructed once and used, but probably the compiler will anyways do the optimization so up to you

Unreal/Plugins/AirSim/Source/AirBlueprintLib.cpp
```diff
@@ -585,6 +587,27 @@ std::vector<msr::airlib::MeshPositionVertexBuffersResponse> UAirBlueprintLib::Ge
...
+        }
+	}
+}
+
+void UAirBlueprintLib::setDetectionFilterRadius(const float radius_cm)
+{
+	for (TObjectIterator<UDetectionComponent> detectionIterator; detectionIterator; ++detectionIterator)
+	{
+		UDetectionComponent* detection = *detectionIterator;
+        detection->MaxDistanceToCamera = radius_cm;
```
rajat2004, line 607: Same, snake case and maybe shorten to `itr`.
Inside loop could be reduced to `(*itr)->MaxDist..` but just my pref, again up to you and maintainers

Unreal/Plugins/AirSim/Source/DetectionComponent.h
```diff
@@ -0,0 +1,62 @@
...
+
+	UPROPERTY()
+		FBox2D BBox;
+
+	UPROPERTY()
+		FVector Location;
+};
+
+UCLASS( ClassGroup=(Custom), meta=(BlueprintSpawnableComponent) )
+class AIRSIM_API UDetectionComponent : public USceneComponent
```
rajat2004, line 26: Airsim classes are generally prefixed by A, i.e `ADetectionComponent`. U is for UE4 classes
Whether to change or not is up to the maintainers

Unreal/Plugins/AirSim/Source/ObjectFilter.cpp
```diff
@@ -0,0 +1,250 @@
...
+	
+	// If not set we assume a match
+/*
+	bool bMatchesStaticMesh = !StaticMesh;
+	bool bMatchesSkeletalMesh = !SkeletalMesh;
+	bool bMatchesWildcardMeshName = WildcardMeshNames.Num() == 0;//  IsEmpty();
+	bool bMatchesActorClass = !ActorClass;
+	bool bMatchesActorTag = ActorTag.IsNone();
+	bool bMatchesComponentClass = !ComponentClass;
+	bool bMatchesComponentTag = ComponentTag.IsNone();*/
```
rajat2004, line 38: Should this still be kept?

Also, lots of commented out code below, that also needs to be looked at

Unreal/Plugins/AirSim/Source/PawnSimApi.h
```diff
@@ -21,6 +21,7 @@
 
 
 #include "PawnEvents.h"
+#include <map>
```
rajat2004, line 24: Unused

rajat2004 (commented):
Just a brief review. I haven't yet looked at the detection part of the code, will do so later

One main point is that AirSim uses 4 spaces instead of tabs, it'll be best to convert to remain consistent. Another is that snake case is used for variables, which would also be another good thing to fix

evroon:
I don't want to undermine your hard work, but I want to note that there is a simpler way of getting bounding boxes around objects such as cars by using the segmentation images. I explained it [here](https://github.com/microsoft/AirSim/issues/232#issuecomment-792300498). This way, you get pixel-perfect bounding boxes. By projecting a 3D box, you will always end up (except for objects that have the shape of a box of course) with bigger bounding boxes than optimal.

PPakalns:
@evroon The bounding box retrieval from segmentation images works only in cases when objects do not overlap in the camera view. If there is a forest with a lot of trees or car in front of an another car, then these tightly coupled different objects can not be differentiated in segmentation images. So this bounding box feature is very welcomed in cases such as these.

evroon:
Yes that's true, in case of occlusion "my" method does not work. But in my case I use it to train yolov4 for example and then you don't need such data AFAIK. I value correct bounding boxes more.
So I'm just curious what the use case is for having bounding boxes for (partially) occluded objects. Is it for training NNs or something else?

Btw I think you can also make my method work by taking multiple segmentation images of the same frame and changing the visibility of objects, but that is more complicated and less performant of course.

PPakalns:
@evron I will be using it for prototyping object detection model where objects can be tightly located in the scene next to each other, like standard case of people passing in front of each other or, in my case, prototyping survey drone where it is important to correctly recognize each separate object, these objects can occlude each other little bit. Segmentation image approach makes it hard to annotate such objects with separate bounding boxes because their regions overlap.  

@alonfaraj At least now DetectionInfo data returns only object geolocation, for generating annotated data it would be useful if object position and orientation relative to the camera could be returned additionally. At least I will try to add such information myself :)  

Tomorrow will test this code and see how it works. Thanks @alonfaraj for such implementation  :1st_place_medal:

MoBaT:
@alonfaraj This is great! Was building the exact same thing but stumbled upon this. I have a few suggestions..

1) Can the `simSetDetectionFilterRadius` and `simAddDetectionFilterMeshName` api change where a camera name can be given? I would like to have different detections and radiuses done on different cameras. Looking at your implementation, it's a very easy add. So I suggest:

```
client.simSetDetectionFilterRadius("0", 80 * 100) # in [cm]
client.simAddDetectionFilterMeshName("0", "Car_*") 
```

2) Change the `simAddDetectionFilterMeshName` to accept a regex string instead of just a single wildcard similar to the `simListSceneObjects` api.

3) Add a `simClearDetections("regex")` call or to make it easier, a `simClearDetections()` with no search criteria. I would like this in case I have a scene with dynamic objects and I want to poll for new objects at a certain frequency. By clearing detections and adding detections.

4) Modify `simGetDetections` to return a Detection with the 3D boundingBox info also.

Unreal/Plugins/AirSim/Source/ObjectFilter.cpp
```diff
@@ -0,0 +1,250 @@
...
+		{
+			UStaticMeshComponent* StaticMeshComponent =
+				Cast<UStaticMeshComponent>(ActorComponent);
+			if (StaticMeshComponent) {
+				if (StaticMesh && StaticMeshComponent->GetStaticMesh() == StaticMesh) 
+				{
+					//bMatchesStaticMesh = true;
+					return true;
+				}
+				if (!WildcardMeshNames.Num() == 0 /*IsEmpty()*/ &&
```
PPakalns, line 62: While compiling suggested changes on Unreal Engine in linux environment, the following compiler warnings/errors are raised for all `/* isEmpty */` occurences where logical not operator `!` is used before `WildcardMeshNames.Num()`.
```
.../Plugins/AirSim/Source/ObjectFilter.cpp:151:8: error: logical not is only applied to the left hand side of this comparison [-Werror,-Wlogical-not-parentheses]
                        if (!WildcardMeshNames.Num() == 0 /*IsEmpty()*/ &&
```

As this code is checking that array is not empty, it can be simplfied to `WildcardMeshNames.Num() != 0` or `WildcardMeshNames.Num()`.

This warning needs to be resolved for all such occurrences in this file.

Unreal/Plugins/AirSim/Source/ObjectFilter.cpp
```diff
@@ -0,0 +1,250 @@
...
+					IsMatchAnyWildcard(StaticMeshComponent->GetStaticMesh()->GetName()))//.MatchesWildcard(WildcardMeshNames))
```
PPakalns, line 63: `StaticMeshComponent->GetStaticMesh()` can return nullptr if object doesn't have static mesh and the whole program crashes.

Locally tested code with added `StaticMeshComponent->GetStaticMesh() != nullptr` check. It worked correctly.

Unreal/Plugins/AirSim/Source/PawnSimApi.cpp
```diff
@@ -318,6 +318,31 @@ void PawnSimApi::reportState(msr::airlib::StateReporter& reporter)
...
+{
+    const APIPCamera* camera = getCamera(camera_name);
+    const TArray<FDetectionInfo> Detections = camera->getDetectionComponent(image_type, false)->GetDetections();
+    int detectionCount = Detections.Num();
+    std::vector<msr::airlib::DetectionInfo> result(detectionCount);   
+    
+    
+    for (int i = 0; i < Detections.Num(); i++)
+    {
+        result[i].name = std::string(TCHAR_TO_UTF8(*(Detections[i].Actor->GetFName().ToString())));
```
PPakalns, line 331: Locally tested pull request and got crash here.

`Detections[i].Actor` sometimes can be a nullptr and causes crash when de-referencing is done to call `GetFName()`.

Locally tested that discarding result when Actor is nullptr works correctly.

PPakalns (commented):
Additionally, object detection results for some objects are flickering (in some frames object is visible, in some it is not even when camera position is not changed).   Will look into cause of it.

AirLib/include/api/RpcLibAdapatorsBase.hpp
```diff
@@ -221,6 +221,64 @@ class RpcLibAdapatorsBase {
...
+		) {
+			std::vector<msr::airlib::DetectionInfo> request;
+			for (const auto& item : request_adapter)
+				request.push_back(item.to());
+
+			return request;
+		}
+
+	};
+
```
evroon, line 281: You seem to be using tabs, but the AirSim has by default four spaces as indentation.

evroon (changes requested):

alonfaraj:
> @alonfaraj This is great! Was building the exact same thing but stumbled upon this. I have a few suggestions..
> 
> 1. Can the `simSetDetectionFilterRadius` and `simAddDetectionFilterMeshName` api change where a camera name can be given? I would like to have different detections and radiuses done on different cameras. Looking at your implementation, it's a very easy add. So I suggest:
> 
> ```
> client.simSetDetectionFilterRadius("0", 80 * 100) # in [cm]
> client.simAddDetectionFilterMeshName("0", "Car_*") 
> ```
> 
> 1. Change the `simAddDetectionFilterMeshName` to accept a regex string instead of just a single wildcard similar to the `simListSceneObjects` api.
> 2. Add a `simClearDetections("regex")` call or to make it easier, a `simClearDetections()` with no search criteria. I would like this in case I have a scene with dynamic objects and I want to poll for new objects at a certain frequency. By clearing detections and adding detections.
> 3. Modify `simGetDetections` to return a Detection with the 3D boundingBox info also.

alonfaraj:
> @alonfaraj This is great! Was building the exact same thing but stumbled upon this. I have a few suggestions..
> 
> 1. Can the `simSetDetectionFilterRadius` and `simAddDetectionFilterMeshName` api change where a camera name can be given? I would like to have different detections and radiuses done on different cameras. Looking at your implementation, it's a very easy add. So I suggest:
> 
> ```
> client.simSetDetectionFilterRadius("0", 80 * 100) # in [cm]
> client.simAddDetectionFilterMeshName("0", "Car_*") 
> ```
> 
> 1. Change the `simAddDetectionFilterMeshName` to accept a regex string instead of just a single wildcard similar to the `simListSceneObjects` api.
> 2. Add a `simClearDetections("regex")` call or to make it easier, a `simClearDetections()` with no search criteria. I would like this in case I have a scene with dynamic objects and I want to poll for new objects at a certain frequency. By clearing detections and adding detections.
> 3. Modify `simGetDetections` to return a Detection with the 3D boundingBox info also.

Unreal/Plugins/AirSim/Source/ObjectFilter.cpp
alonfaraj: Thanks! I didn't check it on linux. Will fix it soon

alonfaraj:
@MoBaT Thanks for the suggestions!
I will probably make those changes soon.

About #4 - what is the purpose of 3D BB? Should it be in Geo as well?

alonfaraj:
@PPakalns Thank you very much for testing and fix those bugs!
I will make some fixes soon.

@evroon Seems like you already discussed it but I like your approach too :)
As @evroon said, my scenario is mostly when object are partially occluded by others and in this case I think it might be easier to use a "real" detection and not a segmentation.

PPakalns:
@alonfaraj 
Thank you again for this PR, it is very usable.

About 3D bounding box:
It is usable in cases when information about object placement can be used to make bounding boxes more precise, relevant in further data processing steps.

In my case I have similar situation as you, a lot of objects with partially overlapping boundaries, with bad visibility or rotated away from the observer. 

In processing steps I am improving `DetectionInfo` bounding box by shrinking it to the smallest size that still includes all pixels of this object type in segmentation image. This provides good approximation that for each separate object I have a bounding box and it is the small as possible. 

Next, these objects have very distinct two sides. I am planning to filter out all bounding boxes, where object is rotated away from the observer, because I am interested only in the front side of the object.  If `DetectionInfo` would return object position and its rotation (relative to the camera) it would be very usable as a lot of manual API calls for each individual object to retrieve its pose would not be needed. I am getting familiar with UE and Airsim implementation to get this additional info added to `DetectionInfo`.

I would suggest that this data is not provided in Geo coordinate space, but in coordinate space that is relative to the camera. This would be more consistent with other simulation APIs as `DetectionInfo` describes the world seen by the camera. Geo coordinates etc. can be easily calculated later using vehicle and camera positions.

alonfaraj:
@PPakalns Thanks for sharing you, defiantly help me understand other needs better.
I will try to add it soon, in addition to relative position and orientation of the detected object.

Unreal/Plugins/AirSim/Source/PawnSimApi.cpp
alonfaraj: That's weird because it iterate over the result of GetDetections() so all members should be valid.
Checking for nullptr will cause us lose detection we already found.
Any idea why is it happening?

PPakalns: Looked into your code, I agree, it shouldn't be due to that, it shouldn't be a nullptr. 

Is it possible that Actor can get de-allocated from memory if player moves around?

Will try to reproduce this error again and see what causes it.

PythonClient/detection/detection.py
alonfaraj: Good idea, thanks!
Added and update my post.

Unreal/Plugins/AirSim/Source/PawnSimApi.cpp
PPakalns: Removed workaround and didn't have crash again in this part of code. Will close this for now.

PPakalns:
@alonfaraj
Nice changes, thanks for adding them. Already tested them and it is very usable for me with small changes:

One change that I think should be made, the relative pose doesn't take into account camera orientation and object orientation. I suggest that all poses, coordinates are returned in camera coordinate space.

I made modifications for relative_pose.orientation to correctly return object orientation as seen from camera. This allowed me to easily filter out objects that were oriented away from the camera.

Applied modifications: [diff.txt](https://github.com/microsoft/AirSim/files/6236103/diff.txt) . Maybe this is usable :)

alonfaraj:
@PPakalns 
You right!
Thank you for this fix, already applied :)

Unreal/Plugins/AirSim/Source/ObjectFilter.cpp
```diff
@@ -0,0 +1,239 @@
...
+			UStaticMeshComponent* StaticMeshComponent =
+				Cast<UStaticMeshComponent>(ActorComponent);
+			if (StaticMeshComponent) {
+				if (StaticMesh && StaticMeshComponent->GetStaticMesh() == StaticMesh) 
+				{
+					//bMatchesStaticMesh = true;
+					return true;
+				}
+				if (WildcardMeshNames.Num() != 0 &&
+					StaticMeshComponent->GetStaticMesh()->IsValidLowLevel() &&
```
PPakalns, line 52: This doesn't seem to work correctly.

I am getting error: `[2021.03.31-12.53.05:810][126]LogUObjectBase: Warning: NULL object`, which is raised when IsValidLowLevel is called with NULL object.

I would suggest that `GetStaticMesh` is checked against `nullptr` before this call.

Unreal/Plugins/AirSim/Source/PawnSimApi.cpp
PPakalns: Today with the newest version from this branch got error in this part of code multiple times.

```
Signal 11 caught.
Malloc Size=65538 LargeMemoryPoolOffset=65554 
CommonUnixCrashHandler: Signal=11
[2021.03.31-13.11.18:774][671]LogCore: === Critical error: ===
Unhandled Exception: SIGSEGV: invalid attempt to read memory at address 0x0000000000000018

[2021.03.31-13.11.18:774][671]LogCore: Fatal error!

0x00007ff5fa4c734d libUE4Editor-AirSim.so!PawnSimApi::getDetections(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, msr::airlib::ImageCaptureBase::ImageType) const [/Plugins/AirSim/Source/PawnSimApi.cpp:331]
```

Unreal/Plugins/AirSim/Source/ObjectFilter.cpp
alonfaraj: I thought it equivalent to check against nullptr, I will fix it

Unreal/Plugins/AirSim/Source/PawnSimApi.cpp
```diff
@@ -318,6 +318,34 @@ void PawnSimApi::reportState(msr::airlib::StateReporter& reporter)
...
```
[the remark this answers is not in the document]
alonfaraj: Any idea why? Do you run it on linux?
Seems like it's not reproduce in my machine (Windows).
Also - this part of set Actor and get its name didn't change..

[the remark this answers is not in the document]
PPakalns: This is the same issue that I previously reported with older version in this place too, but couldn't reproduce.  (I had left some other changes in this function, maybe that was the reason why I couldn't reproduce)

Will investigate the reason why it happens and report it here.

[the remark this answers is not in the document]
alonfaraj: Thank you for that!
It's interesting I didn't have any crash on my side.
Also, I'm wondering how does AirSim implement all these get/set API functions? I didn't see they using lock's.
For example get sensors data or getImages().

I used your commit but made one small change, I prefer not using std inside pure Unreal classes unless I have too, so replaced `std::move` with `MoveTemp`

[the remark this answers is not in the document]
alonfaraj: So if I understood correctly, seems like they way AirSim taking care of multithread is using `UAirBlueprintLib::RunCommandOnGameThread`.

I wrapped `getDetections()` inside `RunCommandOnGameThread` without implicit using lock inside `DetectionComponent`.
It didn't crash on my machine before so it would be great if you can test it and report if it's work :)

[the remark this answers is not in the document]
PPakalns: I have a lot of objects in my scene that get detected so it results in large changes of this array, therefore there is a bigger probability of it being modified at the same time.

I am not sure about how multi threading is exactly done in UE4, we probably need some advice from more experienced UE4 programmers :)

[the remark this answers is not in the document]
alonfaraj: `DetectionComponent` and all other `Tick` functions run on the main thread in UE - `GameThread`.
To avoid manage read and write from different threads, like in our case, AirSim using `UAirBlueprintLib::RunCommandOnGameThread` so basically everything happen form the same thread.

I believe https://github.com/alonfaraj/AirSim/commit/2f0f0db03e4f8e716d2e1d609158ed114f41950c should fix all these problem you had in `getDetections()`

alonfaraj:
@rajat2004 @PPakalns @evroon 
I think I'm done here for now. 
Do you mind take a look whenever you free? Any feedback would be greatly appreciated!

Unreal/Plugins/AirSim/Source/DetectionComponent.cpp
```diff
@@ -0,0 +1,214 @@
...
+    
+    // initialize viewinfo for projection matrix
+    FMinimalViewInfo Info; 
+    Info.Location = SceneCaptureComponent2D->GetComponentTransform().GetLocation();
+    Info.Rotation = SceneCaptureComponent2D->GetComponentTransform().GetRotation().Rotator();
+    Info.FOV = SceneCaptureComponent2D->FOVAngle;
+    Info.ProjectionMode = SceneCaptureComponent2D->ProjectionType;
+    Info.AspectRatio = float(TextureTarget->SizeX) / float(TextureTarget->SizeY);
+    Info.OrthoNearClipPlane = 1;
+    Info.OrthoFarClipPlane = 0;//1000;
```
evroon, line 101: Shouldn't this be a much larger value than 0, like 1000 as in the comment?

Unreal/Plugins/AirSim/Source/DetectionComponent.cpp
```diff
@@ -0,0 +1,214 @@
...
+                    Detection.RelativeTransform = FTransform(GetRelativeRotation(Actor->GetActorLocation(), Actor->GetActorRotation()),
+                        GetRelativeLocation(Actor->GetActorLocation()));
+                    CachedDetections.Add(Detection);
+
+                    /*  ---Debug only---
+                    FVector Origin;
+                    FVector Extend;
+                    Actor->GetActorBounds(false, Origin, Extend);
+                    DrawDebugBox(GetWorld(), Origin, Extend, FColor::Red, false, 0.03, 0, 4.f);
+                        ---------------- */
```
evroon, line 68: I assume this is something you want to keep in case it is needed or did you commit it accidentally?

Unreal/Plugins/AirSim/Source/ObjectFilter.cpp
```diff
@@ -0,0 +1,246 @@
...
+    bool bMatchesSkeletalMesh = false;
+    bool bMatchesWildcardMeshName = false;
+    bool bMatchesActorClass = false;
+    bool bMatchesActorTag = false;
+    bool bMatchesComponentClass = false;
+    bool bMatchesComponentTag = false;
+    
+    if (!WildcardMeshNames.Num() == 0 && IsMatchAnyWildcard(Actor->GetName()))
+    {
+        //bMatchesActorName = true;
```
evroon, line 40: Why are there a lot of lines like this in this file? Maybe you can also add some comment(s) in this function explaining why you need those if-statements.

Unreal/Plugins/AirSim/Source/ObjectFilter.cpp
```diff
@@ -0,0 +1,246 @@
...
+                StaticMeshComponent->GetStaticMesh()->IsValidLowLevel() &&
+                IsMatchAnyWildcard(StaticMeshComponent->GetStaticMesh()->GetName()))
+            {
+                bMatchesWildcardMeshName = true;
+            }
+        }
+    }
+    if (SkeletalMesh || WildcardMeshNames.Num() != 0) 
+    {
+        USkeletalMeshComponent* SkeletalMeshComponent =	Cast<USkeletalMeshComponent>(ActorComponent);
```
evroon, line 157: The tab here is a bit weird.

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
```diff
@@ -482,17 +492,28 @@ void APIPCamera::enableCaptureComponent(const APIPCamera::ImageType type, bool i
...
     USceneCaptureComponent2D* capture = getCaptureComponent(type, false);
     if (capture != nullptr) {
+        UDetectionComponent* detection = getDetectionComponent(type, false);
         if (is_enabled) {
             //do not make unnecessary calls to Activate() which otherwise causes crash in Unreal
             if (!capture->IsActive() || capture->TextureTarget == nullptr) {
                 capture->TextureTarget = getRenderTarget(type, false);
                 capture->Activate();
+                if (detection != nullptr)
+				{
```
evroon, line 502: The indentation in this file is still a bit weird.

Unreal/Plugins/AirSim/Source/PawnSimApi.cpp
```diff
@@ -318,6 +318,69 @@ void PawnSimApi::reportState(msr::airlib::StateReporter& reporter)
...
 
+void PawnSimApi::addDetectionFilterMeshName(const std::string& camera_name, ImageCaptureBase::ImageType image_type, const std::string& mesh_name)
+{
+    UAirBlueprintLib::RunCommandOnGameThread([this, camera_name, image_type, mesh_name]() {
+        const APIPCamera* camera = getCamera(camera_name);
+        UDetectionComponent* detection_comp = camera->getDetectionComponent(image_type, false);
+
+        FString name = FString(mesh_name.c_str());
+
+        if (!detection_comp->ObjectFilter.WildcardMeshNames.Contains(name))
```
evroon, line 329: I think btw the style in this project is to put the opening bracket on the same line as the if-statement, and not on a new line, but I don't think you have to change that.

evroon (changes requested):
The code looks really nice, thanks for converting the tabs. I just had some minor tips. I don't have that much experience with AirSim, so probably more people have to review this.

Oh and there are conflicting files now.

Unreal/Plugins/AirSim/Source/DetectionComponent.cpp
alonfaraj: You right, forgot it somehow :)

Unreal/Plugins/AirSim/Source/ObjectFilter.cpp
alonfaraj: Weird, VS says spcs for the whole file but there are few tabs left, thanks!

Unreal/Plugins/AirSim/Source/PawnSimApi.cpp
alonfaraj: Spaces and opening bracket on the same line in one project, definitely the opposite of my style :)
I will leave it like that for now unless maintainers will ask to change it.

Unreal/Plugins/AirSim/Source/DetectionComponent.cpp
alonfaraj: I think set `OrthoFarClipPlane = 0` will take the max value but I will change it to higher implicit value

Unreal/Plugins/AirSim/Source/ObjectFilter.cpp
alonfaraj: The original code assumed a match if all of the search parameters (tag name, component class etc.) wasn't set.
There was a check of and between all of them. That's why there was only one return value at the end of the function.

I modified the code so now I assume no match if there are no search parameters.
The function immediately return true in case a match was found.
I left the bool values remarked in case someone want to change the logic, it will make it easier to understand all the if-else logic. Thought it will be a good idea unless maintainers think it's not necessary.

One last thing, most of the parameters are not relevant for use in API (at least for now), it can easily be used from the editor.

alonfaraj:
@evroon Thanks for the feedbacks!

PPakalns:
Overall everything works very nicely now.

One more last feature request. This could be done in follow up merge request too:

Returned `DetectionInfo` object relative pose field contains:
* [x] Object position,
* [x] Object rotation (orientation),
* [ ] Object scale.

The object scale data is missing, as object orientation is already known and object local scale axis can be calculated from that, simple copy of object local scale values could be provided in DetectionInfo too to provide full (position, rotation, scale) information for each object. This is usable if scene contains multiple objects each scaled to different size.

alonfaraj:
@PPakalns Thanks!
I will try to add it soon, probably in a follow up PR like you suggested.

jonyMarino:
Hi! I still couldn't see this job in detail. But I want to tell you that we are very interested in this functionality, and I thank you all very much for your work, especially to the author @alonfaraj. Keep it up!

alonfaraj:
@jonyMarino Thanks!

Unreal/Plugins/AirSim/Source/ObjectFilter.cpp
```diff
@@ -0,0 +1,246 @@
...
+    bool bMatchesActorName = false;
+    bool bMatchesStaticMesh = false;
+    bool bMatchesSkeletalMesh = false;
+    bool bMatchesWildcardMeshName = false;
+    bool bMatchesActorClass = false;
+    bool bMatchesActorTag = false;
+    bool bMatchesComponentClass = false;
+    bool bMatchesComponentTag = false;
+    
+    if (!WildcardMeshNames.Num() == 0 && IsMatchAnyWildcard(Actor->GetName()))
```
PPakalns, line 38: On linux Unreal engine compilation fails with the following error:

In file included from Plugins/AirSim/Intermediate/Build/Linux/B4D820EA/UE4Editor/Development/AirSim/Module.AirSim.cpp:10:
Plugins/AirSim/Source/ObjectFilter.cpp:38:9: error: logical not is only applied to the left hand side of this comparison [-Werror,-Wlogical-not-parentheses]
    if (!WildcardMeshNames.Num() == 0 && IsMatchAnyWildcard(Actor->GetName()))

PPakalns:
I am testing this and it looks like object detections are flickering, they are available in some api calls, are missing in some api calls even if camera didn't move by very much and object is still very well visible.

Trying to find the cause of this non-determined behavior.

alonfaraj:
@PPakalns 
You right. This is how I handle the detection process but I couldn't find any better way.
There is a function [WasRecentlyRendered()](https://docs.unrealengine.com/en-US/API/Runtime/Engine/GameFramework/AActor/WasRecentlyRendered/index.html), but it relevent only to the GameViewport.
I open to hear any other suggestions.
I can expose an API to set number of random points but I don't think it's the best way to solve it.

zimmy87:
hey @alonfaraj, what is the current status of this PR? Will you be able to resolve the current conflicts?

alonfaraj:
hi @zimmy87, the PR is was complete on my side.
I tried to solve the conflicts, hopefully I did it right.
I get some errors about clang-format which I'm not exactly sure why.

rajat2004:
@alonfaraj The clang-format workflow is a new one to enforce the [Coding Guidelines](https://microsoft.github.io/AirSim/coding_guidelines/#clang-format), using an extension or plugin with your editor and running it on the modified files should fix the problem. Install clang-format 11 to avoid any incompatibilities

alonfaraj:
@rajat2004 thanks!
I will try to fix it asap.

alonfaraj:
@zimmy87 Seems like all set now.

Unreal/Plugins/AirSim/Source/DetectionComponent.cpp
```diff
@@ -0,0 +1,189 @@
...
+}
+
+void UDetectionComponent::BeginPlay()
+{
+    Super::BeginPlay();
+    SceneCaptureComponent2D = CastChecked<USceneCaptureComponent2D>(GetAttachParent());
+    ObjectFilter = FObjectFilter();
+}
+
+void UDetectionComponent::TickComponent(float DeltaTime, ELevelTick TickType, FActorComponentTickFunction* ThisTickFunction)
```
rajat2004, line 33: Just thinking about the possible effect on performance, fetching the detections seems to be a heavy computation that will run on every tick. Is it possible to instead do it only when the API is called, i.e in `GetDetections()`? Note that I haven't tested this till now and have no numbers on this currently

If concerned about recomputing inside the same tick due to fast API requests and slow FPS (which probably won't IMO), an easy way could be to keep a boolean to indicate whether it has already been computed in the current tick, and return the cached data if so. If not, compute and set it to false. And the Tick method can just set to false as well.

Unreal/Plugins/AirSim/Source/DetectionComponent.cpp
```diff
@@ -0,0 +1,189 @@
...
+                    Detection.RelativeTransform = FTransform(GetRelativeRotation(Actor->GetActorLocation(), Actor->GetActorRotation()),
+                                                             GetRelativeLocation(Actor->GetActorLocation()));
+                    CachedDetections.Add(Detection);
+                }
+            }
+        }
+    }
+}
+
+TArray<FDetectionInfo> UDetectionComponent::GetDetections() const
```
rajat2004, line 61: Maybe return a const ref here to avoid copy

Unreal/Plugins/AirSim/Source/DetectionComponent.cpp
```diff
@@ -0,0 +1,189 @@
...
+                                                                                    FPlane(1, 0, 0, 0),
+                                                                                    FPlane(0, 1, 0, 0),
+                                                                                    FPlane(0, 0, 0, 1));
+
+    if (SceneCaptureComponent2D->bUseCustomProjectionMatrix) {
+        ProjectionData.ProjectionMatrix = SceneCaptureComponent2D->CustomProjectionMatrix;
+    }
+    else {
+        ProjectionData.ProjectionMatrix = Info.CalculateProjectionMatrix();
+        ;
```
rajat2004, line 121: Stray semicolon

Unreal/Plugins/AirSim/Source/PawnSimApi.cpp
```diff
@@ -316,6 +316,69 @@ void PawnSimApi::reportState(msr::airlib::StateReporter& reporter)
...
+                                             true);
+}
+
+std::vector<PawnSimApi::DetectionInfo> PawnSimApi::getDetections(const std::string& camera_name, ImageCaptureBase::ImageType image_type) const
+{
+    std::vector<msr::airlib::DetectionInfo> result;
+
+    UAirBlueprintLib::RunCommandOnGameThread([this, camera_name, image_type, &result]() {
+        const APIPCamera* camera = getCamera(camera_name);
+        const TArray<FDetectionInfo> detections = camera->getDetectionComponent(image_type, false)->GetDetections();
```
rajat2004, line 358: Same, return type as const ref will avoid a copy

Unreal/Plugins/AirSim/Source/DetectionComponent.cpp
alonfaraj: you right, thanks.

Unreal/Plugins/AirSim/Source/DetectionComponent.cpp
alonfaraj: Thanks for your help and reviews @rajat2004!

Yes that might be a good idea. Thought about similar idea as well.

I was thinking to wait for the maintainers reviews and see if they interested to merge this branch.
It's becoming a pretty massive branch, so be easier if we would have the same baseline.
I think opening another branch for optimization/performance will be more efficient and the guys can bring their own insights/ideas about it.

alonfaraj: Hi again @rajat2004, I was rethinking about your suggestion.
I moved the compute detections logic to `GetDetections()` so it will be trigger only by API request.
Thanks again!

zimmy87 (changes requested):
Mostly comments on naming guidelines

alonfaraj:
@zimmy87 Thanks for the review! Hopefully I didn't miss anything

zimmy87 (approved):
latest revision looks good to me; will move ahead with merge once all checks pass

jonyMarino:
Hi, @alonfaraj! Congratulations on this merged Pull Request. You are in the top 5 AirSim contributors! However, this contribution would have a much greater impact if it had associated documentation. Can you create a new PR with documentation?

alonfaraj:
@jonyMarino  Thank you!
Sure, no problem.

rajat2004:
Should have mentioned this earlier, the API has a `image_type` argument, but the detection component is only present for Scene in [PIPCamera.cpp](https://github.com/microsoft/AirSim/blob/4206ac57affa77f2071629018899945c88c4f3ff/Unreal/Plugins/AirSim/Source/PIPCamera.cpp#L74), and different image types also don't really make sense since all the images will be same for the detections. Should the image_type arg be kept at all?

alonfaraj:
@rajat2004 you right, this PR currently support only `Scene` as mentioned in the first post above and it's easy to extend it for all other types. I thought it would be mostly relevant to `Scene` so didn't add it for all other types.

I added the `image_type` argument because different image types can have different parameters such as resolution, FOV etc. which can lead to different detection results.

I'm wondering if add it to all other types is necessary or remove the `image_type` argument.

rajat2004:
Yeah, makes sense to have different detections for each image type as well. Since the API already has the image_type arg, adding support for other types will be good

LIU-Xueming:
Hi, @alonfaraj 
Thank you very much for your work, it is very helpful to me！

But Could you please explain these parameters in detail? ,such as 
geoPoint = GeoPoint()
relative_pose = Pose() 
 
I am confused about these two parameters 

Thanks a lot！

alonfaraj:
Thank you @LIU-Xueming,

`geoPoint` is the geographical coordinate of the detected object. It's relevant in case you specify `OriginGeopoint` you can read more about it [here](https://microsoft.github.io/AirSim/settings/#origingeopoint).
`relative_pose` is position and orientation of the detected object, relative to the camera which generate the detection.

zohaibjan:
Hi, @alonfaraj

I have a quick question. How will you accommodate for occluded objects?. Is there a way to determine whether the object is occluded or not ?. I potentially want to exclude the bounding box of an object if it is occluded. 

Thank you.

PR #3475, opened 2021-03-16
Add moveByVelocityZBodyFrame

### LINKED ISSUE #3476: moveByVelocityZ in body frame
### Overview:
I experienced a small problem with function `moveByVelocityBodyFrame` (when used to move in X-Y plane): multiple calls, separated by a small amount of time (for example, the time to process and analyze an image) leads the UAV to move downwards gradually.

A small video illustrating this problem is available here: [https://youtu.be/iiWdGzhMKqA](https://youtu.be/iiWdGzhMKqA)

The use of `moveByVelocityZ` solved this problem, but unfortunately its associated `moveByVelocityZBodyFrame` does not exist, so I propose to add it.

### Smaller Details: 

The implementation would be (almost) exactly the same as the `moveByVelocityBodyFrame` added in #3169.

Usage: `moveByVelocityZBodyFrameAsync(vx, vy, z, duration, drivetrain (optional), yaw_mode (optional), vehicle_name (optional)).`

### Nature of Request:  

- Addition

To be consistent in the methods proposed in the API for moving using velocity.

Indeed, the function `moveByVelocity` has its associated `moveByVelocityBodyFrame`. It is not the case for the function `moveByVelocityZ`.

meurissemax:
Fixes: #3476 

## About
To be consistent, I propose to add function `moveByVelocityZBodyFrame` to Multirotor API.

Indeed, the function `moveByVelocity` has its associated `moveByVelocityBodyFrame`. It is not the case for the function `moveByVelocityZ`.

More over, I experienced a small problem with function `moveByVelocityBodyFrame` (when used to move in X-Y plane): multiple calls, separated by a small amount of time (for example, the time to process and analyze an image) leads the UAV to move downwards gradually.

A small video illustrating this problem is available here: [https://youtu.be/iiWdGzhMKqA](https://youtu.be/iiWdGzhMKqA).

The use of `moveByVelocityZ` solved this problem, but unfortunately its associated `moveByVelocityZBodyFrame` does not exist, so I propose to add it.

The implementation is (almost) exactly the same as this PR: #3169.

Usage: `moveByVelocityZBodyFrameAsync(vx, vy, z, duration, drivetrain (optional), yaw_mode (optional), vehicle_name (optional))`.

## How Has This Been Tested?
I used the modified version of the AirSim plugin in my Python project. I tested the function in multiple environments by simply moving the UAV using `moveByVelocityZBodyFrame` . The UAV successfully moves by velocity in its direction's X-Y frame.

A use case of my project is available here: [https://youtu.be/mODIlYtCzP8?t=15](https://youtu.be/mODIlYtCzP8?t=15).

In this example, the drone moves using `moveByVelocityZBodyFrame`. It can be seen that the drone stops very briefly from time to time to perform calculations and still keeps a constant altitude during its movement.

/

### DIFF
--- a/AirLib/src/vehicles/multirotor/api/MultirotorApiBase.cpp
+++ b/AirLib/src/vehicles/multirotor/api/MultirotorApiBase.cpp
@@ -75,13 +75,14 @@ bool MultirotorApiBase::goHome(float timeout_sec)
 bool MultirotorApiBase::moveByVelocityBodyFrame(float vx, float vy, float vz, float duration, DrivetrainType drivetrain, const YawMode& yaw_mode)
 {
     SingleTaskCall lock(this);
+
+    if (duration <= 0)
+        return true;
+
     float pitch, roll, yaw;
     VectorMath::toEulerianAngle(getKinematicsEstimated().pose.orientation, pitch, roll, yaw);
     float vx_new = (vx * (float)std::cos(yaw)) - (vy * (float)std::sin(yaw));
     float vy_new = (vx * (float)std::sin(yaw)) + (vy * (float)std::cos(yaw));
-    
-    if (duration <= 0)
-        return true;
     
     YawMode adj_yaw_mode(yaw_mode.is_rate, yaw_mode.yaw_or_rate);
     adjustYaw(vx_new, vy_new, drivetrain, adj_yaw_mode);
@@ -95,13 +96,14 @@ bool MultirotorApiBase::moveByVelocityBodyFrame(float vx, float vy, float vz, fl
 bool MultirotorApiBase::moveByVelocityZBodyFrame(float vx, float vy, float z, float duration, DrivetrainType drivetrain, const YawMode& yaw_mode)
 {
     SingleTaskCall lock(this);
+
+    if (duration <= 0)
+        return true;
+
     float pitch, roll, yaw;
     VectorMath::toEulerianAngle(getKinematicsEstimated().pose.orientation, pitch, roll, yaw);
     float vx_new = (vx * (float)std::cos(yaw)) - (vy * (float)std::sin(yaw));
     float vy_new = (vx * (float)std::sin(yaw)) + (vy * (float)std::cos(yaw));
-
-    if (duration <= 0)
-        return true;
 
     YawMode adj_yaw_mode(yaw_mode.is_rate, yaw_mode.yaw_or_rate);
     adjustYaw(vx_new, vy_new, drivetrain, adj_yaw_mode);
--- a/AirLib/src/vehicles/multirotor/api/MultirotorApiBase.cpp
+++ b/AirLib/src/vehicles/multirotor/api/MultirotorApiBase.cpp
@@ -91,6 +91,27 @@ bool MultirotorApiBase::moveByVelocityBodyFrame(float vx, float vy, float vz, fl
         return false; //keep moving until timeout
         }, duration).isTimeout();
 }
+
+bool MultirotorApiBase::moveByVelocityZBodyFrame(float vx, float vy, float z, float duration, DrivetrainType drivetrain, const YawMode& yaw_mode)
+{
+    SingleTaskCall lock(this);
+    float pitch, roll, yaw;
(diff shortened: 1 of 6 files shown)


PythonClient/airsim/client.py
```diff
@@ -932,7 +932,10 @@ def moveByVelocityBodyFrameAsync(self, vx, vy, vz, duration, drivetrain = Drivet
             msgpackrpc.future.Future: future. call .join() to wait for method to finish. Example: client.METHOD().join()
         """
         return self.client.call_async('moveByVelocityBodyFrame', vx, vy, vz, duration, drivetrain, yaw_mode, vehicle_name)
-        
+
+    def moveByVelocityZBodyFrameAsync(self, vx, vy, z, duration, drivetrain = DrivetrainType.MaxDegreeOfFreedom, yaw_mode = YawMode(), vehicle_name = ''):
+        return self.client.call_async('moveByVelocityZBodyFrame', vx, vy, z, duration, drivetrain, yaw_mode, vehicle_name)
```
rajat2004, line 937: docstring should also be added. Is `z` relative to the body or global coordinates?

rajat2004 (commented):
The implementation is almost exactly the same as `moveByVelocityBodyFrame`, so maybe it can be extracted out to a common method? I can't think of any good way to do this currently, if anyone has any ideas that would be great, else can be left as is

meurissemax:
Hello @rajat2004 👋 , thank you for your review!

As I mentioned, the implementation is almost exactly the same as `moveByVelocityBodyFrame`. More precisely, they have in common two lines of code that transform the velocities in the world frame to the corresponding values in the drone frame.

This transformation can certainly be computed in another function, but to be honest, I don't known how and where to add properly such a function, so I prefer to leave it as it is for now.

I create a new commit with modifications according to your advice.

AirLib/src/vehicles/multirotor/api/MultirotorApiBase.cpp
```diff
@@ -91,6 +91,27 @@ bool MultirotorApiBase::moveByVelocityBodyFrame(float vx, float vy, float vz, fl
...
+bool MultirotorApiBase::moveByVelocityZBodyFrame(float vx, float vy, float z, float duration, DrivetrainType drivetrain, const YawMode& yaw_mode)
+{
+    SingleTaskCall lock(this);
+    float pitch, roll, yaw;
+    VectorMath::toEulerianAngle(getKinematicsEstimated().pose.orientation, pitch, roll, yaw);
+    float vx_new = (vx * (float)std::cos(yaw)) - (vy * (float)std::sin(yaw));
+    float vy_new = (vx * (float)std::sin(yaw)) + (vy * (float)std::cos(yaw));
+
+    if (duration <= 0)
+        return true;
```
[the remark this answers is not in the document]
meurissemax: This check can indeed be move to the beginning. I also moved the corresponding check in the `moveByVelocityBodyFrame` to be consistent.

If I understand correctly, this function (and all others) returns a boolean indicating whether the execution of action is complete. If the duration is less than (or equal to) 0, the drone has nothing to do and the function immediately returns that the execution of the action is complete (`true`).

PythonClient/airsim/client.py
meurissemax: I have added docstring. `z` is the desired Z value in local NED frame of the vehicle.

rajat2004: Adding to this, `z` + current_alt (at the time of API call) is the altitude that the vehicle will try to attain in `duration` seconds right?

rajat2004:
@meurissemax Yeah, the duplication is pretty small so shouldn't be a problem ig

PythonClient/airsim/client.py
meurissemax: Yes, I think so. Functions `moveByVelocityZBodyFrame` (and `moveByVelocityZ`) move the drone in X-Y plane according to velocities `vx` and `vy` while reaching (or maintaining) altitude `z`.

jonyMarino:
Hi @meurissemax! Thanks for your contribution!

PR #3514, opened 2021-03-23
Docs: Fix step numbering in Custom env page

rajat2004:
Fixes: #    

## About

The numbering in steps in https://microsoft.github.io/AirSim/unreal_custenv/ is wrong, it goes from 1-6 and then again 1-6 which can cause confusion

image

## How Has This Been Tested?

`mkdocs serve`

image

image

Added above


jonyMarino:
Thanks @rajat2004!

PR #3527, opened 2021-03-25
Add iter to unpack Vector3r, Quaternionr and Pose

laurelkeys:
## About
Implements `__iter__` for `Vector3r`, `Quaternionr` and `Pose` so that the `*` operator can be used for [unpacking](https://www.python.org/dev/peps/pep-0448/#abstract).

This leads to some "quality of life" improvements since it removes the need to call `.x_val`, `.y_val`, etc. separately when we want to extract the components from a vector or quaternion.

Currently, this unpacking behavior can only be achieved by first calling `.to_numpy_array()`. However, this leads to the following error when used with `moveToPositionAsync`:
```
AttributeError: 'numpy.float32' object has no attribute 'to_msgpack'
```

## How Has This Been Tested?
The following test program illustrates the changes this introduces:
```python
import time
import airsim

def main():
    client.confirmConnection()
    client.enableApiControl(True)
    client.armDisarm(True)

    try:
        while True:
            pose = client.simGetVehiclePose()
            position, orientation = pose  # unpacks Pose into a Vector3r and a Quaternionr
            new_position = position + airsim.Vector3r(0, 0, -2)

            print("1:", position)   # prints a Vector3r
            print("2:", *position)  # unpacks the Vector3r's x_val, y_val and z_val components
            x, y, z = position      # ditto
            print("3:", x, y, z)    # this has the same result as the previous call to print

            # thus, we can now get the same result of doing:
            print("4:", orientation.x_val, orientation.y_val, orientation.z_val, orientation.w_val)
            # by simply using the * operator:
            print("5:", *orientation)

            # furthermore, this allows calling functions that expect separate x, y, z values:
            client.moveToPositionAsync(*new_position, velocity=2).join()

            # which currently do not work if we try to unpack the return of to_numpy_array,
            # with "AttributeError: 'numpy.float32' object has no attribute 'to_msgpack'":
            # # client.moveToPositionAsync(*new_position.to_numpy_array(), velocity=2).join()

            client.hoverAsync().join()
            time.sleep(1)
            print()
    except KeyboardInterrupt:
        client.reset()

if __name__ == "__main__":
    main()
```

## Screenshots:
Output from the test program listed above:
image

### DIFF
--- a/PythonClient/airsim/types.py
+++ b/PythonClient/airsim/types.py
@@ -136,6 +136,8 @@ class Vector3r(MsgpackMixin):
     def to_numpy_array(self):
         return np.array([self.x_val, self.y_val, self.z_val], dtype=np.float32)
 
+    def __iter__(self):
+        return iter((self.x_val, self.y_val, self.z_val))
 
 class Quaternionr(MsgpackMixin):
     w_val = 0.0
@@ -226,14 +228,16 @@ class Quaternionr(MsgpackMixin):
     def to_numpy_array(self):
         return np.array([self.x_val, self.y_val, self.z_val, self.w_val], dtype=np.float32)
 
+    def __iter__(self):
+        return iter((self.x_val, self.y_val, self.z_val, self.w_val))
 
 class Pose(MsgpackMixin):
     position = Vector3r()
     orientation = Quaternionr()
 
     def __init__(self, position_val = None, orientation_val = None):
-        position_val = position_val if position_val != None else Vector3r()
-        orientation_val = orientation_val if orientation_val != None else Quaternionr()
+        position_val = position_val if position_val is not None else Vector3r()
+        orientation_val = orientation_val if orientation_val is not None else Quaternionr()
         self.position = position_val
         self.orientation = orientation_val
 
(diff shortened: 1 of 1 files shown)


laurelkeys:
Also, implementing `__iter__` could be used as a replacement for (or alternative to) the `to_lists` method of [`AngleRateControllerGains`](https://github.com/microsoft/AirSim/blob/2c07d985f7a42816b1f385392843a39392e1d777/PythonClient/airsim/types.py#L484), [`AngleLevelControllerGains`](https://github.com/microsoft/AirSim/blob/2c07d985f7a42816b1f385392843a39392e1d777/PythonClient/airsim/types.py#L503), [`VelocityControllerGains`](https://github.com/microsoft/AirSim/blob/2c07d985f7a42816b1f385392843a39392e1d777/PythonClient/airsim/types.py#L522) and [`PositionControllerGains`](https://github.com/microsoft/AirSim/blob/2c07d985f7a42816b1f385392843a39392e1d777/PythonClient/airsim/types.py#L541).

zimmy87:
Thanks for the contribution @laurelkeys! Tested this locally and it works for me so I am moving ahead with merging

PR #3545, opened 2021-03-30
Update moveToPosition to use setpoint control message instead of velocity control message

### LINKED ISSUE #1677: Incorrect behavior of MoveToPositionAsync #1643，how to solve it?

zimmy87:
Fixes: #1677

## About
Override moveToPosition() implementation in MavLinkMultirotorApi.hpp to use MavLink setpoint control messages for movement instead of velocity control messages. In place of velocity control messages, the MPC_XY_CRUISE parameter is set to the inputted velocity for the duration of the moveToPosition() call and then reset to its original value at the end of the call.

## How Has This Been Tested?
tested with hello_drone3.py script in description of #1677 

### DIFF
--- a/AirLib/include/vehicles/multirotor/firmwares/mavlink/MavLinkMultirotorApi.hpp
+++ b/AirLib/include/vehicles/multirotor/firmwares/mavlink/MavLinkMultirotorApi.hpp
@@ -382,15 +382,17 @@ public: //methods
         unused(lookahead);
         unused(drivetrain);
 
-        // save current cruise velocity value
+        // save current manual, cruise, and max velocity parameters
         bool result = false;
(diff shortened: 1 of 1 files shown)


jonyMarino:
Thanks, David! Let's see if someone wants to give it a try. @ahmed-elsaharti?

ahmed-elsaharti:
Testing ASAP and reporting back with results 👍

jonyMarino:
Thanks, David! You did an incredible job!

PR #3561, opened 2021-04-02
Add cross-compilation section in build_linux.md

evroon:
## About
Cross compilation is possible but requires some instructions. Specifically, the AirLib libraries have to be build from a Linux environment.

## How Has This Been Tested?
I did this step-by-step (from Windows 10, for Ubuntu 20.04) and for me it works.

### DIFF
--- a/docs/build_linux.md
+++ b/docs/build_linux.md
@@ -56,6 +56,9 @@ cd AirSim
 # use ./build.sh --debug to build in debug mode
 ```
 
+## Cross-compile AirSim from Windows for Linux
+You can cross compile for Linux from a Windows machine. First install clang as instructed [here](https://docs.unrealengine.com/latest/INT/Platforms/Linux/GettingStarted/index.html). Then, build AirSim from Bash for Windows, as instructed above, such that the Linux .lib files are created. Subsequently, build AirSim again but this time for Windows (as instructed [here](build_windows.md)) such that you can open the project in UE4. After building UE4 for Windows, open the project in Unreal Engine. Under Project Settings, go to `Supported Platforms` and enable `Linux` by clicking the checkbox. Now, go to `Plugins` and search for `VR`. Disable `Oculus VR` and `SteamVR` as they will result in build errors. Now, you can start packaging the project by going to File &rarr; Package Projct &rarr; Linux &rarr; Linux.
+
 ### Build Unreal Environment
 
 Finally, you will need an Unreal project that hosts the environment for your vehicles. AirSim comes with a built-in "Blocks Environment" which you can use, or you can create your own. Please see [setting up Unreal Environment](unreal_proj.md) if you'd like to setup your own environment.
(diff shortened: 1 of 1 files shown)


ahmed-elsaharti:
I tried to make this work for a good 5 hours but it just wouldn't (or I'm messing something up).
The first issue I could see was that on building AirSim on Windows, all the Linux (WSL2) compilation results (all the .a components of AirLib) would get deleted. I'm assuming its something that the build.cmd script does?

Moreover, even when trying to bypass this error by grabbing a WSL2 compiled AirSim plugins folder and placing it into the UE4 project folder, although things end up packaging just fine through UE4, when running the resulting .sh binary the environment works fine but the drone is stagnant mid-air with no physics or prop movement. The car simulation does work though, so I'm assuming its something within AirLib that breaks while doing this since car-mode isn't dependent on AirLib or AirSim's physics (AFAIK).

evroon:
The first problem I did not encounter as far as I know, but you could also build for linux and copy the 3 or 4 lib files to a temporary location and then build AirLib for windows and paste the files in the original location.

The second issue is indeed something I oversaw, I only noticed it today by accident and I don't know exactly why it happens, but has probably indeed to do with AirLib. Sorry about that.

docs/build_linux.md
```diff
@@ -56,6 +56,9 @@ cd AirSim
 # use ./build.sh --debug to build in debug mode
 ```
 
+## Cross-compile AirSim from Windows for Linux
+You can cross compile for Linux from a Windows machine. First install clang as instructed [here](https://docs.unrealengine.com/latest/INT/Platforms/Linux/GettingStarted/index.html). Then, build AirSim from Bash for Windows, as instructed above, such that the Linux .lib files are created. Subsequently, build AirSim again but this time for Windows (as instructed [here](build_windows.md)) such that you can open the project in UE4. After building UE4 for Windows, open the project in Unreal Engine. Under Project Settings, go to `Supported Platforms` and enable `Linux` by clicking the checkbox. Now, go to `Plugins` and search for `VR`. Disable `Oculus VR` and `SteamVR` as they will result in build errors. Now, you can start packaging the project by going to File &rarr; Package Projct &rarr; Linux &rarr; Linux.
```
zimmy87, line 60: A couple notes here:

1. Instead of Bash for Windows, I think it would be better to specify Windows Subsystem for Linux (WSL). 
2. Where is the AirSim repo that is being built located? When I build from WSL out of a repo located in my windows directory structure, I get build errors. If I instead build out of the WSL directory structure, I get build errors when I build from Windows. Is the expectation to copy binaries from a repo located in WSL to one located in windows?

zimmy87 (changes requested):
I think we need extra clarification in this doc. I personally ran into some confusion and wasn't able to get this scenario working locally

docs/build_linux.md
evroon: WSL is indeed the correct/current term.
I think I used `/mnt/c` (or another drive) to build AirSim from in WSL. But this method does not work in its current form (atleast AirLib is not working), as in the comments above. So, maybe I should close this PR.

PR #3596, opened 2021-04-15
Add new build configuration "--RelWithDebInfo" which makes it easier to debug

lovettchris:
Fixes: #    
makes it easier to debug AirSim inside Unreal environment.  This change is a subset of PR 3546 that I'm pulling out to make this change independently.

## About
Unreal environment wants Release bits (takes more work to make it link against debug airsim bits) and Release bits are hard to debug because of program optimization.  So RelWithDebInfo is still release but without optimization, so it can be used in place of Release bits in an Unreal environment and is then easier to debug.

## How Has This Been Tested?
run command line build.cmd with --RelWithDebInfo


zimmy87:
tested that release and debug configurations did not regress with this change, so I am moving ahead with merge

PR #3604, opened 2021-04-16
Fix LogViewer bugs

lovettchris:
Fixes: #    

## About
- Optimize layout and avoid WPF performance bugs.
- Fix loading of .ulg log files.
- Add mavlink2 support.

This is a view of a .mavlink log file created by AirSim while doing take off and landing, the top line is the "actuator controls" signal coming from PX4, and the bottom line is the simulated GPS altitude.

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -386,3 +386,4 @@ build_docs/
 
 # api docs
 PythonClient/docs/_build
+/LogViewer/LogViewer/publish.cmd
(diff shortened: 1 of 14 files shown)


zimmy87:
Tested that the LogViewer works with a .mavlink log we collected in a AirSim+PX4 on WSL scenario, so we are moving ahead with merging this

PR #3626, opened 2021-04-23
External physics engine

jonyMarino:
## About

This contribution helps in a recurring issue that is to use the AirSim vehicle with all its perception capabilities and integrations together with an external physics engine or flight dynamics model or driven by another simulator. Together with the possibility of being able to change the visualization of the vehicle from AirSim to another particular one (another mesh), this makes AirSim a very flexible simulator to create your own models (see for example [Fixedwing-Airsim](https://github.com/AOS55/Fixedwing-Airsim)).
This also helps in another recurring theme that is to reproduce a trajectory, already preset or saved, in the Airsim vehicle.

## How Has This Been Tested?

Running external_physics_engine.py already present in this PR.


zimmy87 (changes requested):
couple style/naming comments

zimmy87 (approved):

jonyMarino:
Thanks for the review @zimmy87!

nikitabeebe:
Has there been any more progress towards an external engine model? Looking to use it for a fixed-wing plane physics (in lieu of JSBSim physics like the link of FixedWing-AirSim posted above)

jonyMarino:
Hi @nikitabeebe! Which external physics engine are you looking for to use?

nikitabeebe:
@jonyMarino I'm still not sure. I'm looking into the different options, aiming to try to sync it with AirSim over the next month or so, as I've read about some attempts using ArduPilot as well as JSBSim, but haven't seen too much success yet. I was hoping to avoid rewritting AirSim's FastEnginePhysics entirely. Do you have suggestions? (My only main criteria is it models a fixed wing UAV for GPS and IMU sensory processing)

PR #3656, opened 2021-05-04
ArduPilot: Reduce data copying, stack memory usage

rajat2004:
Fixes: #    

## About

Small cleanups to improve performance, memory usage

- Remove `char buf[65000]` being allocated in each `update()` call
- Use `ostringstream` for encoding all data, remove `snprintf`. This removes 1 copy in `snprintf` for the `%s` specifier with `oss.str().c_str()` which creates a temporary string and then a char*. Note that there's still 1 copy for get the char*, explicitly mentioned, this is because there's no way (at least that I could find) to get the char* without a temporary string.
- Reduces the precision of floats to 7 from earlier 12 which is much more appropriate.

TODO
- Handle sending too long strings
- Check reset implementation

## How Has This Been Tested?

Tested both Copter and Rover with ArduPilot

### DIFF
--- a/AirLib/include/vehicles/car/firmwares/ardurover/ArduRoverApi.hpp
+++ b/AirLib/include/vehicles/car/firmwares/ardurover/ArduRoverApi.hpp
@@ -21,10 +21,6 @@
 #include "UdpSocket.hpp"
 
 #include <sstream>
-#include <iostream>
-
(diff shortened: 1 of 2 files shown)


jonyMarino:
Thanks, Rajat!

PR #3666, opened 2021-05-07
add logic for setting visibility of ExponentialHeightFog actor

### LINKED ISSUE #3585: ExponentialHeightFog
- AirSim Version/#commit: 46a9b3a53811175e198e3608316fa4c983f7c53c
- UE/Unity version: Unreal Engine 4.25.4
- autopilot version: n/a
- OS Version: Ubuntu 18.04.5

Is there a way to prevent Airsim from spawning this element in Unreal Engine?  Currently, I have to manually reduce the Fog level to 0 (or disable its Visibility parameter) every time I start a new session, but this is cumbersome, prone to forgetting,  and prevents automating my workflow.  The python script that I run does not make any calls to simEnableWeather, so I don't know why Airsim spawns this element at all.

Screenshot from 2021-04-12 17-20-34

zimmy87:
Fixes: #3585    

## About
set visibility of the ExponentialHeightFog actor to false when it is spawned inside WeatherActor and then update the visibility to true once a weather parameter has been altered

## How Has This Been Tested?
weather parameters were set via the F10 weather menu as well as calling the client.simEnableWeather() and client.simSetWeatherParameter() APIs from python


PythonClient/multirotor/set_fog.py
```diff
@@ -0,0 +1,34 @@
...
+client = airsim.MultirotorClient()
+client.confirmConnection()
+client.enableApiControl(True)
+
+client.armDisarm(True)
+
+client.simEnableWeather(True)
+
+print("Setting fog to 25%") # NED
+client.simSetWeatherParameter(7, 0.25)
```
rajat2004, line 14: More readable to use `airsim.WeatherParameter.Fog` rather than direct enum value. Also, the `NED` comment above is not applicable here

PythonClient/multirotor/set_fog.py
```diff
@@ -0,0 +1,34 @@
...
+    print("taking off...")
+    client.takeoffAsync().join()
+else:
+    print("already flying...")
+    client.hoverAsync().join()
+
+time.sleep(5)
+
+print("Setting fog to 50%") # NED
+client.simSetWeatherParameter(7, 0.5)
```
rajat2004, line 28: Same, `airsim.WeatherParameter.Fog` and comment

PythonClient/multirotor/set_fog.py
```diff
@@ -0,0 +1,34 @@
...
+
+time.sleep(5)
+
+# Set wind to 0
+print("Resetting fog to 0%")
+client.simSetWeatherParameter(7, 0)
```
rajat2004, line 34: `airsim.WeatherParameter.Fog` and not applicable comment about wind

Unreal/Plugins/AirSim/Source/Weather/WeatherLib.cpp
```diff
@@ -5,6 +5,7 @@
 #include "Blueprint/UserWidget.h"
 #include "Blueprint/WidgetBlueprintLibrary.h"
 
+AExponentialHeightFog* UWeatherLib::weather_fog_ = nullptr;
```
rajat2004, line 8: Can this initial value be set in the header file itself rather than here? That's cleaner IMO

jonyMarino: Hi Rajat. That would make that variable constant I think, and it is not what we want.

rajat2004: Yeah, you're correct, initializing a static var in header file doesn't work unless it's a `const` var, or basic data types such as `int`, `char`, etc.  https://stackoverflow.com/questions/185844/how-to-initialize-private-static-members-in-c/185848 has more details as well. The entire matter's a bit confusing TBH, but this is the correct way to go about using this. Also tested compiling with initialization in the header file, and gives a compile error -

```
error: non-const static data member must be initialized out of line
        static AExponentialHeightFog* weather_fog_ = nullptr;
```

Sorry for not digging deeper into this before commenting!

PythonClient/multirotor/set_fog.py
zimmy87: fixed

PythonClient/multirotor/set_fog.py
zimmy87: fixed

PythonClient/multirotor/set_fog.py
zimmy87: fixed

Unreal/Plugins/AirSim/Source/Weather/WeatherLib.cpp
zimmy87: Yea, it took a decent amount to searching to figure out how to properly initialize a private static member. I think we may've even been looking at the same stackoverflow.com question. I'm gonna leave this as is because I don't think we have any better options at the moment.

PR #3671, opened 2021-05-07
update PX4 info on ControlPortLocal and ControlPortRemote

lovettchris:
Fixes: #    
doc fix

## About
last PR broke ControlPort into 2 settings, ControlPortLocal and ControlPortRemote.  This fixes the docs for that.

### DIFF
--- a/docs/px4_multi_vehicle.md
+++ b/docs/px4_multi_vehicle.md
@@ -33,7 +33,8 @@ However, the provided script does not let us view the PX4 console. If you want t
                 "UseSerial": false,
                 "UseTcp": true,
                 "TcpPort": 4560,
-                "ControlPort": 14580,
+                "ControlPortLocal": 14540,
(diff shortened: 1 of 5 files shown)


zimmy87:
Can we add a blurb to the first paragraph in [https://github.com/microsoft/AirSim/blob/master/docs/px4_sitl.md](https://github.com/microsoft/AirSim/blob/master/docs/px4_sitl.md) mentioning that WSL2 is currently more stable than Cygwin and thus is the recommended way of running PX4 SITL?

lovettchris:
@zimmy87, I think we'd need more data on that.  We have one test we know of that ran into trouble with cygwin?  I think we'd need more scientific testing on this to really come to a solid conclusion.

lovettchris:
I'm going ahead with this merge for now, since we need to fix the docs.  We can add more later.

PR #3680, opened 2021-05-10
Fix bugs in px4 .ulg viewing, and make tooltip avoid the closebox 

lovettchris:
Fixes: #    
LogViewer was not displaying data loaded from px4 .ulg log file.

## About
LogViewer

## How Has This Been Tested?
LogViewer shows data in .ulg log file.

image


zimmy87:
tested locally with a .mavlink and .ulg log and both appear to display correctly, so I'm moving ahead with merging this

PR #3684, opened 2021-05-11
Fix typo (collison -> collision)

laurelkeys:
## About
Fix a typo in the Python API (replace "collison" with "collision").


jonyMarino:
Thanks @laurelkeys!

PR #3692, opened 2021-05-13
add workflow for checking clang formatting

lovettchris:
Fixes: #    
clang-formatting of all airsim code.

## About
Adds a new github workflow to check clang-formatting of code and did one big formatting pass to make it pass.

This is how all the source files were formatted in one batch process:
```shell
git clean -dfx
dir /s /b *.h *.hxx *.hpp *.cxx *.c *.cpp > list.txt
sed -e "s/^/clang-format -i /" list.txt  > format.cmd
format.cmd
```

The git clean step removed external code dependencies like `Eigen` and `rpclib`.

## How Has This Been Tested?
Hit clang-formatting bug, see [open issue](https://github.com/jidicula/clang-format-action/issues/40).
Found a workaround - which is disable clang-formatting for the mavlink headers.


rajat2004:
Getting done with all the formatting in 1 swoop makes sense, but I feel some things need to be handled. 

- Needs to be added to the documentation, to know the exact clang format version to be used. Different versions could have slightly different output and cause problems. Along with installation instructions if possible. Editor integrations will be useful after this (need to do this myself :sweat_smile: )

- Should external code be formatted? I personally don't think so, for eg. [bitmap_image.hpp](https://github.com/microsoft/AirSim/blob/master/AirLib/include/common/common_utils/bitmap_image.hpp) (btw this file adds 15k changes itself), [json.hpp](https://github.com/microsoft/AirSim/blob/master/AirLib/include/common/common_utils/json.hpp), similarly in MavlinkCom as well.

Clang format doesn't seem to have the option of ignoring files, but there's a tool which might be useful for this case - https://github.com/Sarcasm/run-clang-format. Can be used in CI as well

lovettchris:
@rajat2004, yep, I've reverted those files and added `// clang-format off` at the top of those files to disable future clang formatting in those files, this will make it easier to merge updates from those original libraries.  This also resolves the clang-formatting bug regarding bogus Objective-C classification.  So good call, thanks!

rajat2004:
@lovettchris Thanks for that! There are a few more files - [ctpl_stl.h](https://github.com/microsoft/AirSim/blob/master/AirLib/include/common/common_utils/ctpl_stl.h), [optional.hpp](https://github.com/microsoft/AirSim/blob/master/AirLib/include/common/common_utils/optional.hpp), [pretty_print.hpp](https://github.com/microsoft/AirSim/blob/master/AirLib/include/common/common_utils/prettyprint.hpp) and similar in MavlinkCom

Since there are only a few of these files, the comment should be fine, otherwise might need to use the clang-format-ingore tool. Mostly running the format tool on the entire AirSim folder won't be done locally, otherwise the Eigen, rpclib would also get picked up

lovettchris:
@rajat2004, I reverted changes to ctpl_stl.h, optional.hpp, pretty_print.hpp and added documentation to `coding_guidelines.md`.  And I'm expecting from here on out folks just do automatic Visual Studio clang-formatting of individual files they edit.

PR #3697, opened 2021-05-16
Docs: Update use_precompiled.md

### LINKED ISSUE #3683: Location of run.bat script for CityEnviron
### What's your question?
I read the tip on environment setup in case of inadequate GPU subsystem. The short [tutorial](https://microsoft.github.io/AirSim/use_precompiled/#dont-have-good-gpu) references a script called `run.bat` under the parent folder of each environment. However, I coudn't find it. In case this file lies at an older version, is it safe to use that version? For example, I found the [version](https://github.com/microsoft/AirSim/releases/tag/v1.1.6) that first announced the City Environment and there is a disclaimer ("Note: City environment is under heavy development with more updates coming soon.").

### Include context on what you are trying to achieve
I'm trying to make City Environment run with less FPS, lower resolution (ex. 640x480 like the tutorial), because my GPU doesn't support the default setting. Is there a setting for lower FPS regarding the City Environment?

### System information
AirSim version 1.4
Microsoft Windows 10 Education Version 10.0.19042 Build 19042
Unreal Engine 4.26.1

rajat2004:
Fixes: #3683    

## About

Adds some info about using the binaries including limiting the fps.

## How Has This Been Tested?

The API has been tested with a binary of the latest master, docs using `mkdocs server`

### DIFF
--- a/docs/use_precompiled.md
+++ b/docs/use_precompiled.md
@@ -4,25 +4,48 @@ You can simply download precompiled binaries and run to get started immediately.
 
 ### Unreal Engine
 
-**Windows, Linux**: Download the binaries for the environment of your choice from the [latest release](https://github.com/Microsoft/AirSim/releases). Some pre-compiled environment binaries may include multiple files (i.e. City.zip.001, City.zip.002). Make sure to download both files before starting the environment. You may need to concatenate both files to unzip them.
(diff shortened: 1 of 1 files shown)


jonyMarino:
Thanks again!

PR #3699, opened 2021-05-16
Make falling leaves visible in depth and segmentation

saihv:
Partially addresses #3695.

### About
Enable depth passes for 'falling leaf' particles to show up in depth and segmentation images. 
The same method can be followed for snow, in case someone is interested in seeing snow particles in depth etc. 

The fix is two-fold:

1. Make the corresponding material opaque instead of translucent

drawing

2. Enable custom depth pass for the particle emitter in the rendering options within Weather Actor. 

image

### How Has This Been Tested?
Enable Weather and falling leaves.

### Screenshots

image


saihv:
Unfortunately the same method doesn't apply for dust/fog, as they are volume materials applied as an additive blend (add material color to actual background color during scene render), and it is unlikely additive blends are used in depth buffer.

Deepakgthomas:
Thank you @saihv for the quick fix!

jonyMarino:
Thanks, Sai! Does this also affect the lidar output?

PR #3705, opened 2021-05-19
Fix incorrect parsing of CameraDirector settings

MinnDevelopment:
## About

I found an issue in the parsing of the `CameraDirector` settings. It seems to use the wrong JSON object for parsing the position and rotation.

## How Has This Been Tested?

The change compiles and fixes the problem. It now correctly parses the position and rotation.


jonyMarino:
Thanks, @MinnDevelopment  for contributing!

PR #3720, opened 2021-05-21
Fix OverrideExisting implementation for Segmentation settings

### LINKED ISSUE #3710: "OverrideExisting" setting in "SegmentationSettings" has opposite effect
## Bug report 

- AirSim Version/#commit: 1.5.0
- UE/Unity version: UE 4.25
- autopilot version: N/A
- OS Version: Windows 10

The "OverrideExisting" setting in "SegmentationSettings" works backwards, i.e. true and false are reversed.

### Settings

```
{
  "SettingsVersion": 1.2,
  "SimMode": "Car",
  "SubWindows": [
    {"WindowID": 1, "CameraName": "front_left", "ImageType": 5, "VehicleName": "", "Visible": true}
  ],
  "SegmentationSettings": {
    "InitMethod": "",
    "MeshNamingMethod": "OwnerName",
    "OverrideExisting": false
  }
}
```

1. Build Vanilla 1.5.0 release
2. Open blocks environment in UE 4.25
3. Use above settings.json
4. Hit play. No need to drive the vehicle.
5. Notice how the semantic segmentation color for all blocks in front of the vehicle is the same

image

6. Stop the simulation
7. Pick one of the blocks in front of the vehicle and change its "CustomDepth Stencil Value" property to a different value

image

8. Hit play
9. Notice how the color of the block did NOT change and how the stencil value gets reset to the initial value of 241, despite the fact that "OverrideExisting" was set to false 
10. Repeat the above, but set the "OverrideExisting" to true. Now the modified stencil value will NOT change when we run the simulation and the block will have a different color in the semantic segmentation subwindow.

image

It seems the setting is interpreted incorrectly (boolean opposite).

### How to fix

The believe the problem is in AirBlueprintLib.h, line 217

image

This should read
```
if(!ignore_existing || ...
```

I tried this fix and it works for me, except for the fact that if you start simulation with the setting set to false, semantic segmentation no longer works. You have to run it first with the setting set to true, and then you can switch between true and false and the behavior is correct. 
Did not check the exact reason yet. Would like to first confirm my interpretation of the expected behavior of "OverrideExisting" is correct.

rajat2004:
Fixes: #3710    

## About

Override means Ignoring, and not the other way around. This PR fixes the inversion, and changes the variable name to be more clear. The Settings default also gets changed to reflect the behaviour

The second commit fixes the segmentation when override is set to false at beginning

## How Has This Been Tested?

If override is true -
segmentation_override_true

Override false with custom stencil value -
final_override_false_working

If override is set to false when running for first time in Editor, the segmentation was blank
segmentation_override_false

Changing the override to true and then false fixed that.

The latest commit fixes that behaviour, and segmentation is correct at startup as well
override_false_initial_working

Pasted above

### DIFF
--- a/AirLib/include/common/AirSimSettings.hpp
+++ b/AirLib/include/common/AirSimSettings.hpp
@@ -946,7 +946,7 @@ namespace airlib
                     //TODO: below exception doesn't actually get raised right now because of issue in Unreal Engine?
                     throw std::invalid_argument(std::string("SegmentationSetting init_method has invalid value in settings_json ") + init_method);
 
-                segmentation_setting.override_existing = json_parent.getBool("OverrideExisting", false);
+                segmentation_setting.override_existing = json_parent.getBool("OverrideExisting", true);
 
                 std::string mesh_naming_method = Utils::toLower(json_parent.getString("MeshNamingMethod", ""));
                 if (mesh_naming_method == "" || mesh_naming_method == "ownername")
--- a/Unreal/Plugins/AirSim/Source/AirBlueprintLib.cpp
+++ b/Unreal/Plugins/AirSim/Source/AirBlueprintLib.cpp
@@ -354,20 +354,20 @@ std::string UAirBlueprintLib::GetMeshName(ALandscapeProxy* mesh)
     return std::string(TCHAR_TO_UTF8(*(mesh->GetName())));
 }
 
-void UAirBlueprintLib::InitializeMeshStencilIDs(bool ignore_existing)
+void UAirBlueprintLib::InitializeMeshStencilIDs(bool override_existing)
 {
     for (TObjectIterator<UStaticMeshComponent> comp; comp; ++comp) {
-        InitializeObjectStencilID(*comp, ignore_existing);
+        InitializeObjectStencilID(*comp, override_existing);
     }
     for (TObjectIterator<USkinnedMeshComponent> comp; comp; ++comp) {
-        InitializeObjectStencilID(*comp, ignore_existing);
+        InitializeObjectStencilID(*comp, override_existing);
     }
     //for (TObjectIterator<UFoliageType> comp; comp; ++comp)
     //{
     //    InitializeObjectStencilID(*comp);
     //}
     for (TObjectIterator<ALandscapeProxy> comp; comp; ++comp) {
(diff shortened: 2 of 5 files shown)


CarloCloet:
This is a clean fix indeed. Tx.

I performed a quick test by manually applying the changes in my local 1.5.0 copy and can confirm the interpretation of OverrideExisting is now correct. 

Two follow-up questions:
1) if we change the default value for "OverrideExisting" from false to true, we will also need to update the settings documentation.
I assume you chose to change the default value in order to not affect existing applications? (where the setting may not have been explicitly set)

2) I am still puzzled why the segmentation view does not work at start-up, even with this fix.

When you use the settings in #3710 and you start the Blocks environment for the first time, you get a solid green subwindow:

semseg_not_working

This is not what I would expect, given the stencil values are different for the different objects in the scene:

semseg_not_working2
(stencil value 241)

semseg_not_working3
(stencil value 148)

When I set "OverrideExisting" to true and re-run the simulation, semantic segmentation works fine:

semseg_working

Setting it back to false and re-running the simulation results in:

semseg_working2

So now semantic segmentation still seems to work, i.e. I get a different behavior compared to the initial run despite the identical settings. Is that what you would expect?

Note that the stencil value is now write-able (white box, as opposed to grey). This may have something to do with it.

rajat2004:
@CarloCloet Sorry for the late reply, have been busy the last few days with some other stuff. 

1. Yeah, the settings documentation needs to be updated. The intention of changing the default value was to keep existing behaviour if the setting hasn't been specified as you mentioned.
2. No idea why it's causing the segmentation image tobe complete blank. Looking at the code, there's an additional value being set -

https://github.com/microsoft/AirSim/blob/4bf6587b19cde568ddd47914e4965b6040345314/Unreal/Plugins/AirSim/Source/AirBlueprintLib.h#L240

Maybe try if modifying this value makes any difference?

rajat2004:
@CarloCloet I've pushed a fix which fixes the problem for me at least, by setting the RenderCustomDepth flag as mentioned above. It would be great if you test it as well!

zimmy87:
Thanks for the contribution @rajat2004! Tested locally with the repro steps in #3710; works as expected, so I am moving ahead with merging.

PR #3721, opened 2021-05-21
Move the private members GaussianMarkovTest.hpp to above public members.

BillyFrcs:
## About

Just change and move private members GaussianMarkovTest.hpp to above public members.

## How Has This Been Tested?

When compiled it's gonna be the same as usual.


jonyMarino:
Hi @BillyFrcs! I think we usually put the private section at the end. Why do you see this necessary?

Issue #3719, opened 2021-05-20
getRotorState Error


KIYONG-KWON:
### What's your question?
// Setting.json
{
"SeeDocsAt": "https://github.com/Microsoft/AirSim/blob/master/docs/settings.md",
"SettingsVersion": 1.5,
"SimMode": "Multirotor",

"Vehicles": {
"Drone1": {
"VehicleType": "SimpleFlight",
"X": 4,
"Y": 0,
"Z": -2,
"Yaw": -180
},
"Drone2": {
"VehicleType": "SimpleFlight",
"X": 8,
"Y": 0,
"Z": 0,
"Yaw": -180
}
}
}

// main.cpp
string droneID = "Drone1";
client.enableApiControl(true, droneID);
client.getRotorStates(droneID); // error on this line , rpc::rpc_error

Cound you advise me why error occurred at this line? other codes to get sensor datas such as gps, imu etc do not make error like this.

#### Context details
Airsim version as the lastest 1.5.0 and unreal version is 4.26.

rajat2004:
You're likely using the PyPi airsim package which needs to be updated, till then, try using the latest master Python library using the instructions in https://github.com/microsoft/AirSim/pull/3611

KIYONG-KWON:
Do I have to update python lib even if i am using c++ API?

rajat2004:
Ah no, sorry for missing that. I'll try to test on my system as well. Just to confirm, both the programs are running on the same system? And is the platform Windows or Linux?

KIYONG-KWON:
Both the programs are running on the same system and window 10 platform. Above code main.cpp is part of HelloDrone example  of Airsim Soluiton.

rajat2004:
So the `getRotorStates` API seems to be working fine for me, my additional code -
```cpp
auto rotor_states = client.getRotorStates();
std::cout << "Rotor States: " << rotor_states.timestamp << std::endl;
for (auto& state: rotor_states.rotors) {
    std::cout << state.thrust << " " << state.torque_scaler << " " << state.speed << std::endl;
}
```

I'm however getting exceptions with the sensor APIs if the sensors aren't specified, maybe a behavioural change from one of the recent PRs. And the sensor APIs are working fine for you but not the rotor state API?

KIYONG-KWON:
Based on your link #3726, i updated AIrSimSettings.hpp and rebuild
However, below error is not still solved. 
I think my problem is different with #3726 because other sensor signals come in well.
[...]

PR #3726, opened 2021-05-23
Use default sensors for default vehicle

rajat2004:
Fixes: #    

## About

This fixes a behavioural change I came across while testing #3719 

Use the settings -
```json
{
    "SettingsVersion": 1.2,
    "SimMode": "Multirotor"
}
```

and try using one of the sensor APIs such as `getImuData()`, etc. It throws an rpc error. Here's the output from `HelloDrone` -

```
Exception raised by the API, something went wrong.
rpclib: function 'getBarometerData' (called with 2 arg(s)) threw an exception. The exception is not derived from std::exception. No further information available.
```

This should work however, and it does work in 1.4.0, I think the change is due to [this](https://github.com/microsoft/AirSim/commit/006d4e2ac54f31e33ed595c6cdfbe2d5483027fd#diff-8ac7352e75eab96519f4de88358272d3b4de89e4b343ff35f228c5d7c3fc79ee). This error won't occur if a vehicle is present, even without any settings

It also adds the default sensors for vehicles created using the `simAddVehicle` API. Earlier, using any sensor APIs for these vehicle results in an error, and succeeds after this PR.

I tried using `make_unique` but that is giving some strange linker errors with the ROS wrapper somehow - `/home/testuser/AirSim/ros/devel/.private/airsim_ros_pkgs/lib/libairsim_settings_parser.so: undefined reference to `msr::airlib::AirSimSettings::kVehicleTypePhysXCar'` so have removed that for now

## How Has This Been Tested?

Above settings, and HelloDrone,

 - Works with 1.4.0
 - Doesn't work with 1.5.0, master
 - Works after this PR

`simAddVehicle` - Added `client.getImuData(vehicle_name=vehicle_name)` in `add_drone.py`, doesn't work earlier, works with this PR 

### DIFF
--- a/AirLib/include/common/AirSimSettings.hpp
+++ b/AirLib/include/common/AirSimSettings.hpp
@@ -257,6 +257,15 @@ namespace airlib
             std::map<std::string, std::shared_ptr<SensorSetting>> sensors;
 
             RCSettings rc;
+
+            VehicleSetting()
+            {
+            }
+
+            VehicleSetting(const std::string& vehicle_name_val, const std::string& vehicle_type_val)
+                : vehicle_name(vehicle_name_val), vehicle_type(vehicle_type_val)
+            {
+            }
         };
 
         struct MavLinkConnectionInfo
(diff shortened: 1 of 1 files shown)


jonyMarino:
Thanks, Rajat!

PR #3729, opened 2021-05-25
move AirSimAssets.umap to inside plugin

zimmy87:
Fixes: n/a    

## About
AirSimAssets.umap is now included the plugin's content folder so that it does not have to be manually copied into a custom environment. Documentation has also been updated to include a step on adding configuration for the umap's new location.

## How Has This Been Tested?
Both car and multirotor mode were tested within the editor and in a packaged version of the Blocks project.


jonyMarino:
Thanks David!

PR #3746, opened 2021-05-27
Added movetoGPS

ahmed-elsaharti:
## About
Adds a `movetoGPS` API to let users specify a GPS position to go to rather than being limited to NED space with movement APIs when using simpleflight.

Almost an exact copy of the `movetoposition` call but uses the home-geopoint to calculate the NED coordinates of the GPS position provided and passes that to the move method (in NED).

Still a WIP but would love to hear any feedback

## How Has This Been Tested?
Tested locally on Windows


MinnDevelopment:
Would it make sense to also support a function similar to moveOnPath?

zimmy87:
Thanks for the contribution @ahmed-elsaharti! Tested this locally and it works as expected for me, so I am moving ahead with merging. As for @MinnDevelopment's comment, a GPS-enabled version of moveOnPath would be great, but I think it's a little out of scope for this PR (it'd be a great idea for a separate PR).

PR #3755, opened 2021-06-02
PythonClient: Add __version__ attribute and use it to indicate version

### LINKED ISSUE #3634: Version attribute for python airsim package
### Overview:

the idea is to be able to run something like this:
```
import airsim
print (airsim .__ version__)
```
and get a string like `1.5.0`

### Nature of Request:  

- Addition

This will allow you to know which version of AirSim is being used.

rajat2004:
Fixes: #3634    

## About

Adds a `__version__` attribute to the airsim package, and uses it in `setup.py` and Sphinx `conf.py`.

This was done since the [API Reference Docs](https://microsoft.github.io/AirSim/api_docs/html/) title shows the version as 1.2.8. It could be updated in just `conf.py`, but that 2 places to be modified every time. Therefore added the `__version__` attribute which only needs to be modified once.

Idea from https://stackoverflow.com/questions/26141851/let-sphinx-use-version-from-setup-py

## How Has This Been Tested?

Local airsim install -

```
>>> import airsim
>>> airsim.__version__
'1.5.0'
```

Docs with `mkdocs serve` shows the 1.5.0 version in the title as well.  I haven't seen any issues arising from this till now

### DIFF
--- a/PythonClient/airsim/__init__.py
+++ b/PythonClient/airsim/__init__.py
@@ -2,3 +2,4 @@ from .client import *
 from .utils import *
 from .types import *
 
+__version__ = "1.5.0"
--- a/PythonClient/docs/conf.py
+++ b/PythonClient/docs/conf.py
@@ -17,6 +17,7 @@ import sys
 sys.path.insert(0, os.path.abspath('..'))
 
 import sphinx_rtd_theme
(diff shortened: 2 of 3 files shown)


jonyMarino:
Thanks, Rajat!

PR #3784, opened 2021-06-11
add wrapper for line of sight APIs

zimmy87:
Fixes: n/a

## About
This PR adds wrappers for simTestLineOfSightToPoint, simTestLineOfSightBetweenPoints, and simGetWorldExtents. It also adds an example script for calling these APIs.

## How Has This Been Tested?
Example script was called and output was sanity checked.

### DIFF
--- a/AirLib/include/api/RpcLibClientBase.hpp
+++ b/AirLib/include/api/RpcLibClientBase.hpp
@@ -111,8 +111,8 @@ namespace airlib
         vector<ImageCaptureBase::ImageResponse> simGetImages(vector<ImageCaptureBase::ImageRequest> request, const std::string& vehicle_name = "");
         vector<uint8_t> simGetImage(const std::string& camera_name, ImageCaptureBase::ImageType type, const std::string& vehicle_name = "");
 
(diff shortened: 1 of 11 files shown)


rajat2004:
Isn't it cleaner to use `GeoPoint` directly instead of separate lat, lon & alt?

zimmy87:
good point @rajat2004, I've updated the python wrapper as well as the original APIs to use GeoPoint parameters instead of separate lat, lon & alt parameters.

rajat2004:
Also, this is somewhat separate from this PR but the `simGetWorldExtents` & `simTestLineOfSightBetweenPoints` APIs are better suited in WorldSimApi.cpp rather than PawnSimApi IMHO since they both don't deal with vehicles at all. 
Their actual implementation can use the wrapper `RunCommandOnGameThread` instead of the FTaskGraph stuff, there's a bit of code duplication between `simTestLineOfSightBetweenPoints` & `simTestLineOfSightToPoint` also, the latter could call the former but that'll be likely bit painful to implement. Just some things which jumped out to me when looking at the code

zimmy87:
Thanks @rajat2004 for the suggestions. I think we are going to implement those changes in a separate PR.

PR #3788, opened 2021-06-13
Add detection support for all image types

alonfaraj:
- Extend detection support for all image types
- `UDetectionComponent` now generated via c++ to keep it simpler and and without changing BP_PIPCamera asset

Fixes: #    

https://github.com/microsoft/AirSim/pull/3472#issuecomment-860091855

## About
Last PR https://github.com/microsoft/AirSim/pull/3472# included `image_type` in API but supported only `Scene`.
This PR fix it.

## How Has This Been Tested?
Using detection.py script on different image types with Block environment.


Unreal/Plugins/AirSim/Source/PIPCamera.cpp
```diff
@@ -234,6 +237,10 @@ void APIPCamera::EndPlay(const EEndPlayReason::Type EndPlayReason)
         captures_[image_type] = nullptr;
         render_targets_[image_type] = nullptr;
     }
+
+    for (unsigned int image_type = 0; image_type < imageTypeCount(); ++image_type) {
+        detections_[image_type] = nullptr;
+    }
```
rajat2004, line 243: Can be moved into the above loop itself

rajat2004:
Slightly separate from this, there are 3 methods declared in AirBlueprintlib.h without any definition and usage, can be removed

https://github.com/microsoft/AirSim/blob/4206ac57affa77f2071629018899945c88c4f3ff/Unreal/Plugins/AirSim/Source/AirBlueprintLib.h#L196-L198

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
```diff
@@ -71,10 +71,13 @@ void APIPCamera::PostInitializeComponents()
     captures_[Utils::toNumeric(ImageType::SurfaceNormals)] =
         UAirBlueprintLib::GetActorComponent<USceneCaptureComponent2D>(this, TEXT("NormalsCaptureComponent"));
 
-    detections_[Utils::toNumeric(ImageType::Scene)] =
-        UAirBlueprintLib::GetActorComponent<UDetectionComponent>(this, TEXT("DetectionComponent"));
-    if (detections_[Utils::toNumeric(ImageType::Scene)]) {
-        detections_[Utils::toNumeric(ImageType::Scene)]->Deactivate();
+    for (unsigned int i = 0; i < imageTypeCount(); ++i) {
+        detections_[i] = NewObject<UDetectionComponent>(this);
+        if (detections_[i]) {
```
rajat2004, line 76: This is just something which I got to know about right now, in UE4, it's preferable to use `IsValid(ptr)` instead of nullptr check since it checks whether the object has been properly initialized or not, and whether it's slated to be garbage collected, started from [this article](https://mikelis.net/garbage-collection-in-ue4-a-high-level-overview/) and then some other google searches which indicate the same that it's better to use IsValid. It _might_ be something which needs to be handled in other places also, and might help in some strange crashes seen long time ago

Just came to my mind, up to you and the maintainers if this is something which needs to be handled

alonfaraj: You right.
Personally I use `IsValid(ptr)` in my own projects, but I have noticed AirSim using a simple nullptr check so didn't want to do something different than their convention.
I think the maintainers should make this change at once for the while project, if they think it's the right check to use.

zimmy87: Will discuss with the rest of my team about using `IsValid(ptr)` in rest of repo. For the purposes of this PR, the nullptr check is fine for the time being.

zimmy87:
Tested with detection.py script using a variety of different image types. Everything worked as intended so I am moving ahead with merging.

Issue #1154, opened 2018-06-11
Sensor Data Coordinate Convention


jameswalmsley:
I've read in the documentation that the NED coordinate convention is used throughout AirSim.

Can someone confirm that the sensor data coming into e.g. SimpleFlight is also in NED convention.

1. When looking down on top of the drone, a positive rotation (Clockwise) in Yaw (Z down axis) will cause a positive gyro signal?
2. A pitch upwards of the drone (causing the front of the drone to pitch up by a positive angle) will cause a positive gyro signal?
3. A roll to the right will cause a positive gyro signal?

If the drone is upright I can see an acceleration of -9.81 m/s^2
If the drone is rolled 90 degrees on its right, I should see -9.81 on the Y axis?
If the drone is pitched 90 degrees forward (on its nose) I should see -9.81 on the X axis?

Can anyone confirm that my assumptions are correct.

I would really love to be able to pick up the drone and spin it around in the simulator to test this.

Many thanks..

James

sytelus:
Yes, I think that's what should happen. Sensor data is also in NED. Try it out and if that's not the case we will look at it.

jameswalmsley:
@sytelus Many thanks for getting back to me.

After further investigating it seems to be in NED.
Can you confirm what gravity should look to a NED sensor. If the axis is pointing to the ground, should it see -9.81 ms/s^2 or +9.81?

Currently in AirSim its all -ve, so I assume this is correct. 

I've fed the imu data into the MadgwickAHRS filter, and it also has a NED coordinate system. However I have to negate all the accel data, hence the question.

Thanks

James

kylelindgren:
At complete resting state, the IMU should report -9.81 m/s^2 though the AirSim 1.2 API for simGetGroundTruthKinematics() does not account for gravity and it needs to be manually added after converting to body frame.

madratman:
Yes, you're right in that simGetGroundTruthKinematics() is in world NED frame and needs to be converted into body frame and gravity needs to be added. See discussion in https://github.com/Microsoft/AirSim/issues/1909. I think you basically did what the first comment about. 
https://github.com/Microsoft/AirSim/pull/1920 adds an IMU API, so I am closing this

yuyangch:
Hi, I am confused here,the body frame also uses NED convention right? where z points downward, so if the drone is resting upright, shouldn't the downward pointing  z direction acceleration reads +9.81 instead of -9.81?

[...]

PR #3801, opened 2021-06-16
change sign of gravity in IMU

### LINKED ISSUE #3074: IMU data generation error for the linear acceleration?
When collecting the IMU data, I found the changing range of linear accelerations of x and y direction are always very small (usually below 0.1) while z acceleration looks normal (bigger than 1). I did a little debugging and found that the gravity direction seems to be the opposite.
Specifically, in the ImuSimple.hpp, it is:
`output.linear_acceleration = ground_truth.kinematics->accelerations.linear - ground_truth.environment->getState().gravity;` , which gives a z_acc of -9.8. But from my understanding, the acc should be in NED, so the a_acc should be +9.8?

hturki:
Fixes: #3074    
https://github.com/microsoft/AirSim/issues/1154

## About
Since AirSim uses NED coordinates, we should be *adding* gravity to the z acceleration and not substracting it

## How Has This Been Tested?
After changing the sign of gravity, I attempted to use https://github.com/HKUST-Aerial-Robotics/VINS-Fusion on data I collected from AirSim and seem to get reasonable results

image

The dotted grey line is the ground truth and the solid blue line is the prediction

### DIFF
--- a/AirLib/include/sensors/imu/ImuSimple.hpp
+++ b/AirLib/include/sensors/imu/ImuSimple.hpp
@@ -55,7 +55,7 @@ namespace airlib
             const GroundTruth& ground_truth = getGroundTruth();
 
             output.angular_velocity = ground_truth.kinematics->twist.angular;
-            output.linear_acceleration = ground_truth.kinematics->accelerations.linear - ground_truth.environment->getState().gravity;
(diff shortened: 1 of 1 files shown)


jonyMarino:
Thanks, @hturki! and @Amigoshan for catching this!

Issue #3415, opened 2021-02-22
Bug: PX4 SITL EKF-Failure/GPS-loss


ahmed-elsaharti:
(Similar to the post over at #3406 but a bug report is more appropriate at this point)
## Bug report 

- AirSim Version/#commit: 1.4 0096b5e
- UE/Unity version: 4.25
- autopilot version: PX4 1.10.1 SITL
- OS Version: Windows 10 x64 (Running SITL on Cygwin 0.8)

### What's the issue you encountered?
As seen in #3167, #3213, #3087 and many more, after launching AirSim in PX4 SITL mode (with lockstep mode enabled by default) the terminal displays the following errors continuously:
![image](https://user-images.githubusercontent.com/49626509/108399886-3b833680-71e0-11eb-8225-ff627f6535fe.png)
Although they do not affect the flight instantaneously, an EKF error shows up, QGC throws an 'no global position' error and the drone flies diagonally into the ground (and a line is actually traced on QGC showing that the GPS estimate just threw itself way off). As this happens `timeout on HilActuatorControlMessage, resetting lock step mode` appears on UE's side.
Here's a gif showing the problem in action:
![PX4Bug](https://user-images.githubusercontent.com/49626509/108400027-64a3c700-71e0-11eb-9ea6-9cf5a5e716fe.gif)

A workaround that fixes this by disabling lockstep mode is discussed in #3406. **If you are facing a similar issue and need a temp fix have a look at that post**

Also, I'm curious as to whether this problem occurs on Linux systems too. If not then this might be caused by Cygwin and AirSim not communicating well?
Anyway, it would be great if someone could test this out on a Linux system.

### Settings
```json
{
  "SettingsVersion": 1.2,
  "SeeDocsAt": "https://github.com/Microsoft/AirSim/blob/master/docs/settings.md",
  "SimMode": "Multirotor",
  "DefaultVehicleConfig": "PX4",
  "Vehicles": {
    "PX4": {
            "VehicleType": "PX4Multirotor",
            "UseSerial": false,
            "UseTcp": true,
            "TcpPort": 4560,
            "ControlPort": 14570
        }
  },
  "OriginGeopoint": {
    "Latitude": 47.641468,
    "Longitude": -122.140165,
     "Altitude": 0
    }
}

```

### How can the issue be reproduced?
**Note** this error is inconsistent, ie: sometimes it happens as soon as the drone takes off and sometimes it takes a good 2-3 mins of flying before it happens.

To reproduce:

[...]

PR #3828, opened 2021-06-25
Update px4_lockstep.md

zimmy87:
Fixes: #    

## About

This adds a new section to this page about disabling lockstep, derived from steps described in #3406 and useful as a workaround for #3415.

## How Has This Been Tested?

Web editor preview

### DIFF
--- a/docs/px4_lockstep.md
+++ b/docs/px4_lockstep.md
@@ -39,7 +39,9 @@ in the simulated flight if you look at the updates on screen in realtime.
 
 # Disabling LockStep
 
-PX4 is configured to use lockstep by default. To disable this feature, first [disable it in 
+If you are running PX4 in cygwin, there is an [open issue with 
+lockstep](https://github.com/microsoft/AirSim/issues/3415). PX4 is configured to use lockstep by 
+default. To disable this feature, first [disable it in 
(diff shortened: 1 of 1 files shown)


docs/px4_lockstep.md
```diff
@@ -37,4 +37,30 @@ There will be some side effects to `lockstep`, namely, slower update loops cause
...
+    ```
+        {
+            ...
+            "Vehicles": {
+                "PX4": {
+                    "VehicleType": "PX4Multirotor",
+                    "UseTcp": true,
+                    "LockStep": false,
+                    ...
+    ```
```
lovettchris, line 64: and remove any SteppableClock if you configured that.

lovettchris: We also need to check if you remove lockstep does PX4 revert to using UDP for the HIL mavlink channel ?

zimmy87: Added extra text for removing SteppableClock setting. Also removed the reference to the "UseTcp" setting since it's unrelated to lockstep specifically. I did some testing and PX4 still uses TCP with and without lockstep so there's no necessity in changing that setting.

lovettchris (approved):

jonyMarino:
Thanks @zimmy87 and @lovettchris!

Issue #3829, opened 2021-06-25
add msr::airlib::EarthUtils::GeodeticToNed method


zimmy87:
### Overview:

Right now there is a msr::airlib::EarthUtils::GeodeticToNedFast for converting GeoPoints to NED coordinates, but the algorithm implemented is a approximate that can result in errors of up to 10km. There is a corresponding method for converting the opposite way, msr::airlib::EarthUtils::nedToGeodeticFast, that has a non-fast version, but we don't yet have a non-fast version of msr::airlib::EarthUtils::GeodeticToNedFast.

### Nature of Request:  

- Addition
 

## Why would this feature be useful?
This would remove any possible error in the WorldSimApi::testLineOfSightBetweenPoints API

rajat2004:
There's a file [GeodeticConverter.hpp](https://github.com/microsoft/AirSim/blob/master/AirLib/include/common/GeodeticConverter.hpp), it was discussed a bit in https://github.com/microsoft/AirSim/pull/3364#issuecomment-826106220 and later as well. I'll try to bring in the changes I made and see if this turns out to be useful, currently it's unused

zimmy87:
Ah I forgot about that file, I've updated calls to msr::airlib::EarthUtils::GeodeticToNedFast with calls to msr::airlib::GeodeticConverter::geodetic2Ned in #3810

PR #3810, opened 2021-06-22
move line-of-sight APIs to WorldSimApi

zimmy87:
Fixes: #    

## About

moving testLineOfSightBetweenPoints and getWorldExtents to WorldSimApi which is a more appropriate location for these APIs

## How Has This Been Tested?

built and launched with SimpleFlight for sanity check

## Screenshots (if appropriate):

PR #3832, opened 2021-06-29
Add small helper methods for GeodeticConverter

rajat2004:
Fixes: #    

## About

Related to #3829, might be useful for #3810
Ping @zimmy87

## How Has This Been Tested?

Hasn't been explicitly tested for this PR apart from compilation, extracted out of some earlier work. In that, it was tested by replacing the currently being used EarthUtils with GeodeticConverter which might be interesting as well, will try to open a PR for that as well later

I haven't currently added a `geodetic2Ned` method since it will cause a narrowing of NED data from `double` to `float`. If that is allowable then I'll add that method as well

### DIFF
--- a/AirLib/include/common/GeodeticConverter.hpp
+++ b/AirLib/include/common/GeodeticConverter.hpp
@@ -20,6 +20,11 @@ namespace airlib
             setHome(home_latitude, home_longitude, home_altitude);
         }
 
+        GeodeticConverter(const GeoPoint& home_geopoint)
+        {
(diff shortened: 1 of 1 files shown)


zimmy87:
Thanks for the contribution @rajat2004! I tested the new constructor and the `GeodeticConverter.ned2Geodetic()` method with a modified version of the `PawnSimApi::testLineOfSightToPoint()` method and the line_of_sight.py script. Everything worked as expected so I am moving ahead with merging.

Issue #1855, opened 2019-03-25
NED location and GPS


KeyneZzz:

DavidLSmyth:
If the distance is far enough the curvature of the earth may be responsible (equations that calculate the distance between gps coordinates takes the curvature of the earth into account). Do you have the figures outlining the inconsistency? Apologies if you've already ruled this out

KeyneZzz:
> If the distance is far enough the curvature of the earth may be responsible (equations that calculate the distance between gps coordinates takes the curvature of the earth into account). Do you have the figures outlining the inconsistency? Apologies if you've already ruled this out

@DavidLSmyth thanks for your reply, and I am not considering the curvature thing.
My route is firstly heading north for 1.5 kilometer, then then turn to east for 1.3 km. When arrived, it's about 20 meters northeast from the set endpoint. The latitude is about 30 degrees.

DavidLSmyth:
I did some of the calculations and as far as I could see there should be a small error in the gps location due to the non-planar geometry of the earth, the points are close enough to be considered on a plane. I'm not sure how the NED figures are calculated, it might be worth having a look at the source code. The calculation of GPS location is easy to find in the source code but it could be worth checking that too

madratman:
So, I did some digging:

the NED<->geodetic conversions are handled by `EarthUtils::nedToGeodetic` which you can inspect in https://github.com/Microsoft/AirSim/blob/123bbd6d6416bbe98e775101ba8ebcb5ccbf8501/AirLib/include/common/EarthUtils.hpp#L288-L304

Now, [PawnSimApi.cpp](https://github.com/Microsoft/AirSim/blob/39fb3b0b051050f4bc9d770838808756b64ff5ab/Unreal/Plugins/AirSim/Source/PawnSimApi.cpp) uses the above `EarthUtils::nedToGeodetic` in two places:
 -  to set the correspondence b/w home geo point (from settings.json) and unreal pawn's spawn coordinates and the NED frame. This can be seen in https://github.com/Microsoft/AirSim/blob/39fb3b0b051050f4bc9d770838808756b64ff5ab/Unreal/Plugins/AirSim/Source/PawnSimApi.cpp#L67
  
[...]

PR #3839, opened 2021-07-01
Use Geodetic Converter for NED<->GPS conversions

rajat2004:
Fixes: #    

## About

This PR updates the environment code to use GeodeticConverter instead of EarthUtils which is potentially more inaccurate, described a bit in https://github.com/microsoft/AirSim/issues/1855#issuecomment-478166060, https://github.com/microsoft/AirSim/pull/3364#issuecomment-827241798

The PR also includes the fixes described in https://github.com/ethz-asl/geodetic_utils/issues/36, https://github.com/ethz-asl/geodetic_utils/issues/28

## How Has This Been Tested?

A fair bit of testing needs to be done, to verify that GPS inaccuracy doesn't increase. Unit tests for GeodeticConverter itself which be a good option, something to keep in mind. This was earlier tested with ArduPilot without any problems, but also without any noticeable improvement in the GPS performance. Testing with PX4 will also be good

### DIFF
--- a/AirLib/include/common/GeodeticConverter.hpp
+++ b/AirLib/include/common/GeodeticConverter.hpp
@@ -20,6 +20,11 @@ namespace airlib
             setHome(home_latitude, home_longitude, home_altitude);
         }
 
+        GeodeticConverter(const GeoPoint& home_geopoint)
+        {
+            setHome(home_geopoint);
+        }
+
         void setHome(double home_latitude, double home_longitude, float home_altitude)
         {
             home_latitude_ = home_latitude;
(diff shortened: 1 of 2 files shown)


jonyMarino:
Hi @rajat2004! What is the state of this PR?

rajat2004:
I haven't been able to test this out, and verify if the GPS accuracy is better or worse. It can be closed if preferred

zimmy87:
We did some sanity-checking on the new output of the GPS sensor, and it looks reasonable to us, plus we didn't observe any performance regression with this PR applied, so we decided we'd like to merge this. thank you for the submission, @rajat2004!

PR #3320, opened 2021-01-19
Add support for fixed external cameras

rajat2004:
Fixes: #3208 
Fixes: #1896, 
Fixes: #1709, 
Fixes: #610, 
Fixes: #599
 

## About

This PR adds the capability to create a fixed camera separate from the cameras on the vehicle, such as a CCTV camera. I'm not aware of any simple way to go about doing this currently, and therefore thought to try it out. I haven't checked the issues very thoroughly, so please let me know if there are other related issues and ways to go about doing this.

PR has been opened to get ideas and suggestions and whether this is actually required. This will likely become a big PR, so it can be split into parts for easier review also.

Currently have added a boolean field `external` in the `getCameraInfo` API, went for this since there are currently 7 camera related APIs, which could increase later on as well. Having completely separate APIs will duplicate the code in many places, the new API while doesn't look the cleanest felt like the better choice to me. Do comment if splitting into a different API or some other way is more preferred.

TODO (Comment if something's missing) -
- Camera APIs implementation
  - [x] `getImages`
  - [x] `set/getDistortionParams`
  - [x] `getCameraInfo`
  - [x] `setCameraFov`
  - [x] `setCameraPose`
  - [x] Detection APIs
- Allow External Cameras in -
  - [x] Subwindows
  - [ ] Recording (This will likely require a bit of tinkering, after #2861, possibly even in a separate PR)
- [x] Documentation updates
- [x] Docstrings update
- [x] Unity fixes to preserve earlier behaviour (No external camera implementation currently)

Also removes the `simSetCameraOrientation` API kept around for backward compatibility

The last commit to add an internal `CameraDetails` struct is a NFC commit, just to clean up the multiple parameters being passed everywhere. If it's not desirable then the coomit can be dropped easily. Any suggestions on the struct naming are welcome!

## How Has This Been Tested?

<details>
<summary>Example Settings</summary>

```json
{
    "SettingsVersion": 1.2,
    "SimMode": "Car",
    "ExternalCameras": {
        "fixed1": {
            "X": 0, "Y": 0, "Z": -10,
            "Pitch": -90, "Roll": 0, "Yaw": 0
        },

        "fixed2": {
            "CaptureSettings": [
                {
                  "ImageType": 0,
                  "Width": 256,
                  "Height": 144,
                  "FOV_Degrees": 90,
                  "AutoExposureSpeed": 100,
                  "AutoExposureBias": 0,
                  "AutoExposureMaxBrightness": 0.64,
[...]

PR #3865, opened 2021-07-08
Convert line breaks

zimmy87:
Fixes: n/a    

## About

update CRLF to LF line breaks


rajat2004:
TBH I don't really see the point of this mass formatting. Agree that having a common line break will be good, but this will again lead to mass conflicts and problems with the git history. The clang formatting made the code more readable and enforced common styles, but this doesn't really add any such benefits. Again, up to you and the maintainers though

jonyMarino:
> TBH I don't really see the point of this mass formatting. Agree that having a common line break will be good, but this will again lead to mass conflicts and problems with the git history. The clang formatting made the code more readable and enforced common styles, but this doesn't really add any such benefits. Again, up to you and the maintainers though
> 
> Also, if this does go through, it'll be best to ensure this is enforced, probably through clang-format itself

@rajat2004, Do you know if git doesn't resolve conflicts for line break format automatically?

rajat2004:
Well, it's kind of a mixed response from my testing out merging this branch with others. #3720 didn't get any conflicts, but #3320 had conflicts in one of the later commits, no idea why. So might not be a big problem with the conflicts.

Is there any way to enforce that only LF line endings are committed after this? Maybe through .gitattributes file - https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings ?

jonyMarino:
Good idea! I think this configuration will accomplish that: 

> text eol=lf Git will always convert line endings to LF on checkout. You should use this for files that must keep LF endings, even on Windows.

We can wait to merge this until we get no conflicts from the current open PRs

jonyMarino:
Let's close this for safety

PR #3922, opened 2021-08-01
Change getRotorStates to return const ref

rajat2004:
Fixes: #    

## About

NFC, just don't see any reason why the internal method should return a copy. If someone wants to use, they can still create a copy and modify

## How Has This Been Tested?

Compilation


jonyMarino:
Thanks, Rajat!

PR #3936, opened 2021-08-07
Add GazeboDrone compilation to Actions

rajat2004:
Fixes: #    

## About

Ensures that GazeboDrone is supported and at least compilation is tested
The compiler used for GazeboDrone was also changed to GCC 8 since AirLib was compiled using this, and different compiler version for this can cause problems

## How Has This Been Tested?

Through Actions CI only. Haven't yet changed the documentation since would like someone to test this out locally also


jonyMarino:
Thanks, Rajat!

PR #3937, opened 2021-08-07
Fix compression bug in segmentation palette

### LINKED ISSUE #3423: Image Segmentation Color mapping wrong
## Bug report 

- AirSim Version/#commit: 288c90ca78ce09ed58de8508622cb97071c8f217
- UE/Unity version: 2.45
- autopilot version: N/A
- OS Version: OSX Catalina

The Image segmentation colors changed when upgrading to latest version of Airsim. Unfortunately I am not sure what commit I was using before but the priori version I had pulled master around September. I still have a world with the old Airsim plugin so if there is a way to extract the commit for Airsim from that I can provide the old version of Airsim that I was using.

Environment: Mountain Landscape

With the September Airsim Code:
image_0_segmented_original

With the Latest Airsim Code:
image_0_segmented_new
 
As suggested in this page: https://microsoft.github.io/AirSim/image_apis/#segmentation I set the object ID for the bright pink lake in the original segmentation to object id 37.
According to the RGB values to Object id file (https://microsoft.github.io/AirSim/seg_rgbs.txt) this maps to RGB 254,12,229

I am use that RGB value to extract out the Lake and everything works fine with the old version of Airsim.

Now when I upgraded Airsim, using the same exact code, i.e it sets the same static mesh to object ID 37, it now fails to extract the lake because the RGB values have changed. The RGB value for the lake is now 253,0,200.

Therefore, it seems that the color mapping file is no longer accurate and the colors changed. Was this intentional? I do not see any updates on the segmentation color mapping file. 

### Settings
The settings are the same for both configurations

{
  "SettingsVersion": 1.2,
  "SimMode": "Car",
  "CameraDefaults": {
    "CaptureSettings": [
      {
        "ImageType": 0,
        "Width": 1080,
        "Height": 1080,
        "FOV_Degrees": 90
      },
      {
        "ImageType": 5,
        "Width": 1080,
        "Height": 1080,
        "FOV_Degrees": 90
      }
    ]
  }
}

To reproduce:
1. First call client.simSetSegmentationObjectID(37, "SM_IcePlane_9") so the main lake is the proper object ID 
2. Using Unreal 4.25 with the latest Airsim commit, use client.simGetImages with image type airsim.ImageType.Segmentation to retrieve the image of the Mountain Landscape environment
3. Now load the image in opencv or Pillow load the image to numpy array
4. Inspect the image and go to the location in the image that contains the lake and verify the RGB values do not line up with the color pallet on the airsim website.

saihv:
Fixes: https://github.com/microsoft/AirSim/issues/3423

## About
The modified segmentation material in https://github.com/microsoft/AirSim/pull/3053 used a wrong compression type on the palette texture, causing the colors to become inconsistent with the old seg_rgbs.txt. Compression should be HDR instead of Default, and the sampler type in SegmentationMaterial.uasset for the texture should be LinearColor instead of Color.

## How Has This Been Tested?
With the script from https://github.com/microsoft/AirSim/issues/3423#issuecomment-788077713 (thanks @TidalPaladin !) - the output of this script should match the (flipped) values from https://microsoft.github.io/AirSim/seg_rgbs.txt.

Before fix: 

image

After fix:

image

Output of script vs. actual seg_rgb values:

image


jonyMarino:
Thanks, Sai!

PR #3939, opened 2021-08-08
Fix crash in ros wrapper on JetsonTX2

alonfaraj:
Fixes: #    

## About

ROS wrapper keep crashing on Jetson TX2, after build it in debug mode, turns out `has_gimbal_cmd_` is not initialized which caused a crash when calling `simSetCameraPose`.
It's doesn't reproduce on my VMware under widows host, probably due to difference architectures.

## How Has This Been Tested?

Build and run the ros wrapper on Jetson TX2


rajat2004:
It might be worth looking into enabling more compiler warnings for the ROS wrapper, have started #3945, however it does not catch this variable unfortunately

PR #3945, opened 2021-08-10
Enable some warnings

rajat2004:
Fixes: #    

## About

Enables a few warnings for AirSim and the ROS wrapper, needs testing

## How Has This Been Tested?

Compilation and running Multirotor in UE4
ROS testing apart from compilation is left


AirLib/include/common/ImageCaptureBase.hpp
```diff
@@ -32,21 +32,20 @@ namespace airlib
         struct ImageRequest
         {
             std::string camera_name;
-            ImageCaptureBase::ImageType image_type = ImageCaptureBase::ImageType::Scene;
+            ImageCaptureBase::ImageType image_type;
             bool pixels_as_float = false;
             bool compress = true;
 
-            ImageRequest()
```
rajat2004, line 39: This is a breaking change, will require changes in any existing C++ client code if they're using it. Can be added back easily if not preferred

rajat2004: Change reverted

cmake/cmake-modules/CommonSetup.cmake
```diff
@@ -44,13 +44,15 @@ macro(CommonSetup)
...
+        set(CMAKE_CXX_STANDARD 17)
+
         if (APPLE)
-            set(CMAKE_CXX_STANDARD 17)
             set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wstrict-aliasing -D__CLANG__")
         else ()
             set(CMAKE_CXX_FLAGS "\
-                -std=c++17 -ggdb -Wall -Wextra \
-                -Wno-variadic-macros -Wno-parentheses -Wno-unused-function -Wno-unused \
+                -Wall -Wextra \
```
zimmy87, line 53: Should I be seeing extra warnings in the output of build.cmd due to these added flags? I'm getting the same number of warnings both with and without this PR applied, is that to be expected?

DroneShell/include/SimpleShell.hpp
```diff
@@ -142,7 +144,7 @@ namespace airlib
             virtual bool execute(const ShellCommandParameters&) { return false; };
 
         private:
-            ShellCommand() {}
+            // ShellCommand() {}
```
zimmy87, line 147: Is there any reason to keep this code commented out?

ros/src/airsim_ros_pkgs/src/pd_position_controller_simple.cpp
```diff
@@ -177,21 +177,20 @@ bool PIDPositionController::gps_goal_srv_cb(airsim_ros_pkgs::SetGPSPosition::Req
...
         msr::airlib::GeoPoint gps_home(gps_home_msg_.latitude, gps_home_msg_.longitude, gps_home_msg_.altitude);
-        bool use_eth_lib = true;
-        if (use_eth_lib_for_geodetic_conv_) {
-            double initial_latitude, initial_longitude, initial_altitude;
-            geodetic_converter_.getReference(&initial_latitude, &initial_longitude, &initial_altitude);
+        if constexpr (use_eth_lib_for_geodetic_conv_) {
+            // double initial_latitude, initial_longitude;
+            // float initial_altitude;
+            // geodetic_converter_.getHome(&initial_latitude, &initial_longitude, &initial_altitude);
+            // ROS_INFO_STREAM("[PIDPositionController] geodetic_converter_ GPS reference initialized correctly (lat long in radians) " << initial_latitude << ", "<< initial_longitude << ", " << initial_altitude);
```
zimmy87, line 184: Any reason for this commented out code? Also same for ln 230-233

zimmy87 (changes requested):
Tested building this locally and everything appears to be working fine, so this looks like a safe change to make, but I have a question about what the expected output of build.cmd should be as well as a few style comments.

cmake/cmake-modules/CommonSetup.cmake
rajat2004: The CMake setup is used only in Linux, not in Windows. I haven't currently looked into enabling more warnings for Windows with Visual Studio

DroneShell/include/SimpleShell.hpp
rajat2004: Currently have reverted the change since I'm not very clear why there's a private constructor

ros/src/airsim_ros_pkgs/src/pd_position_controller_simple.cpp
rajat2004: Removed, might have been useful for debugging purposes

cmake/cmake-modules/CommonSetup.cmake
zimmy87: my mistake, I got confused about which build scripts were consuming the CMake configuration files; resolving this comment

zimmy87 (approved):

zimmy87:
Tested build.sh and the ros wrapper are still working on Ubuntu 20.04; going to move ahead with merging this.

PR #3779, opened 2021-06-09
Camera replaced by Cinematographic Camera (CinemAirSim)

ppueyor:
Fixes: #    
-Integrates the CineCameraComponent into AirSim

## About
An AddOn for AirSim that includes all the tools to integrate a cinematographic camera and the needed tools to control/access to it.
More info available in repo: https://github.com/ppueyor/CinemAirSim
Or conference paper: https://arxiv.org/abs/2003.07664

## How Has This Been Tested?
Ubuntu 20 and UE4 4.26
Different demos and API calls
Results available in repo

## Screenshots:
https://www.youtube.com/watch?v=OJGNaitWZVA

PR #3949, opened 2021-08-10
Camera replaced by Cinematographic Camera (CinemAirSim)

### LINKED ISSUE #1489: Cine camera
Wondering if we can use CineCamera instead of the current CameraComponent that's used, since it provides control over Focal Length and other parameters? Are there any challenges to it? Or reasons why it's not an option to switch in Airsim? 
Thanks. 

@sytelus 
@lovettchris

### LINKED ISSUE #2311: CinematicCamera on Drones
Hi!

I would like to know if there is a way to change the manual focus distance of the cameras of the drones, either through settings.md or by calling a function, like it is done in UE4 with a Cinematics Camera:

image

Is it possible to place a Cinematics Camera on the drones?

ppueyor:
Fixes: #1489   
Fixes: #2303
Fixes: #2311
-Integrates the CineCameraComponent into AirSim
-New PR from https://github.com/microsoft/AirSim/pull/3779

## About
An AddOn for AirSim that includes all the tools to integrate a cinematographic camera and the needed tools to control/access to it.
More info available in repo: https://github.com/ppueyor/CinemAirSim
Or conference paper: https://arxiv.org/abs/2003.07664

## How Has This Been Tested?
Ubuntu 20 and UE4 4.26
Different demos and API calls
Results available in repo

## Screenshots:
https://www.youtube.com/watch?v=OJGNaitWZVA


Unreal/Plugins/AirSim/Source/PIPCamera.cpp
```diff
@@ -588,3 +595,188 @@ void APIPCamera::onViewModeChanged(bool nodisplay)
...
+}
+
+std::string APIPCamera::getCurrentFieldOfView()
+{
+    std::ostringstream ss;
+    ss << "Current Field Of View:\n\tHorizontal Field Of View: " << camera_->GetHorizontalFieldOfView() << ";\n\t Vertical Field Of View: " << camera_->GetVerticalFieldOfView();
+    return ss.str();
+}
+
+void APIPCamera::CopyCameraSettingsToAllSceneCapture(UCameraComponent* camera_)
```
zimmy87, line 750: According to the [naming conventions](https://microsoft.github.io/AirSim/coding_guidelines/#naming-conventions), the underscore suffix is used for member variables, this should just be `camera`: this is also causing a build error on my side:

C:/__w/13/s/Unreal/Environments/Blocks/Plugins/AirSim/Source/PIPCamera.cpp(750): error C4458: declaration of 'camera_' hides class member
  C:\__w\13\s\Unreal\Environments\Blocks\Plugins\AirSim\Source\PIPCamera.h(100): note: see declaration of 'APIPCamera::camera_'

AirLib/include/api/RpcLibClientBase.hpp
```diff
@@ -111,6 +111,24 @@ namespace airlib
         vector<ImageCaptureBase::ImageResponse> simGetImages(vector<ImageCaptureBase::ImageRequest> request, const std::string& vehicle_name = "", bool external = false);
         vector<uint8_t> simGetImage(const std::string& camera_name, ImageCaptureBase::ImageType type, const std::string& vehicle_name = "", bool external = false);
 
+        //CinemAirSim
+        std::vector<std::string> simGetPresetLensSettings(const std::string& camera_name, const std::string& vehicle_name = "", bool external = false);
+        std::string simGetLensSettings(const std::string& camera_name, const std::string& vehicle_name = "", bool external = false);
+        void simSetPresetLensSettings(const std::string& preset_lens_settings = "", const std::string& camera_name = "", const std::string& vehicle_name = "", bool external = false);
```
rajat2004, line 117: `preset_lens_settings` and `camera_name` defaults probably shouldn't be present

AirLib/include/api/RpcLibClientBase.hpp
```diff
@@ -111,6 +111,24 @@ namespace airlib
...
+        std::vector<std::string> simGetPresetFilmbackSettings(const std::string& camera_name, const std::string& vehicle_name = "", bool external = false);
+        void simSetPresetFilmbackSettings(const std::string& preset_filmback_settings = "", const std::string& camera_name = "", const std::string& vehicle_name = "", bool external = false);
```
rajat2004, line 119: Same, first 2 defaults shouldn't be there

AirLib/src/api/RpcLibClientBase.cpp
```diff
@@ -271,6 +271,106 @@ __pragma(warning(disable : 4239))
...
 
+        //CinemAirSim
+        std::vector<std::string> RpcLibClientBase::simGetPresetLensSettings(const std::string& camera_name, const std::string& vehicle_name, bool external)
+        {
+            vector<std::string> result = pimpl_->client.call("simGetPresetLensSettings", camera_name, vehicle_name, external).as<vector<std::string>>();
+
+            if (result.size() == 1) {
+                // rpclib has a bug with serializing empty vectors, so we return a 1 byte vector instead.
+                result.clear();
+            }
```
rajat2004, line 282: No longer needed, can be removed from other places also

AirLib/src/api/RpcLibServerBase.cpp
```diff
@@ -153,6 +153,86 @@ namespace airlib
...
+            if (result.size() == 0) {
+                // rpclib has a bug with serializing empty vectors, so we return a 1 byte vector instead.
+                result.push_back("");
+            }
+            return result;
+        });
+
+        pimpl_->server.bind("simGetLensSettings", [&](const std::string& camera_name, const std::string& vehicle_name, bool external) -> string {
+            auto result = getWorldSimApi()->getLensSettings(CameraDetails(camera_name, vehicle_name, external));
+            return result;
```
rajat2004, line 168: Just directly return instead of creating a var, can be cleaned up in other places after removing the vector checking also

PythonClient/airsim/client.py
```diff
@@ -265,7 +265,74 @@ def simGetImages(self, requests, vehicle_name = '', external = False):
         """
         responses_raw = self.client.call('simGetImages', requests, vehicle_name, external)
         return [ImageResponse.from_msgpack(response_raw) for response_raw in responses_raw]
-        
+
+#CinemAirSim
+    def simGetPresetLensSettings(self, camera_name, vehicle_name = '', external = False):  
+        result = self.client.call('simGetPresetLensSettings', camera_name, vehicle_name, external)
+        if (result == "" or result == "\0"):
+            return None
```
rajat2004, line 273: Again shouldn't be needed I think, but not 100% sure about this. There's no such handling inside current master I think

rajat2004 (commented):
The video and paper looks pretty awesome, and it's pretty nice that you want this to get merged in the project also
Just had a brief look, and had some concerns hope you might be able to answer, and might interest the maintainers also
- What's the performance diff if any between the current camera and the new one? Some FPS info would be good, #3889 might be useful. Also, does this support all the image types?
- Wondering whether replacing the current camera entirely might be a good idea or not. This adds a lot of camera APIs, and there's enough already. In case of problems with first point, whether having this as a separate camera might be a better option. A separate set of APIs (client) for camera might be something with looking into, it'll improve the current code also by avoiding mixing so many APIs

ppueyor:
Hi @rajat2004,

Thanks for your interest in this PR. My answers are below:

- I used that script and the average FPS is 19.90, don't know how much it is in the regular camera. It does support all kinds of image types, and it is possible to get images and subwindows the same way and updates as the regular camera.

- Well, I think that it's something that the maintainers should answer, if that could be interesting or not. I can talk about my own experience. It opened me to a whole world of possibilities. If you want to use the drone with filming objectives (that it is wide use of the UAVs), this PR is more than helpful because u could not modify the intrinsics of the camera otherwise.

About the corrections, @zimmy87 is having a look at the code, so if he considers that, I can correct the code with your suggestions.

rajat2004:
By the FPS, I meant some results to compare on your system itself, for the Cinematic and normal cameras, for different resolutions like 256x144, 720p, 1080p and maybe 4k also?
Makes sense, UAVs for filming is probably the most common usecase, other applications are also closely vision-related, and improvements in camera to model the real world would definitely be useful

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
```diff
@@ -588,3 +595,188 @@ void APIPCamera::onViewModeChanged(bool nodisplay)
...
+}
+
+std::string APIPCamera::getCurrentFieldOfView()
+{
+    std::ostringstream ss;
+    ss << "Current Field Of View:\n\tHorizontal Field Of View: " << camera_->GetHorizontalFieldOfView() << ";\n\t Vertical Field Of View: " << camera_->GetVerticalFieldOfView();
+    return ss.str();
+}
+
+void APIPCamera::CopyCameraSettingsToAllSceneCapture(UCameraComponent* camera)
```
zimmy87, line 750: function names should be camelCase; this applies to APIPCamera::CopyCameraSettingsToSceneCapture as well

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
zimmy87: this has been addressed in latest iteration

zimmy87 (changes requested):
Thanks for submitting this PR, this is a cool feature! Most of my feedback is style-related. I second @rajat2004's comment about replacing the default camera with the new UCineCameraComponent. I think the default user experience is different enough that there should be an option for switching to the new type of camera. It'd also be nice to see the rest of @rajat2004's review items addressed before merging this.

zimmy87:
Our Azure pipeline check is failing with the following build error:

```

In file included from /__w/13/s/Unreal/Environments/Blocks/Plugins/AirSim/Intermediate/Build/Linux/B4D820EA/UE4/Development/AirSim/Module.AirSim.2_of_2.cpp:2:
/__w/13/s/Unreal/Environments/Blocks/Plugins/AirSim/Source/TextureShuffleActor.cpp:24:49: error: incomplete type 'UMaterialInstanceDynamic' named in nested name specifier
        DynamicMaterialInstances[material_id] = UMaterialInstanceDynamic::Create(DynamicMaterial, this);
                                                ^~~~~~~~~~~~~~~~~~~~~~~~~~
/home/ue4/ue-4.25.1-linux-debugeditor/Engine/Source/Runtime/Engine/Classes/Engine/BlendableInterface.h:10:7: note: forward declaration of 'UMaterialInstanceDynamic'
class UMaterialInstanceDynamic;
      ^
In file included from /__w/13/s/Unreal/Environments/Blocks/Plugins/AirSim/Intermediate/Build/Linux/B4D820EA/UE4/Development/AirSim/Module.AirSim.2_of_2.cpp:2:
/__w/13/s/Unreal/Environments/Blocks/Plugins/AirSim/Source/TextureShuffleActor.cpp:27:33: error: member access into incomplete type 'UStaticMeshComponent'
        components[component_id]->SetMaterial(material_id, DynamicMaterialInstances[material_id]);
                                ^
/home/ue4/ue-4.25.1-linux-debugeditor/Engine/Source/Runtime/CoreUObject/Public/Templates/Casts.h:22:7: note: forward declaration of 'UStaticMeshComponent'
class UStaticMeshComponent;
      ^
In file included from /__w/13/s/Unreal/Environments/Blocks/Plugins/AirSim/Intermediate/Build/Linux/B4D820EA/UE4/Development/AirSim/Module.AirSim.2_of_2.cpp:2:
/__w/13/s/Unreal/Environments/Blocks/Plugins/AirSim/Source/TextureShuffleActor.cpp:30:42: error: member access into incomplete type 'UMaterialInstanceDynamic'
    DynamicMaterialInstances[material_id]->SetTextureParameterValue("TextureParameter", SwappableTextures[tex_id]);
                                         ^
/home/ue4/ue-4.25.1-linux-debugeditor/Engine/Source/Runtime/Engine/Classes/Engine/BlendableInterface.h:10:7: note: forward declaration of 'UMaterialInstanceDynamic'
class UMaterialInstanceDynamic;
      ^
4 errors generated.
```

Are you seeing this in local builds?

PythonClient/airsim/client.py
```diff
@@ -16,7 +16,7 @@ def __init__(self, ip = "", port = 41451, timeout_value = 3600):
             ip = "127.0.0.1"
         self.client = msgpackrpc.Client(msgpackrpc.Address(ip, port), timeout = timeout_value, pack_encoding = 'utf-8', unpack_encoding = 'utf-8')
 
-    # -----------------------------------  Common vehicle APIs ---------------------------------------------
+#-- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- - Common vehicle APIs -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -
```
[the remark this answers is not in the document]
ppueyor: I did not add those spaced, it should be something related to the clang script, I didn't even know they were there. We can delete them of course.

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
ppueyor: True!

Unreal/Plugins/AirSim/Source/PIPCamera.h
```diff
@@ -69,8 +95,9 @@ class AIRSIM_API APIPCamera : public ACameraActor
     UPROPERTY()
     TArray<UDetectionComponent*> detections_;
 
+    //CinemAirSim
     UPROPERTY()
-    UCameraComponent* camera_;
+    UCineCameraComponent* camera_;
```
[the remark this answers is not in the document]
ppueyor: What do you think is the best option to resolve this? Maybe a new entry in the settings.json could enable/disable all the cinematographic features. The thing is that some of the calls to the RPCLib are just enabled with this kind of camera. Is there an option to avoid these functions to be called if the embedded camera is not cinematographic?
Programatically, it would be very easy to implement. When the PIPCamera is embedded in the drones, we can introduce an if to introduce one or another kind of PIPCamera.

ppueyor:
Hi @zimmy87, I answered your suggestions/bug reports. I will fix everything once we decide what to do to integrate it.

No, I haven't seen that error of compilation in local builds. I'm using UE 4.26.1 in Ubuntu 20. Can this be the difference?

What are the next steps we should do? Should I wait for your instructions?

Unreal/Plugins/AirSim/Source/PIPCamera.h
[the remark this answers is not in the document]
zimmy87: Personally, I think the best option here would be to introduce a new boolean setting inside both the CameraDefaults element and in https://microsoft.github.io/AirSim/settings/#camera-settings. This new boolean setting could be stored inside each APIPCamera instance. For CinemAirSim-specific methods, I think it would be best to have these essentially be no-ops whenever the setting is set to a regular UCameraComponent. To communicate to the user that these methods aren't doing anything, we could output a log statement saying the method being run doesn't perform any operation on a UCameraComponent. Another possibility is to throw an exception, but personally I don't think this warrants stopping execution of the program, just messaging to the user that the methods aren't doing anything should be sufficient.

PythonClient/airsim/client.py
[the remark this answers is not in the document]
zimmy87: Our .clang-format configuration file only currently supports cpp, it shouldn't be reformatting python code. Are you by chance using a configuration defined elsewhere? Looks like this has been partially fixed in the latest revision, but I'm still seeing some changes in whitespace. I would like to see these changes in whitespace reverted before merging this.

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
zimmy87: fixed in latest revision

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
```diff
@@ -588,3 +595,188 @@ void APIPCamera::onViewModeChanged(bool nodisplay)
...
+{
+    std::vector<std::string> vector;
+    TArray<FNamedLensPreset> lens_presets = camera_->GetLensPresets();
+    for (FNamedLensPreset preset : lens_presets) {
+        std::ostringstream ss;
+        std::string name = (TCHAR_TO_UTF8(*preset.Name));
+
+        ss << "Name: " << name << ";\n\t MinFocalLength: " << preset.LensSettings.MinFocalLength << "; \t MaxFocalLength: " << preset.LensSettings.MaxFocalLength;
+        ss << "\n\t Min FStop: " << preset.LensSettings.MinFStop << "; \t Max Fstop: " << preset.LensSettings.MaxFStop;
+        //ss << "\n\t Diaphragm Blade Count: " << preset.LensSettings.DiaphragmBladeCount;
```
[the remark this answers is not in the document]
zimmy87: fixed in latest revision

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
```diff
@@ -588,3 +595,186 @@ void APIPCamera::onViewModeChanged(bool nodisplay)
...
+{
+    std::ostringstream ss;
+    ss << "Current Field Of View:\n\tHorizontal Field Of View: " << camera_->GetHorizontalFieldOfView() << ";\n\t Vertical Field Of View: " << camera_->GetVerticalFieldOfView();
+    return ss.str();
+}
+
+void APIPCamera::copyCameraSettingsToAllSceneCapture(UCameraComponent* camera)
+{
+    int image_count = static_cast<int>(Utils::toNumeric(ImageType::Count));
+    for (image_type >= 0) {
```
zimmy87, line 752: Have you tested this method? Unless I'm missing something, I think image_type needs to be initialized and  incremented/decremented. I've never seen a for loop with just a condition before.

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
```diff
@@ -588,3 +595,188 @@ void APIPCamera::onViewModeChanged(bool nodisplay)
...
+{
+    std::ostringstream ss;
+    ss << "Current Field Of View:\n\tHorizontal Field Of View: " << camera_->GetHorizontalFieldOfView() << ";\n\t Vertical Field Of View: " << camera_->GetVerticalFieldOfView();
+    return ss.str();
+}
+
+void APIPCamera::CopyCameraSettingsToAllSceneCapture(UCameraComponent* camera)
+{
+    int image_count = static_cast<int>(Utils::toNumeric(ImageType::Count));
+    for (int image_type = image_count - 1; image_type >= 0; image_type--) {
```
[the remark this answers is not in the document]
zimmy87: this comment no longer applies to the latest revision

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
```diff
@@ -588,3 +595,188 @@ void APIPCamera::onViewModeChanged(bool nodisplay)
...
+    }
+}
+
+void APIPCamera::CopyCameraSettingsToSceneCapture(UCameraComponent* src, USceneCaptureComponent2D* dst)
+{
+    if (src && dst) {
+        dst->SetWorldLocationAndRotation(src->GetComponentLocation(), src->GetComponentRotation());
+        dst->FOVAngle = src->FieldOfView;
+
+        FMinimalViewInfo CameraViewInfo;
```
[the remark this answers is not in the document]
zimmy87: camera_view_info has been fixed, but there's still some remaining locals in this method that need to be fixed: namely, SrcPPSettings, DstPPSettings, and DstWeightedBlendables

Unreal/Plugins/AirSim/Source/PIPCamera.h
```diff
@@ -39,6 +43,28 @@ class AIRSIM_API APIPCamera : public ACameraActor
...
     void onViewModeChanged(bool nodisplay);
 
+    //CinemAirSim methods
+    std::vector<std::string> getPresetLensSettings();
+    void setPresetLensSettings(std::string preset_string);
+    std::vector<std::string> getPresetFilmbackSettings();
+    void setPresetFilmbackSettings(std::string preset_string);
+    std::string getLensSettings();
+    std::string getFilmbackSettings();
+    float setFilmbackSettings(float sensor_width, float sensot_height);
```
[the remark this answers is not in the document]
zimmy87: fixed in latest revision

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
zimmy87: This is causing build errors in our Azure Pipelines on Windows:

```
C:/__w/13/s/Unreal/Environments/Blocks/Plugins/AirSim/Source/PIPCamera.cpp(752): error C2065: 'image_type': undeclared identifier
  C:/__w/13/s/Unreal/Environments/Blocks/Plugins/AirSim/Source/PIPCamera.cpp(752): error C2143: syntax error: missing ';' before ')'
  C:/__w/13/s/Unreal/Environments/Blocks/Plugins/AirSim/Source/PIPCamera.cpp(752): error C2143: syntax error: missing ';' before '{'
  C:/__w/13/s/Unreal/Environments/Blocks/Plugins/AirSim/Source/PIPCamera.cpp(752): error C2143: syntax error: missing ')' before '{'
  C:/__w/13/s/Unreal/Environments/Blocks/Plugins/AirSim/Source/PIPCamera.cpp(754): error C2065: 'image_type': undeclared identifier
  C:/__w/13/s/Unreal/Environments/Blocks/Plugins/AirSim/Source/PIPCamera.cpp(754): error C2660: 'APIPCamera::copyCameraSettingsToSceneCapture': function does not take 1 arguments
  C:\__w\13\s\Unreal\Environments\Blocks\Plugins\AirSim\Source\PIPCamera.h(65): note: see declaration of 'APIPCamera::copyCameraSettingsToSceneCapture'
```

as well as on Ubuntu:

```
/__w/13/s/Unreal/Environments/Blocks/Plugins/AirSim/Source/PIPCamera.cpp:752:10: error: use of undeclared identifier 'image_type'
    for (image_type >= 0) {
         ^
/__w/13/s/Unreal/Environments/Blocks/Plugins/AirSim/Source/PIPCamera.cpp:752:25: error: expected ';' in 'for' statement specifier
    for (image_type >= 0) {
                        ^
/__w/13/s/Unreal/Environments/Blocks/Plugins/AirSim/Source/PIPCamera.cpp:754:61: error: use of undeclared identifier 'image_type'
        copyCameraSettingsToSceneCapture(camera_, captures_[image_type]);
                                                            ^
```

zimmy87 (changes requested):
Additional comment regarding a build failure

zimmy87:
Hi @ppueyor, apologies for the delay in getting back to you; I've responded to the remaining feedback items; please let me know if you have any other questions/concerns. Regarding the build failure in TextureShuffleActor.cpp, I'm still seeing this with your latest revision, so this looks like a legitimate build failure. My guess is you're missing an include somewhere (usually the "incomplete type" error occurs when only a forward declaration is present for a given class and the full header for that class is needed to access one of its members). We're currently using 4.25.1 on our build machines, and we'd like to maintain compatibility with 4.25, so using 4.26.1 could be a potential cause here. Would you be able to test locally with a 4.25.1 build?

Unreal/Plugins/AirSim/Source/PIPCamera.h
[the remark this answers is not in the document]
ppueyor: The problem I see here is that the APIPCamera is an instance of the method UCineCameraComponent, and the header should be a child of: `class AIRSIM_API APIPCamera : public ACineCameraActor `. It is something that depends on the whole class, not only on one attribute of the class that we can select/deselect.

We can eventually have two separate classes, PIPCamera and PIPCinemaCamera, for instance, and attach to the drone one or another depending on the boolean set in the settings, when these cameras are created. What do you think?

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
ppueyor: My mistake, some lines should remain out of the code in the last commit, sorry for this. 

In my local version this method looks like this:

``` 
void APIPCamera::copyCameraSettingsToAllSceneCapture(UCameraComponent* camera)
{
   int image_count = static_cast<int>(Utils::toNumeric(ImageType::Count));
    for (int image_type = image_count - 1; image_type >= 0; image_type--) {
        if (image_type >= 0) { //scene capture components
            CopyCameraSettingsToSceneCapture(camera_, captures_[image_type]);
        }
    }
}
```

PythonClient/airsim/client.py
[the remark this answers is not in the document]
ppueyor: Okay, I was checking all the modified files. Updated!

ppueyor:
Hi @zimmy87, 

Thanks for the review. Yes, I will test it in Unreal 4.25.1 once we decide what to do with the cinematic flag in Settings.hpp.
I fixed the suggested changes. I will commit everything when it is all decided and ready.

Unreal/Plugins/AirSim/Source/PIPCamera.h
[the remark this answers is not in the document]
zimmy87: In that case, would it be possible to change the default settings of the UCineCameraComponent property to resemble those of a default UCameraComponent? We're not concerned about introducing a dependency on UCineCameraComponent or ACineCameraActor, it's moreso we just think it'll be confusing for users to see a camera with a different field of view or other characteristics than before.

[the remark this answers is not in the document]
ppueyor: I think that dependency is needed, so the camera on the actor is a CineCameraComponent instead of a CameraComponent, with all the fields and parameters that are only of that class, and in the way it generates a BluePrint with a CineCamera instead of a regular Camera.


What about two different PIPCameras? One that depends on CameraActor and another that depends on CineCameraActor.

ppueyor:
Hi @zimmy87, 

I implemented all the requested changes and updated my repo to the latest version of your master.
I talked to Jonny and we decided to set the focal length to match with the regular camera. 

I tested it with ubuntu 20 and Unreal 4.25 and works perfectly. 

Please, tell me if you need anything else.

Regards,
Pablo

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
```diff
@@ -370,6 +374,9 @@ void APIPCamera::setupCameraFromSettings(const APIPCamera::CameraSetting& camera
     else
         this->SetActorTickEnabled(false);
 
+    //set initial focal length
+    camera_->CurrentFocalLength = 11.9;
```
zimmy87, line 378: Can we move this to APIPCamera::PostInitializeComponents to ensure it happens for all PIPCamera instances, not just the ones attached to the vehicle pawn?

Currently, APIPCamera::setupCameraFromSettings is called from ASimModeBase::createVehicleApi on all cameras attached to the vehicle pawn, so I see the focal length get updated for all those cameras. But there's another camera attached to the CameraDirector that gets created in ASimModeBase::initializeCameraDirector and this camera ends up being the one used for the main game view. Since APIPCamera::setupCameraFromSettings never gets called on this camera, I still see the default camera have the default focal length of 35.

zimmy87 (changes requested):
The new focal length looks good when I manually override the right camera, but unfortunately I'm not seeing this focal length applied to the default camera, so I still see the same narrow field of view with the latest iteration.

AirLib/include/api/RpcLibClientBase.hpp
zimmy87: this has been addressed in latest revision

zimmy87:
Hi @ppueyor, thanks for the update, I tested locally and noticed a few things that I think should get addressed before merging. I've gone through all previous review feedback and resolved everything that looks to have been addressed. I see that there's one remaining item of feedback from @rajat2004 on PythonClient/airsim/client.py that hasn't been addressed yet (it looks like this may not require a code change, feel free to resolve the comment yourself if that's the case). There also looks like there's one remaining item of feedback from my previous review about the names of SrcPPSettings, DstPPSettings, and DstWeightedBlendables in Unreal/Plugins/AirSim/Source/PIPCamera.cpp. These variables should be renamed to src_pp_settings, dst_pp_settings, and dst_weighted_blendables.

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
```diff
@@ -605,3 +615,187 @@ void APIPCamera::onViewModeChanged(bool nodisplay)
...
+//CinemAirSim methods
+std::vector<std::string> APIPCamera::getPresetLensSettings()
+{
+    std::vector<std::string> vector;
+    TArray<FNamedLensPreset> lens_presets = camera_->GetLensPresets();
+    for (FNamedLensPreset preset : lens_presets) {
+        std::ostringstream ss;
+        std::string name = (TCHAR_TO_UTF8(*preset.Name));
+
+        ss << "Name: " << name << ";\n\t MinFocalLength: " << preset.LensSettings.MinFocalLength << "; \t MaxFocalLength: " << preset.LensSettings.MaxFocalLength;
```
rajat2004, line 628: Is storing the data in a string really the best way, makes it that much harder for users to extract and store data programmatically...
Upto you and the maintainers

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
```diff
@@ -605,3 +615,187 @@ void APIPCamera::onViewModeChanged(bool nodisplay)
...
+    FString preset(preset_string.c_str());
+    camera_->SetLensPresetByName(preset);
+    copyCameraSettingsToAllSceneCapture(camera_);
+}
+
+std::vector<std::string> APIPCamera::getPresetFilmbackSettings()
+{
+    std::vector<std::string> vector;
+    TArray<FNamedFilmbackPreset> lens_presets = camera_->GetFilmbackPresets();
+    for (FNamedFilmbackPreset preset : lens_presets) {
```
rajat2004, line 670: `const FNamedFilmbackPreset& preset`, to avoid unnecessary copies, in other places also

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
```diff
@@ -605,3 +615,187 @@ void APIPCamera::onViewModeChanged(bool nodisplay)
...
+    ss << "Name: " << name;
+    ss << ";\n\t MinFocalLength: " << current_lens_params.MinFocalLength;
+    ss << "; \t MaxFocalLength: " << current_lens_params.MaxFocalLength;
+    ss << "\n\t Min FStop: " << current_lens_params.MinFStop;
+    ss << "; \t Max Fstop: " << current_lens_params.MaxFStop;
+    ss << "\n\t Diaphragm Blade Count: " << current_lens_params.DiaphragmBladeCount;
+    ss << "\n\t Minimum focus distance: " << current_lens_params.MinimumFocusDistance;
+    std::string final_string(ss.str());
+
+    return final_string;
```
rajat2004, line 656: Can just directly return `ss.str()` here rather than extra local var which isn't gonna be used. In other places also

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
```diff
@@ -605,3 +615,187 @@ void APIPCamera::onViewModeChanged(bool nodisplay)
...
+        ss << "\n\t Min FStop: " << preset.LensSettings.MinFStop << "; \t Max Fstop: " << preset.LensSettings.MaxFStop;
+        std::string final_string(ss.str());
+        vector.push_back(final_string);
+    }
+    return vector;
+}
+
+std::string APIPCamera::getLensSettings()
+{
+    std::vector<std::string> vector;
```
rajat2004, line 638: Unused var, in other places also

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
```diff
@@ -605,3 +615,187 @@ void APIPCamera::onViewModeChanged(bool nodisplay)
         }
     }
 }
+
+//CinemAirSim methods
+std::vector<std::string> APIPCamera::getPresetLensSettings()
+{
+    std::vector<std::string> vector;
```
rajat2004, line 622: Instead of `vector`, should use more descriptive var names, maybe `all_settings` or any other appropriate name

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
```diff
@@ -605,3 +615,187 @@ void APIPCamera::onViewModeChanged(bool nodisplay)
...
+std::string APIPCamera::getLensSettings()
+{
+    std::vector<std::string> vector;
+
+    FCameraLensSettings current_lens_params = camera_->LensSettings;
+
+    std::ostringstream ss;
+
+    FString lens_preset_name = camera_->GetLensPresetName();
+    std::string name = (TCHAR_TO_UTF8(*lens_preset_name));
```
rajat2004, line 645: Try to add `const` to vars if they aren't modified, to avoid coding errors, and makes the intention clear, other places also

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
```diff
@@ -605,3 +615,187 @@ void APIPCamera::onViewModeChanged(bool nodisplay)
...
+{
+    camera_->Filmback.SensorWidth = sensor_width;
+    camera_->Filmback.SensorHeight = sensor_height;
+
+    copyCameraSettingsToAllSceneCapture(camera_);
+
+    return camera_->Filmback.SensorAspectRatio;
+}
+
+float APIPCamera::getFocalLength()
```
rajat2004, line 713: Such methods should also be marked as `const` to ensure and indicate to the reader they aren't modifying the state

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
```diff
@@ -605,3 +615,187 @@ void APIPCamera::onViewModeChanged(bool nodisplay)
...
+    for (int image_type = image_count - 1; image_type >= 0; image_type--) {
+        if (image_type >= 0) { //scene capture components
+            copyCameraSettingsToSceneCapture(camera_, captures_[image_type]);
+        }
+    }
+}
+
+void APIPCamera::copyCameraSettingsToSceneCapture(UCameraComponent* src, USceneCaptureComponent2D* dst)
+{
+    if (src && dst) {
```
rajat2004, line 781: Maybe a log message in the `else` block?

Unreal/Plugins/AirSim/Source/PIPCamera.h
```diff
@@ -39,6 +43,28 @@ class AIRSIM_API APIPCamera : public ACameraActor
...
+    void enableManualFocus(bool enable);
+    float getFocusDistance();
+    void setFocusDistance(float focus_distance);
+    float getFocusAperture();
+    void setFocusAperture(float focus_aperture);
+    void enableFocusPlane(bool enable);
+    std::string getCurrentFieldOfView();
+
+    void copyCameraSettingsToAllSceneCapture(UCameraComponent* camera);
+    void copyCameraSettingsToSceneCapture(UCameraComponent* src, USceneCaptureComponent2D* dst);
```
rajat2004, line 65: These should be protected or private methods since they're meant to be used only by other member methods

rajat2004 (commented):
Some code improvement comments. Additionally, would be great if you could add some example scripts to test the various capabilities, and also some documentation.
Thanks!

PythonClient/airsim/client.py
ppueyor: This is from the master branch!

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
ppueyor: This is just informative. Users can use this method to extract all the preset lens available, so they can choose it with "setPresetLensSettings".

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
ppueyor: Moved to APIPCamera::PostInitializeComponents. Can you test if it works in that particular case?

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
ppueyor: For now I haven't added anything, thus it is just for developers. If you have some suggestions of what we can log, then I can edit it.

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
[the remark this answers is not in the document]
ppueyor: Got it! You're absolutely right.

Unreal/Plugins/AirSim/Source/PIPCamera.cpp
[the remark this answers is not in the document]
ppueyor: you're right, sorry for this

ppueyor:
Hi @zimmy87 and @rajat2004 , I resolved all the changes requested, hope everything is correct now. 
There is some docu/demo work to do, but I don't have much time to do it right now. I will do it asap.

ppueyor:
> The new focal length looks good when I manually override the right camera, but unfortunately I'm not seeing this focal length applied to the default camera, so I still see the same narrow field of view with the latest iteration.

Does the "PostProcesing" thing resolves this?

zimmy87 (approved):

zimmy87:
Tested the latest iteration and it works well for me, so I am moving ahead with merging. It would be nice to add some extra documentation and examples, but I feel that can be handled in a separate PR.

PR #3966, opened 2021-08-17
CMake and building/installing improvements

elevir:
## About
This PR improve building/installation using CMake. The things were added:
- installing 3rdparty libraries using pure cmake
- 3rdparty libraries will not install if they were already found by `find_package` (it may be useful for vcpkg users for example)
- building on windows using clang-cl compiler additionally to microsoft compiler
- updated json.hpp (previous version was not be able to compile using clang on windows)
- powershell script for installing libraries (powershell is a cross-platform shell, so it should be nice to maintain just one script for building and installing purposes)
- installation (libraries and executables can be now installed by `cmake --install`)

## How Has This Been Tested?
Tested on Windows using:
- clang 11.1
- clang 12
- cl (x64)

Tested on Ubuntu 18.04 WSL2 using:
- clang-13
- gcc-8

Now it can be tested by invoking cmake --configure/--build/--install. ~~Build scripts may not work.~~

Features tested:
- build and install airsim libraries with 3rd party libraries

Features not tested:
- ~~CI~~
- ~~additional packages and tools (such as ros, etc.)~~


elevir:
It's almost done but I need logs from CI.

elevir:
It seems to be done, but I have a couple of questions/notes.
1. I wanted to make a powershell script for a manual/ci building. But in Windows ps1 scripts execution may require setting execution policy (something like `pwsh -ExecutionPolicy UnRestricted -File .\install.ps1`). Also it will require changes in CI pipelines. For now this PR contains install.ps1 which builds and installs AirLib, MavLinkCom and 3rdparty libraries (if needed). Powershell scripts are cross-platform and it could be really convenient to use only one powershell script instead of two: bat and sh scripts. And the question is: can I replace all the bat/sh scripts by ps1 scripts?
2. I noticed that the source files using includes such as `#include "<airlib_subfolder0>/<airlib_subfolder1>/.../"` instead of `#include "AirLib/<airlib_subfolder0>/<airlib_subfolder1>/..."` it may lead to conflicts with AirLib user`s code, in this PR for compatibility reasons the both variants are valid.

elevir:
Sorry for long pause, I had no time to continue. I've merged last commits to this branch.
Last full check shown some errors with ros build. I've checked this version on my machine and it builds fine, so I need another full check run.

PR #3976, opened 2021-08-22
Add ROS2 wrapper

alonfaraj:
Fixes: https://github.com/microsoft/AirSim/discussions/3830 

## About

Hi,

I found myself using AirSim & ROS but a lot of our logic and code is already implemented with ROS2.
I migrated the existing ROS wrapper to ROS2, with minimal changes as possible.

One major change is a new package contains all messages and services - `airsim_interfaces`.

**Nodes:**
- [x] airsim_node
- [x] pid_position_node 

**Packages:**
- [x] airsim_interfaces
- [x] airsim_ros_pkgs

Not all launch files were migrated, only the main so far - 
- [x] airsim_node.launch.py
- [x] static_transforms.launch.py
- [x] position_controller_simple.launch.py
- [x] airsim_with_simple_PD_position_controller.launch.py
- [x] dynamic_constraints.launch.py
- [x] rviz.launch.py

This PR probably need much more tests to check that everything work like ROS1 wrapper.
I thought it might be useful to share it with you guys, because ROS2 became very popular.

## How Has This Been Tested?

Test configuration:
- ROS2 Foxy/Galactic
- Ubuntu 20.04

It might work on different ROS2/Ubuntu versions as well.
Didn't check it.

Ran this ROS2 wrapper node.
Tested on one Multirotor and one Car with different sensors using ros2 command tool and Rviz2

## How to test it yourself?
- Install gcc >= 8.0.0:
` sudo apt-get install gcc-8 g++-8
  `Verify installation by `gcc-8 --version`
- install ROS2 pkgs:
 `sudo apt-get install ros-${ROS_DISTRO}-tf2-sensor-msgs ros-${ROS_DISTRO}-tf2-geometry-msgs ros-${ROS_DISTRO}-mavros* libyaml-cpp-dev`
- Verify you have colcon installed, if not - install it:
https://colcon.readthedocs.io/en/released/user/installation.html
- ` cd /AirSim/ros2`
- `colcon build --cmake-args -DCMAKE_C_COMPILER=gcc-8 --cmake-args -DCMAKE_CXX_COMPILER=g++-8 --cmake-args -DCMAKE_BUILD_TYPE=Debug` 
or for release `DCMAKE_BUILD_TYPE=Release`
- `source install/setup.bash`
- Run:
`ros2 launch airsim_ros_pkgs airsim_node.launch.py`


ros2/src/airsim_interfaces/package.xml
```diff
@@ -0,0 +1,24 @@
...
+  <maintainer email="a@a.com">AAA</maintainer>
+  <license>Apache License 2.0</license>
+
+  <author email="a@a.com">AAA</author>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
+  <buildtool_depend>rosidl_default_generators</buildtool_depend>
+
+  <depend>action_msgs</depend>
```
jmackay2, line 15: This also depends on `std_msgs`, `geometry_msgs`, and `geographic_msgs`. I don't think there is actually a dependence on `action_msgs`.

jmackay2:
The messages could have been kept in the same package. Having them in their own package is a little cleaner, but makes it a little further out of alignment compared to the ROS1 package. Maybe that's fine.

ros2/src/airsim_interfaces/package.xml
alonfaraj: you right, thanks!

alonfaraj:
@jmackay2 Thanks for the review!
You probably right about keeping the messages in the same package, although I think it's better to separate in ROS1 package as well.
Actually, my first try was to keep them in the same ROS2 package but I didn't manage to build it properly using the ament build system, so at some point I decided to give up and separate it to a different package.

jmackay2:
> @jmackay2 Thanks for the review!
> You probably right about keeping the messages in the same package, although I think it's better to separate in ROS1 package as well.
> Actually, my first try was to keep them in the same ROS2 package but I didn't manage to build it properly using the ament build system, so at some point I decided to give up and separate it to a different package.

You were probably just missing the required `rosidl_target_interfaces(<lib/exe name> ${PROJECT_NAME} "rosidl_typesupport_cpp")` in your CMakeLists mentioned [here](https://docs.ros.org/en/foxy/Tutorials/Single-Package-Define-And-Use-Interface.html#link-against-the-interface).

ros2/src/airsim_ros_pkgs/CMakeLists.txt
```diff
@@ -0,0 +1,163 @@
...
+set(CXX_EXP_LIB
+  "-nostdinc++ -I/usr/include/c++/8 -I/usr/include/x86_64-linux-gnu/c++/8 -nodefaultlibs 
+-l/usr/lib/x86_64-linux-gnu/libc++.so -l/usr/lib/x86_64-linux-gnu/libc++abi.so
+-lm -lc -lgcc_s -lgcc -g    
+-lstdc++fs -fmax-errors=10 -Wnoexcept -Wstrict-null-sentinel")
+set(RPC_LIB_INCLUDES " ${AIRSIM_ROOT}/external/rpclib/rpclib-2.3.0/include")
+set(RPC_LIB rpc)
+message(STATUS "found RPC_LIB_INCLUDES=${RPC_LIB_INCLUDES}")
+
+set(INCLUDE_DIRS include ${rclcpp_INCLUDE_DIRS} ${nav_msgs_INCLUDE_DIRS}
```
jmackay2, line 57: I would remove all of the ROS related packages from here. The include directories should only need `include`, things in `AIRSIM_ROOT`, `OpenCV`, and `Boost.

ros2/src/airsim_ros_pkgs/CMakeLists.txt
```diff
@@ -0,0 +1,163 @@
...
+  ${sensor_msgs_LIBRARIES} ${mavros_msgs_LIBRARIES}
+  ${rosidl_default_generators_LIBRARIES} ${ament_cmake_LIBRARIES}
+  ${std_msgs_LIBRARIES}
+  ${airsim_interfaces_LIBRARIES})
+
+find_package(Boost REQUIRED)
+find_package(OpenCV REQUIRED)
+
+add_library(airsim_settings_parser src/airsim_settings_parser.cpp)
+target_link_libraries(airsim_settings_parser ${LIBS} AirLib)
```
jmackay2, line 100: I don't think anything in `LIBS` is required for this library.
```suggestion
target_link_libraries(airsim_settings_parser AirLib)
```

ros2/src/airsim_ros_pkgs/CMakeLists.txt
```diff
@@ -0,0 +1,163 @@
...
+
+add_library(pd_position_controller_simple src/pd_position_controller_simple.cpp)
+target_link_libraries(pd_position_controller_simple ${LIBS} AirLib)
```
jmackay2, line 103: Instead of `target_link_libraries`, for linking to ROS packages `ament_target_dependencies` should be used.
```
ament_target_dependencies(pd_position_controller_simple rclcpp nav_msgs)
```

alonfaraj:
> > @jmackay2 Thanks for the review!
> > You probably right about keeping the messages in the same package, although I think it's better to separate in ROS1 package as well.
> > Actually, my first try was to keep them in the same ROS2 package but I didn't manage to build it properly using the ament build system, so at some point I decided to give up and separate it to a different package.
> 
> You were probably just missing the required `rosidl_target_interfaces(<lib/exe name> ${PROJECT_NAME} "rosidl_typesupport_cpp")` in your CMakeLists mentioned [here](https://docs.ros.org/en/foxy/Tutorials/Single-Package-Define-And-Use-Interface.html#link-against-the-interface).

I have tried this (it's still commented out  in CMakeLists) but it didn't work for me.
Maybe I did something wrong.

ros2/src/airsim_ros_pkgs/CMakeLists.txt
alonfaraj: fixed

ros2/src/airsim_ros_pkgs/CMakeLists.txt
alonfaraj: fixed

ros2/src/airsim_ros_pkgs/CMakeLists.txt
alonfaraj: fixed

rajat2004 (commented):
The airsim_ros_wrapper.cpp/.h files also seem to have complete whitespace changes, maybe unintended?

rajat2004:
> > The airsim_ros_wrapper.cpp/.h files also seem to have complete whitespace changes, maybe unintended?
> 
> Can you point me where?
> I'm not sure I understand what you mean

Sorry for the late reply, in the diff on Github, all the lines in those files have been modified, without any increase or decrease of lines, and if you tick the "Hide Whitspace" option, it all goes away. I think it's a line-ending modification, from LF to CR LF

ros2/src/airsim_ros_pkgs/include/geodetic_conv.hpp
```diff
@@ -0,0 +1,227 @@
+// full credits to https://github.com/ethz-asl/geodetic_utils
```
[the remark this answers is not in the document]
rajat2004: Makes sense, the ROS 1 change can be done separately

ros2/src/airsim_ros_pkgs/CMakeLists.txt
```diff
@@ -0,0 +1,159 @@
...
+add_subdirectory("${AIRSIM_ROOT}/cmake/rpclib_wrapper" rpclib_wrapper)
+add_subdirectory("${AIRSIM_ROOT}/cmake/AirLib" AirLib)
+add_subdirectory("${AIRSIM_ROOT}/cmake/MavLinkCom" MavLinkCom)
+
+set(CMAKE_CXX_STANDARD 14)
+set(CXX_EXP_LIB
+  "-nostdinc++ -I/usr/include/c++/8 -I/usr/include/x86_64-linux-gnu/c++/8 -nodefaultlibs 
+-l/usr/lib/x86_64-linux-gnu/libc++.so -l/usr/lib/x86_64-linux-gnu/libc++abi.so
+-lm -lc -lgcc_s -lgcc -g    
+-lstdc++fs -fmax-errors=10 -Wnoexcept -Wstrict-null-sentinel")
```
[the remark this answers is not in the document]
rajat2004: Thanks a lot! I'll work on the warnings in my PR as well

alonfaraj:
> > > The airsim_ros_wrapper.cpp/.h files also seem to have complete whitespace changes, maybe unintended?
> > 
> > 
> > Can you point me where?
> > I'm not sure I understand what you mean
> 
> Sorry for the late reply, in the diff on Github, all the lines in those files have been modified, without any increase or decrease of lines, and if you tick the "Hide Whitspace" option, it all goes away. I think it's a line-ending modification, from LF to CR LF

All of these files have been added, not changed, so I think that's why.
They placed under ros2 folder, not ros, so I believe that's the reason.
They not compared to the old ros files.

ros2/src/airsim_ros_pkgs/CMakeLists.txt
```diff
@@ -0,0 +1,159 @@
...
+
+add_executable(pd_position_controller_simple_node
+  src/pd_position_controller_simple_node.cpp)
+target_link_libraries(pd_position_controller_simple_node
+  pd_position_controller_simple airsim_ros AirLib)
+ament_target_dependencies(pd_position_controller_simple_node 
+  rclcpp
+)
+
+# rosidl_target_interfaces(airsim_node
```
jmackay2, line 114: To have the messages in the same package, instead of `airsim_node` and `pd_position_controller_simple_node` you would need `airsim_ros` and `pd_position_controller_simple`, which are the ones that need the defined messages and services. I made a [PR here](https://github.com/alonfaraj/AirSim/pull/4) with the required fixes that would be required to align ROS 1 and ROS 2.

ros2/src/airsim_ros_pkgs/CMakeLists.txt
```diff
@@ -0,0 +1,159 @@
...
+set(LIBS ${rclcpp_LIBRARIES} ${nav_msgs_LIBRARIES} ${geographic_msgs_LIBRARIES}
+  ${std_srvs_LIBRARIES} ${tf2_ros_LIBRARIES} ${tf2_sensor_msgs_LIBRARIES}
+  ${rclpy_LIBRARIES} ${tf2_LIBRARIES} ${image_transport_LIBRARIES}
+  ${geometry_msgs_LIBRARIES} ${cv_bridge_LIBRARIES} ${tf2_geometry_msgs_LIBRARIES}
+  ${sensor_msgs_LIBRARIES} ${mavros_msgs_LIBRARIES}
+  ${rosidl_default_generators_LIBRARIES} ${ament_cmake_LIBRARIES}
+  ${std_msgs_LIBRARIES}
+  ${airsim_interfaces_LIBRARIES})
+
+find_package(Boost REQUIRED)
```
jmackay2, line 72: Boost is not required anymore.
```suggestion
```

ros2/src/airsim_ros_pkgs/CMakeLists.txt
```diff
@@ -0,0 +1,159 @@
...
+find_package(OpenCV REQUIRED)
```
jmackay2, line 73: The `find_package` for OpenCV should be done prior to the use of `OpenCV_INCLUDE_DIRS`. I am actually not sure why this does not cause a build failure.

ros2/src/airsim_ros_pkgs/CMakeLists.txt
```diff
@@ -0,0 +1,159 @@
...
+  ${AIRSIM_ROOT}/AirLib/deps/eigen3
+  ${AIRSIM_ROOT}/AirLib/include
+  ${RPC_LIB_INCLUDES}
+  ${AIRSIM_ROOT}/MavLinkCom/include
+  ${AIRSIM_ROOT}/MavLinkCom/common_utils
+  ${OpenCV_INCLUDE_DIRS}
+  ${Boost_INCLUDE_DIRS})
+include_directories(${INCLUDE_DIRS})
+
+set(LIBRARY_DIRS ${rclcpp_LIBRARY_DIRS} ${nav_msgs_LIBRARY_DIRS}
```
jmackay2, line 51: Setting the `LIBRARY_DIRS` should be deleted, as well as the `link_directories` and setting `LIBS`. The `target_link_libraries` and `ament_target_dependencies` is already doing all of the library linking for us.

ros2/src/airsim_ros_pkgs/CMakeLists.txt
jmackay2: Related, `Boost_INCLUDE_DIRS` should also be deleted.

alonfaraj: fixed

ros2/src/airsim_ros_pkgs/CMakeLists.txt
alonfaraj: Thanks, I wasn't aware of that

ros2/src/airsim_ros_pkgs/CMakeLists.txt
alonfaraj: Thank you for reviewing and helping!
The best practice in ROS is to keep interfaces in a separate package.
I believe it would be best to change ROS1 to work that way as well. Do you not agree?

jmackay2: Yep, I was just trying to figure out why you originally had an issue with having them in the same package originally. You can go ahead and mark this as resolved.

alonfaraj: Thank you very much!
I will merge your PR if we decide to align it with ROS1 and not keep these packages separated.

ros2/src/airsim_ros_pkgs/src/airsim_ros_wrapper.cpp
```diff
@@ -0,0 +1,1486 @@
...
+            auto rpc = static_cast<msr::airlib::MultirotorRpcLibClient*>(airsim_client_.get());
+            drone->curr_drone_state = rpc->getMultirotorState(vehicle_ros->vehicle_name);
+
+            vehicle_time = airsim_timestamp_to_ros(drone->curr_drone_state.timestamp);
+            if (!got_sim_time) {
+                curr_ros_time = vehicle_time;
+                got_sim_time = true;
+            }
+
+            vehicle_ros->gps_sensor_msg = get_gps_sensor_msg_from_airsim_geo_point(drone->curr_drone_state.gps_location);
```
rajat2004, line 996: Is there a reason for the different `drone->curr_drone_state.gps_location` vs `env_data.geo_point` in the Car?

ros2/src/airsim_ros_pkgs/src/airsim_ros_wrapper.cpp
```diff
@@ -0,0 +1,1486 @@
...
+}
+
+std::shared_ptr<sensor_msgs::msg::Image> AirsimROSWrapper::get_img_msg_from_response(const ImageResponse& img_response,
+                                                                                     const rclcpp::Time curr_ros_time,
+                                                                                     const std::string frame_id)
+{
+    unused(curr_ros_time);
+    std::shared_ptr<sensor_msgs::msg::Image> img_msg_ptr = std::make_shared<sensor_msgs::msg::Image>(); //boost::make_shared<sensor_msgs::msg::Image>();
+    img_msg_ptr->data = img_response.image_data_uint8;
+    img_msg_ptr->step = img_response.width * 3; // todo un-hardcode. image_width*num_bytes
```
rajat2004, line 1321: `image_data_uint8.size() / height` might work

ros2/src/airsim_ros_pkgs/src/airsim_ros_wrapper.cpp
```diff
@@ -0,0 +1,1486 @@
...
+        is_used_lidar_timer_cb_queue_ = true;
+    }
+
+    initialize_airsim();
+}
+
+template <typename T>
+const SensorPublisher<T> AirsimROSWrapper::create_sensor_publisher(const string& sensor_type_name, const string& sensor_name, SensorBase::SensorType sensor_type, const string& topic_name, int QoS)
+{
+    std::cout << sensor_type_name << std::endl;
```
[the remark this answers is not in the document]
alonfaraj: You are right, we should use the ROS logging system

ros2/src/airsim_ros_pkgs/src/airsim_ros_wrapper.cpp
alonfaraj: CarState doesn't have gps_location, but MultirotorState has.
As a general rule, I simply migrated the code in ROS2 to make it work, without digging into it too much.
It probably worth do it for both ROS1 & ROS2

ros2/src/airsim_ros_pkgs/scripts/car_joy
```diff
@@ -0,0 +1,150 @@
...
+
+        update_rate_hz = rospy.get_param('~update_rate_hz', 20.0)
+        self.max_curvature = rospy.get_param('~max_curvature', 0.75)
+        self.steer_sign = rospy.get_param('~steer_sign', -1)
+        self.throttle_brake_sign = rospy.get_param('~throttle_brake_sign', 1)
+        self.auto_gear_max = rospy.get_param('~auto_gear_max', 5)
+        self.manual_transmission = rospy.get_param('~manual_transmission', True)
+        self.forward_btn_index = rospy.get_param('~forward_button_index', 0)
+        self.reverse_btn_index = rospy.get_param('~reverse_button_index', 1)
+        self.nuetral_btn_index = rospy.get_param('~nuetral_button_index', 2)
```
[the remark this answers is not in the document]
alonfaraj: It is the original ROS1 script (unused here), so I will take care of it when I migrate it.

ros2/src/airsim_ros_pkgs/src/airsim_ros_wrapper.cpp
```diff
@@ -0,0 +1,1486 @@
...
+    isENU_ = !(odom_frame_id_ == AIRSIM_ODOM_FRAME_ID);
+    nh_->get_parameter_or("coordinate_system_enu", isENU_, isENU_);
+    vel_cmd_duration_ = 0.05; // todo rosparam
+    // todo enforce dynamics constraints in this node as well?
+    // nh_->get_parameter("max_vert_vel_", max_vert_vel_);
+    // nh_->get_parameter("max_horz_vel", max_horz_vel_)
+
+    nh_->declare_parameter("vehicle_name", rclcpp::ParameterValue(""));
+    create_ros_pubs_from_settings_json();
+    airsim_control_update_timer_ = nh_->create_wall_timer(std::chrono::duration<double>(update_airsim_control_every_n_sec), std::bind(&AirsimROSWrapper::drone_state_timer_cb, this));
```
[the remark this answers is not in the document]
alonfaraj: I think that ROS2 best practices as well the [official tutorials](https://docs.ros.org/en/foxy/Tutorials/Writing-A-Simple-Cpp-Publisher-And-Subscriber.html) using `std::bind` so I think it's better stick to it.

ros2/src/airsim_ros_pkgs/src/airsim_ros_wrapper.cpp
```diff
@@ -0,0 +1,1486 @@
...
+
+    // ros params
+    double update_airsim_control_every_n_sec;
+    nh_->get_parameter("is_vulkan", is_vulkan_);
+    nh_->get_parameter("update_airsim_control_every_n_sec", update_airsim_control_every_n_sec);
+    nh_->get_parameter("publish_clock", publish_clock_);
+    nh_->get_parameter_or("world_frame_id", world_frame_id_, world_frame_id_);
+    odom_frame_id_ = world_frame_id_ == AIRSIM_FRAME_ID ? AIRSIM_ODOM_FRAME_ID : ENU_ODOM_FRAME_ID;
+    nh_->get_parameter_or("odom_frame_id", odom_frame_id_, odom_frame_id_);
+    isENU_ = !(odom_frame_id_ == AIRSIM_ODOM_FRAME_ID);
```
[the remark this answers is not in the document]
alonfaraj: agreed, cleaner

AmitNativ1984:
Hi, to run with ros2:galactic I had to add: `#include <tf2_ros/buffer.h>` to `airsim_ros_wrapper.h`. Otherwise build failed

zimmy87 (changes requested):
some style-related comments

zimmy87:
Thank you for this contribution @alonfaraj! this looks like a good change to make. There appears to be some remaining feedback on this PR from @rajat2004 that's still valid for the current version. Could you address these items?

ros2/src/airsim_ros_pkgs/include/airsim_ros_wrapper.h
```diff
@@ -0,0 +1,399 @@
...
+    //             yaw_mode = msr::airlib::YawMode();};
+
+    // VelCmd(const double& x, const double& y, const double& z,
+    //         msr::airlib::DrivetrainType drivetrain,
+    //         const msr::airlib::YawMode& yaw_mode,
+    //         const std::string& vehicle_name) :
+    //     x(x), y(y), z(z),
+    //     drivetrain(drivetrain),
+    //     yaw_mode(yaw_mode),
+    //     vehicle_name(vehicle_name) {};
```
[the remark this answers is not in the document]
alonfaraj: You right.
The commented out code are leftovers from ROS1 sources.

ros2/src/airsim_ros_pkgs/launch/airsim_car_with_joy_control_auto.launch
```diff
@@ -0,0 +1,27 @@
+<launch>
+	<arg name="host" default="localhost" />
+	
+	<node name="airsim_node" pkg="airsim_ros_pkgs" type="airsim_node" output="screen">
+	<!-- <node pkg="nodelet" type="nodelet" name="airsim_nodelet_manager" args="manager" output="screen" /> -->
+	<!-- <node pkg="nodelet" type="nodelet" name="airsim_nodelet" args="standalone airsim_ros_pkgs/airsim_ros_nodelet airsim_nodelet_manager" output="screen"> -->
```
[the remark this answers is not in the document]
alonfaraj: Actually these launch files are ROS1 files.
I copy paste them and edit in place.
There are 3 left, so better to remove them and maybe migrate them in the future.

ros2/src/airsim_ros_pkgs/include/airsim_ros_wrapper.h
```diff
@@ -0,0 +1,399 @@
...
+#include <tf2_ros/buffer.h>
+#include <tf2_ros/static_transform_broadcaster.h>
+#include <tf2_ros/transform_broadcaster.h>
+#include <tf2_ros/transform_listener.h>
+#include <tf2/convert.h>
+#include <unordered_map>
+#include <memory>
+
+    // todo move airlib typedefs to separate header file?
+    typedef msr::airlib::ImageCaptureBase::ImageRequest ImageRequest;
```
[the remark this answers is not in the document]
alonfaraj: looks like the clang-format causes the extra indentation.
I'm not sure why.

ros2/src/airsim_ros_pkgs/src/airsim_ros_wrapper.cpp
```diff
@@ -0,0 +1,1486 @@
...
+    state_msg.header.stamp = airsim_timestamp_to_ros(car_state.timestamp);
+
+    return state_msg;
+}
+
+nav_msgs::msg::Odometry AirsimROSWrapper::get_odom_msg_from_car_state(const msr::airlib::CarApiBase::CarState& car_state) const
+{
+    nav_msgs::msg::Odometry odom_msg;
+
+    odom_msg.pose.pose.position.x = car_state.getPosition().x();
```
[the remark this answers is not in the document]
alonfaraj: I moved some duplications to separate method.
I will mark it as resolve for now, let me now if you see something specific which can be improved.

ros2/src/airsim_ros_pkgs/src/airsim_ros_wrapper.cpp
```diff
@@ -0,0 +1,1486 @@
...
+        airsim_client_images_.confirmConnection();
+        airsim_client_lidar_.confirmConnection();
+
+        for (const auto& vehicle_name_ptr_pair : vehicle_name_ptr_map_) {
+            airsim_client_->enableApiControl(true, vehicle_name_ptr_pair.first); // todo expose as rosservice?
+            airsim_client_->armDisarm(true, vehicle_name_ptr_pair.first); // todo exposes as rosservice?
+        }
+
+        origin_geo_point_ = airsim_client_->getHomeGeoPoint("");
+        // todo there's only one global origin geopoint for environment. but airsim API accept a parameter vehicle_name? inside carsimpawnapi.cpp, there's a geopoint being assigned in the constructor. by?
```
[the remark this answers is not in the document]
alonfaraj: Fixed, thanks for the suggestion.

ros2/src/airsim_ros_pkgs/src/airsim_ros_wrapper.cpp
```diff
@@ -0,0 +1,1486 @@
...
+        origin_geo_point_msg_ = get_gps_msg_from_airsim_geo_point(origin_geo_point_);
+    }
+    catch (rpc::rpc_error& e) {
+        std::string msg = e.get_error().as<std::string>();
+        std::cout << "Exception raised by the API, something went wrong." << std::endl
+                  << msg << std::endl;
```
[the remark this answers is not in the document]
alonfaraj: probably not.
I will shutdown the node in case of failure.

ros2/src/airsim_ros_pkgs/src/airsim_ros_wrapper.cpp
```diff
@@ -0,0 +1,1486 @@
...
+        lidar_msg.is_bigendian = false;
+        lidar_msg.point_step = offset; // 4 * num fields
+        lidar_msg.row_step = lidar_msg.point_step * lidar_msg.width;
+
+        lidar_msg.is_dense = true; // todo
+        std::vector<float> data_std = lidar_data.point_cloud;
+
+        const unsigned char* bytes = reinterpret_cast<const unsigned char*>(data_std.data());
+        vector<unsigned char> lidar_msg_data(bytes, bytes + sizeof(float) * data_std.size());
+        lidar_msg.data = std::move(lidar_msg_data);
```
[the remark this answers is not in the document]
alonfaraj: Hmm yep.
That's the original code from ros1 so I resolve it for now.
Let me know if you have another suggestion.

alonfaraj:
Thank you for your feedback @zimmy87!
I tried to address all your feedbacks, as well as @rajat2004's.

ros2/src/airsim_ros_pkgs/src/airsim_ros_wrapper.cpp
```diff
@@ -0,0 +1,1446 @@
...
+                    if (capture_setting.image_type == 0 || capture_setting.image_type == 5 || capture_setting.image_type == 6 || capture_setting.image_type == 7) {
+                        current_image_request_vec.push_back(ImageRequest(curr_camera_name, curr_image_type, false, false));
+                    }
+                    // if {DepthPlanar, DepthPerspective,DepthVis, DisparityNormalized}, get float image
+                    else {
+                        current_image_request_vec.push_back(ImageRequest(curr_camera_name, curr_image_type, true));
+                    }
+
+                    image_pub_vec_.push_back(image_transporter.advertise("~/" + curr_vehicle_name + "/" + curr_camera_name + "/" + image_type_int_to_string_map_.at(capture_setting.image_type), 1));
+                    cam_info_pub_vec_.push_back(nh_->create_publisher<sensor_msgs::msg::CameraInfo>("~/" + curr_vehicle_name + "/" + curr_camera_name + "/" + image_type_int_to_string_map_.at(capture_setting.image_type) + "/camera_info", 10));
```
rajat2004, line 218: Maybe have `"~/" + curr_vehicle_name + "/" + curr_camera_name + "/" + image_type_int_to_string_map_.at(capture_setting.image_type)` as a separate string

ros2/src/airsim_ros_pkgs/src/airsim_ros_wrapper.cpp
```diff
@@ -0,0 +1,1446 @@
...
+    // nh_->get_parameter("max_vert_vel_", max_vert_vel_);
+    // nh_->get_parameter("max_horz_vel", max_horz_vel_)
+
+    nh_->declare_parameter("vehicle_name", rclcpp::ParameterValue(""));
+    create_ros_pubs_from_settings_json();
+    airsim_control_update_timer_ = nh_->create_wall_timer(std::chrono::duration<double>(update_airsim_control_every_n_sec), std::bind(&AirsimROSWrapper::drone_state_timer_cb, this));
+    // airsim_control_update_timer_ = nh_->createTimer(rclcpp::Duration (update_airsim_control_every_n_sec), std::bind(&AirsimROSWrapper::drone_state_timer_cb, this));
+}
+
+// XmlRpc::XmlRpcValue can't be const in this case
```
rajat2004, line 116: Relevant?

rajat2004 (commented):
Just some small nitpicks, thanks a lot for cleaning up the code duplication from the ROS1 code!

alonfaraj:
Thanks @rajat2004!
It's much cleaner.
Fixed now.

tools/Dockerfile-ROS2
```diff
@@ -0,0 +1,20 @@
+FROM ros:foxy-ros-base
+
+ARG UNAME=testuser
+
+RUN     DEBIAN_FRONTEND=noninteractive apt-get update &&\
+        apt-get install -y\
+        apt-utils \
+        gcc-8 g++-8 \
+        ros-$ROS_DISTRO-tf2-sensor-msgs ros-$ROS_DISTRO-tf2-geometry-msgs ros-$ROS_DISTRO-mavros* \
```
rajat2004, line 9: `ROS_DISTRO` doesn't seem to be defined here

rajat2004 (commented):
One more thing which can be done later is to add it in Actions CI, to ensure that it compiles going forward. Docker container can also be used directly here to build

alonfaraj:
You right, I guess it should be done after the PR will be merged

tools/Dockerfile-ROS2
alonfaraj: Do you get any error?
Because it seems to work fine for me.
ROS_DISTRO is defined automatically after installing ros.

rajat2004: Ah okay, I'm currently not on Linux so can't test, thanks for the info!

zimmy87:
hey @alonfaraj, I tried following your steps to test the wrapper on Ubuntu locally, but I'm receiving the following compile time error:

```
/home/zimmy/Downloads/AirSim/ros2/build/airsim_interfaces/rosidl_generator_c/airsim_interfaces/msg/detail/environment__struct.h:27:10: fatal error: geographic_msgs/msg/detail/geo_point__struct.h: No such file or directory
   27 | #include "geographic_msgs/msg/detail/geo_point__struct.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

I tried remedying this by installing the `ros-galactic-geographic-msgs` package, but apt-get says I'm already at the latest version:

```
ros-galactic-geographic-msgs is already the newest version (1.0.4-5focal.20211011.201914).
```

Do you know if there's a another dependency I'm missing, or if there's a different gepgraphic-msgs package I need?

jmackay2:
> hey @alonfaraj, I tried following your steps to test the wrapper on Ubuntu locally, but I'm receiving the following compile time error:
> 
> ```
> /home/zimmy/Downloads/AirSim/ros2/build/airsim_interfaces/rosidl_generator_c/airsim_interfaces/msg/detail/environment__struct.h:27:10: fatal error: geographic_msgs/msg/detail/geo_point__struct.h: No such file or directory
>    27 | #include "geographic_msgs/msg/detail/geo_point__struct.h"
>       |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> compilation terminated.
> ```
> 
> I tried remedying this by installing the `ros-galactic-geographic-msgs` package, but apt-get says I'm already at the latest version:
> 
> ```
> ros-galactic-geographic-msgs is already the newest version (1.0.4-5focal.20211011.201914).
> ```
> 
> Do you know if there's a another dependency I'm missing, or if there's a different gepgraphic-msgs package I need?

I tested with the `ros:galactic` docker image and had to install the following:
```
apt install ros-galactic-cv-bridge ros-galactic-image-transport ros-galactic-geographic-msgs ros-galactic-mavros-msgs libyaml-cpp-dev
```
and it built fine. 

Was ros sourced before building (i.e. `source /opt/ros/galactic/setup.bash`)?

zimmy87:
Hey @alonfaraj and @jmackay2, thank you for the help. I figured out what was wrong with my ros2 installation: I was following the instructions [here](https://docs.ros.org/en/galactic/Installation/Ubuntu-Install-Binary.html) which installs from a downloaded tar.bz2 and then I was trying to use apt to install the rest of the dependencies. This ended up creating two different ros2 installations that didn't share the same packages. After uninstalling everything and following the instructions [here](https://docs.ros.org/en/galactic/Installation/Ubuntu-Install-Debians.html) instead, I was able to get the ros2 wrapper building.

What services do we expect to be supported in this version of the ros2 wrapper? I tried calling both `/airsim_node/SimpleFlight/takeoff` and `/airsim_node/local_position_goal`, but both seem to be crashing the airsim_node. Here's my repro steps:

1. install ROS2 via steps [here](https://docs.ros.org/en/galactic/Installation/Ubuntu-Install-Debians.html)
2. follow test steps in first post of this PR
3. start airsim_node via `ros2 launch airsim_ros_pkgs airsim_node.launch.py`
4. in a 2nd terminal, start pd_position_controller_simple_node via `ros2 launch airsim_ros_pkgs position_controller_simple.launch.py`
5. in a 3rd terminal, execute either `ros2 service call /airsim_node/SimpleFlight/takeoff airsim_interfaces/srv/Takeoff "{wait_on_last_task: True}"` or `ros2 service call /airsim_node/local_position_goal airsim_interfaces/srv/SetLocalPosition "{x: 1.0, y: 1.0, z: -5.0}"`

expected: drone should move (either to takeoff or to target)
actual: airsim_node crashes with following error message:

```
[ERROR] [airsim_node-2]: process has died [pid 27214, exit code -6, cmd '/home/zimmy/Downloads/AirSim/ros2/install/airsim_ros_pkgs/lib/airsim_ros_pkgs/airsim_node --ros-args -r __node:=airsim_node --params-file /tmp/launch_params_lxyomrpc'].
```

Am I doing something wrong here or is this outside of the intended scope of this PR (in which case do we have a list somewhere of currently functional services so users know which ones to avoid calling)?

alonfaraj:
Hi @zimmy87, glad to hear you solved it!
You are doing everything right, I found those crashes 2 days ago and already fixed it here https://github.com/microsoft/AirSim/pull/3976/commits/8f5f6603573e63811f1ba1134d0ef42dc81c9802.
I guess you not on the latest code in the PR.
Let me know if that's fix it for you.

All the topic/services/nodes in ROS1 are supported in a similar way in this ROS2 PR, except `airsim_tutorial_pkgs `and `airsim_car_with_joy_control.launch`

zimmy87:
Thank you for the contribution @alonfaraj! I was able to get the scenario I mentioned previously working after updating to your latest commit. I think the current level of functionality is sufficient for merging. There's a couple things I noticed when testing that I think may be good things to investigate for any future PRs on this wrapper:

-  The `success` value in every response to my service calls was always set to False irregardless of whether the drone appeared to complete the desired action.
- If I call the `/airsim_node/SimpleFlight/land` service after a call to `/airsim_node/local_position_goal`, the drone will move back to it's previous target position after landing.
- I need to set `SimMode` to `Multirotor` in settings.json in order to launch `airsim_node.launch.py`. Even if I don't set this and select no in the "Choose Vehicle" dialog to select Multirotor mode manually, I will still get an error when launching airsim_node:

```
[airsim_node-2] terminate called after throwing an instance of 'std::invalid_argument'
[airsim_node-2]   what():  Unknown SimMode: , failed to set default vehicle settings
[ERROR] [airsim_node-2]: process has died [pid 15235, exit code -6, cmd '/home/zimmy/Downloads/AirSim/ros2/install/airsim_ros_pkgs/lib/airsim_ros_pkgs/airsim_node --ros-args -r __node:=airsim_node --params-file /tmp/launch_params_3l9lh0m8'].
```

zimmy87 (approved):

zimmy87:
As per the behaviors I called out for future investigation, I didn't realize at the time that these were all occurring in the ROS1 wrapper as well, so these aren't being introduced in this PR. Feel free to disregard as these are just inherited from the ROS1 wrapper and should be addressed there as well.

alonfaraj:
@zimmy87 yep, you right. 
I have seen you already fixed the services response in ros1. 

Regarding the third behavior you mentioned - this is the same behavior in ros1 as well. 
The idea is that when the node starts, it gets the settings.json file from the host as string and parse it like the simulation itself does. 
So, assuming you started the simulation without settings.json, and create drone by clicking yes in the popup, the settings file not get updated, so the ros node has no idea which vehicle are you using because it's empty and not contain SimMode.

Robotgir:
Hi @alonfaraj @zimmy87 , i would like to run basic example with the combination of ros2, airsim and UE.
The steps i followed are as follows..
1. steps from [this](https://microsoft.github.io/AirSim/airsim_ros_pkgs/) , with heading "Using Docker for ROS[#](https://microsoft.github.io/AirSim/airsim_ros_pkgs/#using-docker-for-ros)"
- instead of using the prescribed AirSim/tools/Dockerfile-ROS for ros1 i used AirSim/tools/Dockerfile-ROS2 and i have built the docker image, which i am able to successfully run.
2. I am able to launch airsim_node using 
- ros2 launch airsim_ros_pkgs airsim_node.launch.py
3. when attempting to launch position_controller_simple.launch.py as mentioned by @zimmy87  in https://github.com/microsoft/AirSim/pull/3976#issuecomment-985485519 i see in the output terminal "service not available waiting again", i figured out that this error pops up as the below line fails, line 60 in AirSim/ros2/src/airsim_ros_pkgs/src/pd_position_controller_simple.cpp fails..
4. 
59. auto parameters_client = std::make_shared<rclcpp::SyncParametersClient>(nh_, "/airsim_node");
60. while (!parameters_client->wait_for_service(std::chrono::seconds(1))) {
...
}

- From what i understand std::make_shared<rclcpp::SyncParametersClient>() is basically loading parameters from node /airsim_node into the current node handle nh_, correct me if i am wrong..!
- to check if the parameters from /airsim_node are available i used the command ros2 param list whlie airsim_node is already launched, but it is just stuck there and doesnt print anythink out

Any help would be great !
Thanks in advance!

PR #3978, opened 2021-08-23
Add gazebo airsim plugin files

Jaeyoung-Lim:
Fixes: #    

Related to: https://github.com/microsoft/AirSim/issues/2955

## About
While the Gazebo environment is the most widely used environments for simulation, the rendering quality of gazebo makes it hard to test vision based applications. While [Microsoft/Airsim](https://github.com/microsoft/AirSim) solves this problem by using advanced game engines such as the Unreal Engine or Unity the physics Airsim can simulate are limited.
- Only multirotors are available, and none of the vehicles have gimbals
- Aerodynamic effects such as Fixedwing/VTOL vehicles are not available for PX4 SITL-Airsim

This PR provides a solution mid-way so that the capabilities not available in Unreal Airsim can be used with Gazebo. 
A `gazebo_airsim_plugin` is implemented as a link between gazebo and Airsim. The plugin can be attached to a specific link that orients the camera in Airsim. The specific link can be defined in the sdf flile as the following.
```
    <plugin name='airsim_plugin' filename='libgazebo_airsim_plugin.so'>
      <robotNamespace/>
      <link_name>typhoon_h480::cgo3_camera_link</link_name>
    </plugin>
```

This plugin assumes that Airsim is running in "Computer Vision Mode". The configuration file of Airsim is as the following.
```
{
  "SettingsVersion": 1.2,
  "SimMode": "ComputerVision"
}
```

## How Has This Been Tested?
This has been integrated into the PX4 Autopilot project, and tested using the following command. (Needs inclusion of https://github.com/PX4/PX4-SITL_gazebo/pull/791)
```
make px4_sitl gazebo_typhoon_h480_airsim
```

[GazeboAirsim](https://youtu.be/dwvVRZZmTX0 "Circular trajectory tracking")

[GazeboAirsim](https://youtu.be/eoL4pxmz0Bc "Circular trajectory tracking")

### DIFF
--- a/GazeboDrone/CMakeLists.txt
+++ b/GazeboDrone/CMakeLists.txt
@@ -36,4 +36,4 @@ target_link_libraries(
 )
 
 add_library(gazebo_airsim_plugin SHARED src/gazebo_airsim_plugin.cpp)
-target_link_libraries(gazebo_airsim_plugin ${GAZEBO_LIBRARIES} AirLib rpc MavLinkCom)
+target_link_libraries(gazebo_airsim_plugin ${GAZEBO_LIBRARIES} AirLib rpc)
--- a/GazeboDrone/CMakeLists.txt
+++ b/GazeboDrone/CMakeLists.txt
@@ -11,6 +11,7 @@ endif()
 find_package(gazebo REQUIRED)
 
 include_directories(
+    include
     ${GAZEBO_INCLUDE_DIRS}
     ${AIRSIM_ROOT}/AirLib/deps/eigen3
     ${AIRSIM_ROOT}/AirLib/deps/rpclib/include
@@ -34,3 +35,5 @@ target_link_libraries(
     rpc
 )
 
+add_library(gazebo_airsim_plugin SHARED src/gazebo_airsim_plugin.cpp)
+target_link_libraries(gazebo_airsim_plugin ${GAZEBO_LIBRARIES} AirLib rpc MavLinkCom)
--- a/GazeboDrone/src/gazebo_airsim_plugin.cpp
+++ b/GazeboDrone/src/gazebo_airsim_plugin.cpp
@@ -52,24 +52,24 @@ AirsimPlugin::AirsimPlugin()
 
 AirsimPlugin::~AirsimPlugin()
 {
-    _updateConnection->~Connection();
+    updateConnection_->~Connection();
 }
 
-void AirsimPlugin::Load(physics::ModelPtr _model, sdf::ElementPtr _sdf)
+void AirsimPlugin::Load(physics::ModelPtr model, sdf::ElementPtr sdf)
 {
-    model_ = _model;
+    model_ = model;
     world_ = model_->GetWorld();
 
     namespace_.clear();
-    if (_sdf->HasElement("robotNamespace")) {
-        namespace_ = _sdf->GetElement("robotNamespace")->Get<std::string>();
+    if (sdf->HasElement("robotNamespace")) {
+        namespace_ = sdf->GetElement("robotNamespace")->Get<std::string>();
     }
     else {
         gzerr << "[gazebo_catapult_plugin] Please specify a robotNamespace.\n";
     }
 
-    if (_sdf->HasElement("link_name")) {
-        sdf::ElementPtr elem = _sdf->GetElement("link_name");
+    if (sdf->HasElement("link_name")) {
+        sdf::ElementPtr elem = sdf->GetElement("link_name");
         std::string linkName = elem->Get<std::string>();
         this->link_ = this->model_->GetLink(linkName);
 
@@ -82,7 +82,7 @@ void AirsimPlugin::Load(physics::ModelPtr _model, sdf::ElementPtr _sdf)
         gzerr << "[gazebo_airsim_plugin] linkName needs to be defined in the model SDF";
     }
 
-    _updateConnection = event::Events::ConnectWorldUpdateBegin(boost::bind(&AirsimPlugin::OnUpdate, this, _1));
+    updateConnection_ = event::Events::ConnectWorldUpdateBegin(boost::bind(&AirsimPlugin::OnUpdate, this, _1));
 
     node_handle_ = transport::NodePtr(new transport::Node());
     node_handle_->Init(namespace_);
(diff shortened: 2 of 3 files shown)


Jaeyoung-Lim:
@jonyMarino I tried using `clang-format-10` to format my code, but the settings doesn't seem to agree with what is being checked in CI.

Is there a good way to format the code that is appropriate for this repo?

jonyMarino:
1. check out your pull request

1. run `git checkout master .clang-format` in your repo's root folder

1. if you don't have npm installed, download it from https://www.npmjs.com/get-npm

1. run `npm install -g clang-format`

1. run `clang-format -i path/to/file` for each failing file (failing files can be found in the details for the clang-format check)

1. run `git add -u` to track all style changes

1. run `git commit -am "apply style from clang-format"`

1. run `git merge master`

1. resolve other non-style conflicts if you have any

1. push changes

I hope it helps you!

GazeboDrone/CMakeLists.txt
```diff
@@ -34,3 +35,5 @@ target_link_libraries(
     rpc
 )
 
+add_library(gazebo_airsim_plugin SHARED src/gazebo_airsim_plugin.cpp)
+target_link_libraries(gazebo_airsim_plugin ${GAZEBO_LIBRARIES} AirLib rpc MavLinkCom)
```
jonyMarino, line 39: MavLinkCom is not needed for this project

Jaeyoung-Lim: Removed!

Jaeyoung-Lim:
@jonyMarino Thanks for the pointers! I have addressed the style check failures

rajat2004 (commented):
Just some small nitpicks

zimmy87:
Hey @Jaeyoung-Lim, I'm trying to test this PR with the command you mentioned:

```
make px4_sitl gazebo_typhoon_h480_airsim
```

However, I'm getting the following build errors:

```
ninja: error: unknown target 'gazebo_typhoon_h480_airsim'
make: *** [Makefile:230: px4_sitl] Error 1
```

I have your other PR https://github.com/PX4/PX4-SITL_gazebo/pull/791 checked out inside Tools/sitl_gazebo in my local PX4-Autopilot repo. Do I have to do something else to pick up the new target you added?

Jaeyoung-Lim:
@zimmy87 You would also need https://github.com/PX4/PX4-Autopilot/pull/18566 which adds the make target

GazeboDrone/src/gazebo_airsim_plugin.cpp
```diff
@@ -0,0 +1,115 @@
...
+#if GAZEBO_MAJOR_VERSION >= 9
+    ignition::math::Pose3d T_W_I = link_->WorldPose();
+#else
+    ignition::math::Pose3d T_W_I = ignitionFromGazeboMath(link_->GetWorldPose());
+#endif
+    ignition::math::Vector3d pos = T_W_I.Pos();
+    ignition::math::Quaterniond att = T_W_I.Rot();
+
+    msr::airlib::Vector3r p(pos.X(), -pos.Y(), -pos.Z()); // Convert to NED since airsim position is in NED
+    msr::airlib::Quaternionr o(att.W(), att.X(), -att.Y(), -att.Z()); // Convert to NED since airsim position is in NED
```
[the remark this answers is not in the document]
Jaeyoung-Lim: No, the negative signs are actually for converting the attitude into the NED reference frame

GazeboDrone/src/gazebo_airsim_plugin.cpp
```diff
@@ -0,0 +1,115 @@
...
+        if (!this->link_) {
+            gzerr << "[gazebo_airsim_plugin] Link with name[" << linkName << "] not found. "
+                  << "Airsim plugin will not be able to launch the vehicle\n";
+        }
+    }
+    else {
+        gzerr << "[gazebo_airsim_plugin] linkName needs to be defined in the model SDF";
+    }
+
+    _updateConnection = event::Events::ConnectWorldUpdateBegin(boost::bind(&AirsimPlugin::OnUpdate, this, _1));
```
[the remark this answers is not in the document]
Jaeyoung-Lim: @rajat2004 Not sure if I follow, how could you do this through a lambda?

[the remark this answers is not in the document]
rajat2004: Something like -
```cpp
event::Events::ConnectWorldUpdateBegin([&](const auto& info) { OnUpdate(info); })
```
Anyways, just a personal pref, I think in ROS `boost::bind` is used more so whatever seems best to you

GazeboDrone/src/gazebo_airsim_plugin.cpp
```diff
@@ -0,0 +1,112 @@
...
+    node_handle_ = transport::NodePtr(new transport::Node());
+    node_handle_->Init(namespace_);
+
+    client_->confirmConnection();
+}
+
+void AirsimPlugin::OnUpdate(const common::UpdateInfo&)
+{
+    if (link_ == nullptr)
+        gzthrow("[gazebo_airsim_plugin] Couldn't find specified link \n");
```
rajat2004, line 96: Should it just throw in `Load` itself rather than adding the check here?

Jaeyoung-Lim: @rajat2004 I think this needs to be checked at runtime, in case the link has not been loaded yet or has been removed at runtime

rajat2004: Oh, it was being set only in `Load()` hence felt it might be more appropriate there itself, didn't think it could get modified later on

Jaeyoung-Lim: @rajat2004 I realized that, and hence I edited my comments :sweat_smile: I think you are right.

rajat2004: Didn't see the edited comment, if it allows other plugins to continue then yeah, makes sense, thanks for the explanation!

jonyMarino:
@Jaeyoung-Lim Thanks for continuing to work on this. Can we specifically use Multirotor instead of CV and call [simSetKinematics](https://github.com/microsoft/AirSim/blob/f7fe3997a3457776bbe836a28a62b37b24939430/AirLib/include/api/RpcLibClientBase.hpp#L131) instead of simSetVehiclePose? That way we can get the sensors from AirSim and it may be that it improves the behavior of the camera as well.

Jaeyoung-Lim:
@jonyMarino Could you elaborate on where the sensor data from the airsim is being used? This is not being used from the simulation or the SITL autopilot instance

jonyMarino:
@Jaeyoung-Lim It is convenient for the AirSim user who wants to know for the photos taken what the kinematics state was at that time. This way, you could get everything from the AirSim API in a synchronized way without communicating with Gazebo.
It has no impact on the flight controller or the physics of the vehicle.
At the same time, Frederic from Auterion (I don't know his user name, you can ping him) reported that the camera has a better behavior using the multirotor mode with the External Physics Engine if the kinematics state is updated through simSetKinematics().

PR #4030, opened 2021-09-17
Fix the consistency of H and W defined in function arguments in two scripts

YingkaiFu:
Fix the inconsistency of H and W defined in function arguments in two scripts

Fixes: #

## About
The existing event simulator plugin works well when use --debug parameter to visualize events, however, due to the inconsistent order of H and W in two scripts, when use --save parameter, the saved pkl format events is abnormal, this PR fixes the order of H and W to save desired events.

## Screenshots
The left window show the visualization of the events that saved using the fixed code, the right window show the original  visualization of the saved events.

comparation
:

### DIFF
--- a/PythonClient/eventcamera_sim/event_simulator.py
+++ b/PythonClient/eventcamera_sim/event_simulator.py
@@ -67,8 +67,8 @@ def esim(
 
         current_time = last_time
         for i in range(spike_nums):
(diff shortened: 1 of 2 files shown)


zimmy87:
Thanks for the contribution @YingkaiFu! I had some trouble pushing to your branch, so I merged these changes as part of #4052

Issue #3852, opened 2021-07-05
Shaky behavior when setting continuous camera orientations


ppueyor:
## Bug report 

- AirSim Version/#commit: 1.5
- UE/Unity version: 4.26
- autopilot version:
- OS Version: Ubuntu 20

### What's the issue you encountered?
To avoid aggressive changes in the camera rotations, I implemented a loop where it increments the value of the yaw and pitch softly and calls the function "client.simSetCameraPose()" every 0.05 secs.

The result is that, from time to time, there are glimpses that make the recording very shaky. 

A representative video can be found here: https://drive.google.com/file/d/1TNUsiQthfrNsQh__ObHO0iygBJBGn9y3/view?usp=sharing

### How can the issue be reproduced?
Code using C++ API: https://pastebin.com/BbpDcLeW

jonyMarino:
Hey Pablo! Can you share a code ready to be compiled? that would make things easier for us. Thanks!

zimmy87:
Is it intentional to have the call to `std::this_thread::sleep_for()` outside of your while loop in your code snippet? It looks like there may not be any delay in between your calls to `client.simSetCameraPose()`.

ppueyor:
Hi Zimmy. It is intentional. It is inside a thread that moves the gimbal of the camera softly to a determinate value of pitch, yaw. I want to have a soft camera rotation that increments slowly every X milliseconds. In this way, the footage will look continuous.

Jony, I created a demo code where it is easy to reproduce this. The gimbal stabilization is set to true in the settings, maybe it is related to this problem.

Code: https://pastebin.com/SLHDAkQi
Video (more noticeable from sec 13): https://drive.google.com/file/d/1EEit30PeFTEckco6hxEdxgvONpx5Bs6C/view?usp=sharing

Thanks for answering!

ppueyor:
Hey guys!
Any insights about this bug?

ppueyor:
Hi again guys, 

I found the bug that was causing that problem. It was an if that was just set for the gimbal stabilization mode in PIPCamera.cpp.

To fix it, it is as simple as to add the closing brackets to that line (281-286):

> if (gimbal_stabilization_ > 0) {
        gimbald_rotator_.Pitch = rotator.Pitch;
        gimbald_rotator_.Roll = rotator.Roll;
        gimbald_rotator_.Yaw = rotator.Yaw;
    }
    this->SetActorRelativeRotation(rotator);

to 

[...]

PR #4037, opened 2021-09-23
Fixed camera bug

ppueyor:
Fixes: #    
Method called when it is not supposed to. Fixed with and else to exclude the call when gimbal stabilization is not set to 1.

## About
It fixes a shaky behaviour when using gimbal stabilization as shown in #3852 .

## How Has This Been Tested?
Ubuntu 20 and UE4 4.26
Compiled and used without problems


jonyMarino:
Hi @ppueyor! You are including the commits for the cinematographic camera together with this fix. Can you create a new PR with just the latest commit, please?

ppueyor:
True! I messed up the branches and repos. It is fixed now. I created a new PR in #4061 

I'm sorry for the inconvenience.

Thanks!!

PR #4061, opened 2021-10-05
Fixed camera bug when using gimbal stabilization

ppueyor:
Fixes: #    
Method called when it is not supposed to. Fixed with and else to exclude the call when gimbal stabilization is not set to 1.

## About
It fixes a shaky behavior when using gimbal stabilization as shown in #3852 .

## How Has This Been Tested?
Ubuntu 20 and UE4 4.26
Compiled and used without problems

### DIFF
--- a/Unreal/Plugins/AirSim/Source/PIPCamera.cpp
+++ b/Unreal/Plugins/AirSim/Source/PIPCamera.cpp
@@ -283,7 +283,9 @@ void APIPCamera::setCameraPose(const msr::airlib::Pose& relative_pose)
         gimbald_rotator_.Roll = rotator.Roll;
         gimbald_rotator_.Yaw = rotator.Yaw;
     }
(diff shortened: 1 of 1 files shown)


zimmy87:
Hey @ppueyor, thanks for submitting this fix. Unfortunately, we're having trouble repro'ing the original issue using the sample code you included in #3852. Even with gimbal stabilization enabled, I don't see any shaky behavior with or without this fix. Would you mind sharing the settings you're using to test this fix in case there's additional settings that impact the original issue's repro?

zimmy87:
Hi @ppueyor, we're having trouble accessing https://pastebin.pl/view/76ebcf26, can you post the contents of your settings in a comment here?

zimmy87:
I was able to get a repro of #3852 with the settings posted in the last comment. I was also able to confirm that it gets fixed after applying this fix, so I am moving ahead with merging.

PR #4145, opened 2021-11-05
moveByRollPitchYawThrottleAsync Documentation changed from radians to Degrees

Gastastrophe:
Fixes: #    

Fixed the documentation for moveByRollPitchYawThrottleAsync to reflect that the function uses degrees instead of Radians when operating on a PX4 stack.

## About

After some testing, I have found that moveByRollPitchYawThrottleAsync uses degrees instead of radians for the target orientation in the body frame while simulating with HITL. The documentation says it uses radians, so I have updated it.

## How Has This Been Tested?

I called the moveByRollPitchYawThrottleAsync function with values in ranges [-pi/2, pi/2] and [-90, 90], and I discovered that only the [-90, 90] range produced the expected changes in the drone's orientation in a PX4 HITL Unreal simulation.

### DIFF
--- a/PythonClient/airsim/client.py
+++ b/PythonClient/airsim/client.py
@@ -1175,7 +1175,7 @@ class MultirotorClient(VehicleClient, object):
     def moveByRollPitchYawThrottleAsync(self, roll, pitch, yaw, throttle, duration, vehicle_name = ''):
         """
         - Desired throttle is between 0.0 to 1.0
-        - Roll angle, pitch angle, and yaw angle are given in **degrees**, in the body frame.
+        - Roll angle, pitch angle, and yaw angle are given in **degrees** when using PX4 and in **radians** when using SimpleFlight, in the body frame.
         - The body frame follows the Front Left Up (FLU) convention, and right-handedness.
 
         - Frame Convention:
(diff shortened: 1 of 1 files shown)


jonyMarino:
@Gastastrophe very well.
This may be a difference between SimpleFlight and PX4. Could you test if the same happens using SimpleFlight?

Gastastrophe:
> @Gastastrophe very well. This may be a difference between SimpleFlight and PX4. Could you test if the same happens using SimpleFlight?

Good catch! I tested on SimpleFlight, and that one is in Radians. My PX4 is a Pixhawk 4 mini that I set up with QGroundControl according to the instructions, so I imagine this is a common issue for PX4 users.

jonyMarino:
@Gastastrophe Thanks for the test. If the Sensors and the other controllers use radians, we should change SimpleFlight to use radians. So I believe that We should convert all the pitch roll yaw parameters from [these methods](https://github.com/microsoft/AirSim/blob/f7fe3997a3457776bbe836a28a62b37b24939430/AirLib/include/vehicles/multirotor/firmwares/simple_flight/SimpleFlightApi.hpp#L213-L289) from degrees to radians. Also, We should take out all the calls to Utils::degreesToRadians. It is not a minor change; also, we should change examples and maybe more documentation. And everything needs to be tested. In the meantime, I would clarify (in the API reference that you have modified (but for all the methods that have pith, roll, and yaw parameters)) that simpleflight uses degrees, while PX4 and ardupilot (Am I right @rajat2004?) use radians. @Gastastrophe do you want to clarify that in this document so you can own your first commit?

Gastastrophe:
@jonyMarino I have updated the documentation and my pull request accordingly

rajat2004:
ArduPilot currently doesn't support controlling moving the vehicle through the AirSim APIs hence it won't matter. Best to be consistent across firmwares, but yeah, changing this will be a big behaviour change. Maybe it can be done in a major release v2.0?

PR #4109, opened 2021-10-23
Fix actor name conflict for multiple ComputerVision vehicles

rajat2004:
Fixes: #4092    

## About

Fixes the name conflict by including the vehicle name

## How Has This Been Tested?

With the below settings from #4092 
```json
{
    "SettingsVersion": 1.2,
    "SimMode": "ComputerVision",
    "Vehicles": {
        "Cam1": {
            "VehicleType": "ComputerVision",
            "X": 26,
            "Y": -1,
            "Z": -10
        },
        "Cam2": {
            "VehicleType": "ComputerVision",
            "X": 26,
            "Y": -2,
            "Z": -20
        }
    }
}
```

## Screenshots (if appropriate):

PR #4203, opened 2021-12-09
Fix simAddVehicle for Car vehicle

rajat2004:
Fixes: #    

## About

`simAddVehicle` API failed with Car with the error in https://github.com/microsoft/AirSim/pull/4109#issuecomment-950307936
This PR fixes the extra `reset` call, allowing it to work for Car, CV vehicles

Alos update the docs for this API

## How Has This Been Tested?

Using the script added in the PR -

https://user-images.githubusercontent.com/37938604/145699690-3cf360ba-d6b7-41a6-ae19-ee2cdc551f2e.mp4

For CV mode -
```python
>>> import airsim
>>> cl = airsim.VehicleClient()
>>> cl.confirmConnection()
Connected!

>>> pose = airsim.Pose()
>>> print(cl.simAddVehicle("CV2", "ComputerVision", pose))
True
```

Verified that HelloSpawnedDrones was working as expected

Docs -
Screenshot from 2021-12-12 21-19-23


zimmy87:
Tested this locally and it works for me, so I am moving ahead with merging; thanks for the contribution @rajat2004!

PR #4205, opened 2021-12-10
set response.success values in ROS wrapper

zimmy87:
Fixes: #    

## About

This updates all service callbacks to return response.success = true.

## How Has This Been Tested?

Tested with all of the services detailed in https://microsoft.github.io/AirSim/airsim_ros_pkgs/


jonyMarino:
Thanks @zimmy87!

PR #4237, opened 2021-12-22
keep CollisionInfo.has_collided flag set until simGetCollisionInfo called

### LINKED ISSUE #1004: has_collided not updated
When collision occurs the _has_collided_ sometimes remains set to **False** even though the values of for example the _impact_point_ vector is updated. The below shows two readings of _collision_info_, the first one before any collision and the second one after the first collision. 

`
<CollisionInfo> {   'has_collided': False,
    'impact_point': <Vector3r> {   'x_val': 0.0,
    'y_val': 0.0,
    'z_val': 0.0},
    'normal': <Vector3r> {   'x_val': 0.0,
    'y_val': 0.0,
    'z_val': 0.0},
    'object_id': -1,
    'object_name': '',
    'penetration_depth': 0.0,
    'position': <Vector3r> {   'x_val': 0.0,
    'y_val': 0.0,
    'z_val': 0.0},
    'time_stamp': 0}

<CollisionInfo> {   'has_collided': False,
    'impact_point': <Vector3r> {   'x_val': 7.556523323059082,
    'y_val': -3.422746419906616,
    'z_val': -4.774749755859375},
    'normal': <Vector3r> {   'x_val': -2.086162567138672e-06,
    'y_val': 0.9999998807907104,
    'z_val': 0.0},
    'object_id': 181,
    'object_name': 'wall3',
    'penetration_depth': 9.999999974752427e-07,
    'position': <Vector3r> {   'x_val': 7.556523323059082,
    'y_val': -3.422746419906616,
    'z_val': -4.774749755859375},
    'time_stamp': 1524477282691448320}
`

zimmy87:
Fixes: #1004    

## About

This adds a new method to PawnSimApi called getCollisionInfoAndReset() which both returns collision info as well as reset the has_collided flag. The old behavior for resetting CollisionInfo.has_collided after every clock tick has been removed and simGetCollisionInfo now calls getCollisionInfoAndReset(), which means this method will return whether any collision has occurred since last call to simGetCollisionInfo instead of only within the last clock tick.

## How Has This Been Tested?

Collisions were tested in the Blocks environment with the EnableCollisionPassthrogh parameter set to both true and false.

### DIFF
--- a/AirLib/include/api/VehicleSimApiBase.hpp
+++ b/AirLib/include/api/VehicleSimApiBase.hpp
@@ -58,6 +58,7 @@ namespace airlib
         virtual const msr::airlib::Environment* getGroundTruthEnvironment() const = 0;
 
         virtual CollisionInfo getCollisionInfo() const = 0;
+        virtual CollisionInfo getCollisionInfoAndReset() = 0;
         virtual int getRemoteControlID() const = 0; //which RC to use, 0 is first one, -1 means disable RC (use keyborad)
         virtual RCData getRCData() const = 0; //get reading from RC from simulator's host OS
         virtual std::string getVehicleName() const = 0;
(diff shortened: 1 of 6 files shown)


jonyMarino:
Thanks, @zimmy87!

PR #4266, opened 2022-01-04
Fix trace toggling for multiple vehicles

### LINKED ISSUE #4260: In multi-drone emulation, pressing T toggle Trace line only works on one drone
### Overview:

In the simulation of multip-drones, press the T key to toggle traceline and only one drone changes. Traceline of the remaining drones cannot be hidden or display

### Nature of Request:  

- Addition
- Change
- Removal
  - 

I'm doing multi-drone simulations right now, but I can only show the trajectory of one drone at the moment. 
At present, I collect the location of uav and then use simPlotPoints, but it is very inconvenient . I think it's necessary to be able to switch the trajectories of multiple drones.  thanks

rajat2004:
Fixes: #4260     

## About

Now tracing gets enabled for all the vehicles 
TODO
- [x] Cleanup of SimHUD code
- [ ] Support tracing in Car vehicle (Possibly in a separate PR)

## How Has This Been Tested?

Using the `Pythonclient/multirotor/multi_agent_drone.py` script and settings
Screenshot from 2022-01-04 07-34-39

### DIFF
--- a/Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.h
+++ b/Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.h
@@ -34,13 +34,6 @@ public:
     void inputEventToggleSubwindow2();
     void inputEventToggleAll();
 
-    ImageType getSubwindowCameraType(int window_index);
-    void setSubwindowCameraType(int window_index, ImageType type);
-    APIPCamera* getSubwindowCamera(int window_index);
-    void setSubwindowCamera(int window_index, APIPCamera* camera);
-    bool getSubwindowVisible(int window_index);
-    void setSubwindowVisible(int window_index, bool is_visible);
-
     ASimHUD();
     virtual void BeginPlay() override;
     virtual void EndPlay(const EEndPlayReason::Type EndPlayReason) override;
@@ -51,6 +44,7 @@ protected:
     void toggleRecordHandler();
     void updateWidgetSubwindowVisibility();
     bool isWidgetSubwindowVisible(int window_index);
+    void toggleSubwindowVisibility(int window_index);
 
 private:
     void initializeSubWindows();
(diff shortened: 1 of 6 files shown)


rajat2004:
@zimmy From #4274 looks like a new release is going to happen soon. Is it possible to have this PR included as well?

Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.h
```diff
@@ -34,13 +34,6 @@ class AIRSIM_API ASimHUD : public AHUD
     void inputEventToggleSubwindow2();
     void inputEventToggleAll();
 
-    ImageType getSubwindowCameraType(int window_index);
-    void setSubwindowCameraType(int window_index, ImageType type);
```
zimmy87, line 38: getSubwindowCameraType, getSubwindowCameraType, getSubwindowVisible, and setSubwindowVisible are also defined [here](https://github.com/microsoft/AirSim/blob/master/Unity/AirLibWrapper/AirsimWrapper/Source/SimHUD/SimHUD.cpp#L46). Is there any reason to not clean up the Unity wrapper as well?

zimmy87 (changes requested):
Thanks for fixing this this @rajat2004! Tested locally and this works for me just fine. Left one comment about cleaning up the Unity wrapper. Will talk with the team about whether to include this in the next release.

Unreal/Plugins/AirSim/Source/SimHUD/SimHUD.h
rajat2004: Main reason was that I currently don't have a working Unity installation, going through that code, none of the subwindow code in Unity SimHUD.cpp seems to be actually used, and the subwindows are created in AirsimHUDScript.cs
Will try to go through it a bit more

zimmy87: I have a working Unity installation, so I decided to commit this cleanup change after testing that subwindows still functioned normally in my Unity build. Once I get this merged, I'll include it in the next release.

zimmy87 (approved):

zimmy87:
Thanks for submitting this PR @rajat2004! Moving ahead with merging.

rajat2004:
Thanks a lot @zimmy87 for the Unity changes! I'll get my old setup working over the weekend, sorry for this

PR #4305, opened 2022-01-20
fix clock speed in car mode

### LINKED ISSUE #4232: Wrong timestamp when using small `ClockSpeed`
## Bug report 

- AirSim Version/#commit: 3.1
- UE/Unity version: 4.18
- autopilot version:
- OS Version: Windows 11

It seems that `getCarState().timestamp` isn't affected by the `ClockSpeed` settings. 

Running a simulation for 10 seconds under ClockSpeed of 0.5 - which means that the actual simulation time should be 5 seconds, doesn't affect the `timestamp` since it still thinks it runs for 10 seconds. 

### Settings
```json
{
  "SettingsVersion": 1.2,
  "SimMode": "Car",
  "ClockSpeed": 0.5
}
```

1. Set the `ClockSpeed` to `0.5` (or any value <1)
2. Run the simulation for 10 seconds
3. Read the value from `getCarState().timestamp`
4. The value will be about 10 seconds instead of 5 seconds (10 * 0.5)

No error message

### LINKED ISSUE #4241: clockspeed seems doesn't work
### What's your question?
Clockspeed does not work.  For example, I set clockSpeed to 3, but the timestamp in the actual emulator is still not the scale of the real-world timestamp.  I would like to ask if this means clockSpeed has no effect.  It's just a visual acceleration, and it doesn't really help the training outcome, right? 

#### Context details

UAR% ~XJW5$U0K_VVW7D~UX

(5Q011 WV7%$S{3B0A8$WDM

### Include details of what you already did to find answers
How do I set up clockSpeed to work？@ahmed-elsaharti  Thanks.

zimmy87:
Fixes: #4241    
Fixes: #4232

## About
The car mode currently utilizes a default clock for returning the CarState's timestamp, so the clock speed from the settings isn't reflected in the timestamp. This change adds logic for initializing the appropriate clock type from settings as well as logic for reading the timestamp from the newly initialized clock. 

## How Has This Been Tested?
The following script was used in car mode to confirm the clock was advancing at the proper clock speed. This script outputs the delta of the simulation clock after 1 real-world second elapses. If everything is working correctly, this script should output the same value as the ClockSpeed in the settings:

```
import setup_path 
import airsim

import pprint
import time

# connect to the AirSim simulator 
#client = airsim.MultirotorClient()
client = airsim.CarClient()
client.confirmConnection()
client.enableApiControl(True)

#prev = client.getMultirotorState().timestamp
prev = client.getCarState().timestamp
curr = prev
print(prev)
while True:
    #curr = client.getMultirotorState().timestamp
    curr = client.getCarState().timestamp
    delta = curr - prev
    prev = curr
    print(delta / 1000000000)
    time.sleep(1.0)

client.enableApiControl(False)
```

### DIFF
--- a/Unreal/Plugins/AirSim/Source/Vehicles/Car/CarPawnApi.cpp
+++ b/Unreal/Plugins/AirSim/Source/Vehicles/Car/CarPawnApi.cpp
@@ -35,7 +35,7 @@ msr::airlib::CarApiBase::CarState CarPawnApi::getCarState() const
         movement_->GetEngineMaxRotationSpeed(),
         last_controls_.handbrake,
         *pawn_kinematics_,
-        msr::airlib::ClockFactory::get()->nowNanos());
+        vehicle_api_->clock()->nowNanos());
     return state;
 }
 
--- a/Unreal/Plugins/AirSim/Source/Vehicles/Car/SimModeCar.cpp
+++ b/Unreal/Plugins/AirSim/Source/Vehicles/Car/SimModeCar.cpp
@@ -41,6 +41,8 @@ void ASimModeCar::continueForFrames(uint32_t frames)
 
 void ASimModeCar::setupClockSpeed()
 {
+    Super::setupClockSpeed();
+
     current_clockspeed_ = getSettings().clock_speed;
 
     //setup clock in PhysX


rhalaly:
Does it also support pause and resume? If the simulation is paused for 10 seconds, will the timestamp be the same?

jonyMarino:
@Thanks @zimmy87 !!!

zimmy87:
Hi @rhalaly, thank you for taking a look at this PR. From testing locally with `pause_continue_car.py`, it looks like the new clock does respond to pause and resume commands correctly.

Issue #3291, opened 2021-01-11
Airsim crash while training


CoolCoder54323:
## Bug report 

- AirSim Version/#commit: 1.3.1
- UE/Unity version: 4.25.4
- OS Version: windows 10

### What's the issue you encountered?
The simulation crashes after any amount of training time, could be 2k steps or 100k. It gives this error:
![image](https://user-images.githubusercontent.com/29154547/104158700-e4599d00-53bb-11eb-9611-a329f727bfe7.png)

### Settings
{
  "SeeDocsAt": "https://github.com/Microsoft/AirSim/blob/master/docs/settings_json.md",
  "SettingsVersion": 1.2,
  "SimMode": "Multirotor",
  "ClockSpeed": 1.0,
  "Vehicles": {
    "SimpleFlight": {
      "VehicleType": "SimpleFlight",
      "DefaultVehicleState": "Armed",
      "EnableCollisionPassthrogh": false,
      "EnableCollisions": true,
      "AllowAPIAlways": true,
      "RC": {
        "RemoteControlID": 0,
        "AllowAPIWhenDisconnected": false
      }
    }
  }
}

### How can the issue be reproduced?
It happens every time I train I have yet to get a full session in.

### Include full error message in text form
Unhandled exception at 0x00007FF981AA96AD (UE4Editor-AirSim-Win64-DebugGame.dll) in UE4Editor-Win64-DebugGame.exe: 0xC0000005: Access violation reading location 0x0000000000009074.

jonyMarino:
Hi @CoolCoder54323! Can you provide more details on how you are training the drone so we can reproduce the issue?

CoolCoder54323:
I am using the rl files that came with airsim I modeified the dron_env in airgym and made a custom file to run it called dronetrain
I also moved airgym into pythonclient/multirotor and that is all the changes I made.

Drone_Env: 
[drone_env.txt](https://github.com/microsoft/AirSim/files/5798413/drone_env.txt)

DroneTrain:
[DroneTrain.txt](https://github.com/microsoft/AirSim/files/5798419/DroneTrain.txt)

CoolCoder54323:
One more thing to add. It usually happens when I click into unreal engine after being tabbed out and interact with it, but not always it happened one time when I already had UE in focus and I was away from my computer. Hope this helps!

CoolCoder54323:
@jonyMarino Were you by any chance able to reproduce the error.

saihv:
[...]

PR #4321, opened 2022-01-26
update pixel format of UTextureRenderTarget2D instances

### LINKED ISSUE #4120: Breakpoint in RenderRequest.cpp while taking Scene Images
## Bug report 

- AirSim Version/#commit: v1.5.0-windows (also seen in 1.6.0)
- UE/Unity version: Unreal 4.25 (also seen in 4.26)
- autopilot version: SimpleFlight (Also seen with PX4 SITL)
- OS Version: Windows version 1809

As I brought up in Issue #3291, when taking lots of Scene images as one would do for Reinforcement Learning, Unreal Engine generates a breakpoint inside of RenderRequest.cpp and crashes the simulator. 

AirSim_Issue

### Settings
 Default settings

1. Download and Build AirSim and the default Blocks environment.
2. Change the image request in the _init_ function of drone_env.py from `self.image_request = airsim.ImageRequest(
            3, airsim.ImageType.DepthPerspective, True, False
        )` to `self.image_request = airsim.ImageRequest(
            3, airsim.ImageType.Scene, True, False
        )`
3. Inside of transform_obs, change `img2d = np.reshape(img1d, (responses[0].height, responses[0].width))` to `img2d = np.reshape(img1d, (responses[0].height, responses[0].width, 3))`
4. Inside of transform_obs, change `return im_final.reshape([84, 84, 1])` to `return im_final.reshape([84, 84, 3])`
5. Start Unreal Engine and run dqn_drone.py. Wait for a minute or two for the images to be taken and processed.
 
UE4Editor-Win64-DebugGame.exe has triggered a breakpoint.

zimmy87:
Fixes: #4120    

## About
This PR switches the pixel format inside the UTextureRenderTarget2D for various image types from EPixelFormat::PF_B8G8R8A8 to EPixelFormat::PF_FloatRGBA to enable compatibility with FD3D11DynamicRHI::RHIReadSurfaceFloatData. There is a check inside FD3D11DynamicRHI::RHIReadSurfaceFloatData that ensures the pixel format of the inputted FRHITexture is EPixelFormat::PF_FloatRGBA.

## How Has This Been Tested?
The scenario described in the repro steps of #4120 was confirmed to be working with this fix. Basic movement of a drone with a controller was tested with subwindows for each ImageType enabled to confirm each ImageType renders correctly. hello_drone.py was also run to confirm scene captures can still be taken and look correct.

### DIFF
--- a/Unreal/Plugins/AirSim/Source/PIPCamera.cpp
+++ b/Unreal/Plugins/AirSim/Source/PIPCamera.cpp
@@ -36,16 +36,16 @@ APIPCamera::APIPCamera(const FObjectInitializer& ObjectInitializer)
 
     PrimaryActorTick.bCanEverTick = true;
 
-    image_type_to_pixel_format_map_.Add(0, EPixelFormat::PF_B8G8R8A8);
+    image_type_to_pixel_format_map_.Add(0, EPixelFormat::PF_FloatRGBA);
     image_type_to_pixel_format_map_.Add(1, EPixelFormat::PF_DepthStencil); // not used. init_auto_format is called in setupCameraFromSettings()
     image_type_to_pixel_format_map_.Add(2, EPixelFormat::PF_DepthStencil); // not used for same reason as above
     image_type_to_pixel_format_map_.Add(3, EPixelFormat::PF_DepthStencil); // not used for same reason as above
     image_type_to_pixel_format_map_.Add(4, EPixelFormat::PF_DepthStencil); // not used for same reason as above
-    image_type_to_pixel_format_map_.Add(5, EPixelFormat::PF_B8G8R8A8);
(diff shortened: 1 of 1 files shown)


rajat2004:
Are we sure this is correct, seems strange that this only results in crashes sometimes. 
Could you also put a link (or filename) to the relevant UE4 source code if possible, thanks!

jonyMarino:
Thanks @zimmy87!

zimmy87:
Thanks for taking a look at this PR @rajat2004, I think you might be confusing #4120 with a related issue #3291. When I tried the repro of #4120, I was getting a 100% repro, so this isn't an intermittent or timing issue.

The relevant UE4 source code is [here](https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Windows/D3D11RHI/Private/D3D11RenderTarget.cpp#L843)

PR #4332, opened 2022-02-01
Pin MacOS version to 10.15 in Actions

rajat2004:
Fixes: #    

## About

Recently macos-latest became 11 from the earlier 10.15. Pin the version to be used to 10.15


jonyMarino:
Thanks a lot @rajat2004 !

PR #4349, opened 2022-02-11
Fix ros reset service crash

alonfaraj:
Fixes: #    

## About

Calling to ros service reset caused airsim node to crash.
This PR fix it for both ros1/2

## How Has This Been Tested?

Run blocks, run airsim node and call ros reset from the terminal.


shuaikangma:
I test reset service manually use rqt in ros, it can't take off after reset service being called. 
AIrSim version: #4410 
Unreal version: test v1.7.0-linux and v1.6.0-linux
OS version: ubuntu18.04 with ros-melodic
Use default settiongs :AirSim/ros/settings/front_stereo_and_center_mono.json in #4410

PR #2075, opened 2019-07-14
ArduCopter Support

rajat2004:
Adds support for ArduCopter SITL (https://github.com/ArduPilot/ardupilot)
Also adds a simple single-threaded UDP Socket implementation which is used to implement Lock-Step Scheduling

Please note that this doesn't have support for the Airsim APIs right now

Corresponding PRs in Ardupilot (Have been merged!)- https://github.com/ArduPilot/ardupilot/pull/11367 (Basic interface), https://github.com/ArduPilot/ardupilot/pull/11835 (Lidar, RC, Multi-Vehicle Sim)
Wiki Page in Ardupilot Developer Documentation - http://ardupilot.org/dev/docs/sitl-with-airsim.html

Tested with Unreal Engine for Ubuntu 16.04(everything running natively), 18.04(on WSL) with AirSim in Windows proper.

Review is very much appreciated, especially on the Windows side since this is my first time writing code for Windows

PR #4357, opened 2022-02-16
Betaflight firmware support

nikhil-sethi:
Fixes: 
NA

## About

This PR adds support for betaflight as a new firmware for drone flight. Reference issue: #4351 
The implementation is similar to that of arducopter with some basic changes that I wanted to discuss.

## How Has This Been Tested?

- Tested with betaflight 4.3.0 at commit 3267f041
- Currently only tested on Linux (Ubuntu 20.04). I understand that windows is recommended but my dual boot does not have AirSim + Unreal setup as of yet. Let me know if I should set it up or if someone else can test it! Also, I don't have a mac machine.
- RC input tested with a very basic [custom program](https://github.com/nikhil-sethi/joystick) that outputs normalised joystick values to a UDP socket that betaflight listens to.

Known issues and doubts:
1. The stick responses are good enough for all axes after tuning PID values but the throttle seems sluggish for flying or testing. (takeoff at ~1700; almost full when doing maneuvers). Is this a basic inertia/mass issue? Shouldn't it be decoupled from raw PWM values. I haven't investigated this yet.
2. The arducopter implementation uses json to get the fdm packet here and there. #2075 says this is for easy addition and removal of sensors. However, for better compatibility, I found it easier and minimal to implement an fdm packet struct equivalent to the SITL [target](https://github.com/betaflight/betaflight/blob/7b4415f062656218427ee160b776ac5488eca5f1/src/main/target/SITL/target.h#L238) in betaflight. Are there some performance issues with this implementation that would be better with json? Please let me know if I am missing something.

- Testing video: https://youtu.be/sVEfow3y6rU

## TODO
- [ ] Fix/understand throttle issue
- [x] Fix/Understand fdm data structure
- [x] Format and cleanup according to repo style

### DIFF
--- a/AirLib/include/vehicles/multirotor/firmwares/betaflight/BetaflightApi.hpp
+++ b/AirLib/include/vehicles/multirotor/firmwares/betaflight/BetaflightApi.hpp
@@ -2,7 +2,6 @@
 #define msr_airlib_BetaflightDroneController_hpp
 
 #include "vehicles/multirotor/api/MultirotorApiBase.hpp"
-#include "sensors/SensorCollection.hpp"
 #include "physics/Environment.hpp"
 #include "physics/Kinematics.hpp"
 #include "vehicles/multirotor/MultiRotorParams.hpp"
@@ -12,31 +11,29 @@
 
 // Sensors
 #include "sensors/imu/ImuBase.hpp"
-#include "sensors/gps/GpsBase.hpp"
-#include "sensors/magnetometer/MagnetometerBase.hpp"
-#include "sensors/barometer/BarometerBase.hpp"
-#include "sensors/distance/DistanceSimple.hpp"
-#include "sensors/lidar/LidarSimple.hpp"
+
 #include "UdpSocket.hpp"
 
-#include <sstream>
 
+namespace msr
+{ 
+namespace airlib 
+{
 
-namespace msr{ namespace airlib {
-
-class BetaflightApi: public MultirotorApiBase{
+    class BetaflightApi: public MultirotorApiBase
+    {
 
     public:
-    // Contstructor
-    BetaflightApi(const MultiRotorParams* vehicle_params,  const AirSimSettings::MavLinkConnectionInfo& connection_info): 
-        connection_info_(connection_info), vehicle_params_(vehicle_params)
+        BetaflightApi(const MultiRotorParams* vehicle_params,  const AirSimSettings::MavLinkConnectionInfo& connection_info)
+            : connection_info_(connection_info), vehicle_params_(vehicle_params)
         {
             connect();
         }
 
-    ~BetaflightApi(){
-         closeConnections();
-    }
+        ~BetaflightApi()
+        {
+            closeConnections();
+        }
 
     public:
         virtual void resetImplementation() override
@@ -47,7 +44,8 @@ class BetaflightApi: public MultirotorApiBase{
         }
 
         // Update sensor data & send to betaflight
-        virtual void update() override {
+        virtual void update() override 
+        {
             MultirotorApiBase::update();
 
             sendState();
@@ -323,12 +321,12 @@ class BetaflightApi: public MultirotorApiBase{
             }
         }
 
-        void sendState(){    // send fdmPacket to betaflight i.e udp:9002
-            
-            fdmPacket pkt;
+        void sendState() // send fdmPacket to betaflight i.e udp:9002
+        {   
+            FdmPacket pkt;
 
             // Time
-            pkt.timestamp = ClockFactory::get()->nowNanos() / 1000;
+            pkt.timestamp = ClockFactory::get()->nowNanos() / 1E9;
             
             // IMU
             const auto& imu_output = getImuData("");
@@ -342,7 +340,7 @@ class BetaflightApi: public MultirotorApiBase{
                 pkt.imu_linear_acceleration_xyz[1] = imu_output.linear_acceleration[1];
                 pkt.imu_linear_acceleration_xyz[2] = imu_output.linear_acceleration[2];
 
-                // Orientation Quaternion. In case USE_IMU_CALC is defined on betaflight side
+                // Orientation Quaternion. In case USE_IMU_CALC is not defined on betaflight side
                 pkt.imu_orientation_quat[0] = imu_output.orientation.w();
                 pkt.imu_orientation_quat[1] = imu_output.orientation.x();
                 pkt.imu_orientation_quat[2] = imu_output.orientation.y();
@@ -357,8 +355,9 @@ class BetaflightApi: public MultirotorApiBase{
             }
         }
 
-        void recvControl(){
-            servoPacket pkt;
+        void recvControl()
+        {
+            ServoPacket pkt;
             int recv_ret = udp_socket_->recv(&pkt, sizeof(pkt), 100);
             while (recv_ret != sizeof(pkt)) {
                 if (recv_ret <= 0) {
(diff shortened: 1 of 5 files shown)


AirLib/include/vehicles/multirotor/firmwares/betaflight/BetaflightApi.hpp
```diff
@@ -0,0 +1,413 @@
...
+            if (ip_ == "") {
+                throw std::invalid_argument("UdpIp setting is invalid.");
+            }
+
+            if (port_ == 0) {
+                throw std::invalid_argument("UdpPort setting has an invalid value.");
+            }
+
+            Utils::log(Utils::stringf("Using UDP port %d, local IP %s, remote IP %s for sending sensor data", port_, connection_info_.local_host_ip.c_str(), ip_.c_str()), Utils::kLogLevelInfo);
+            Utils::log(Utils::stringf("Using UDP port %d for receiving rotor power", connection_info_.control_port_local, connection_info_.local_host_ip.c_str(), ip_.c_str()), Utils::kLogLevelInfo);
```
rajat2004, line 310: The log string doesn't have the required `%s` here to print the vars

AirLib/include/vehicles/multirotor/firmwares/betaflight/BetaflightApi.hpp
```diff
@@ -0,0 +1,413 @@
...
+                rotor_controls_[i] = (rotor_controls_[i] - 1000.0f) / 1000.0f;
+            }
+        }
+
+        void sendState(){    // send fdmPacket to betaflight i.e udp:9002
+            
+            fdmPacket pkt;
+
+            // Time
+            pkt.timestamp = ClockFactory::get()->nowNanos() / 1000;
```
rajat2004, line 331: This seems wrong. First, the fdm definition mentions seconds, but this will give value in microseconds.
Also, the type is `double`, but the RHS `nowNanos()` returns uint64, so dividing by 10^3 will lose ns info (probably not needed), and then convert to double. Ensure the required precision is calculated correctly.

AirLib/include/vehicles/multirotor/firmwares/betaflight/BetaflightApi.hpp
```diff
@@ -0,0 +1,413 @@
...
+                pkt.imu_orientation_quat[2] = imu_output.orientation.y();
+                pkt.imu_orientation_quat[3] = imu_output.orientation.z();
+            // GPS
+
+                // Position
+
+                // Velocity
+            if (udp_socket_ != nullptr) {
+                udp_socket_->sendto(&pkt, sizeof(pkt), ip_, port_);
+            }
```
rajat2004, line 357: This check could be done in the start itself, if `nullptr` then just return early

AirLib/include/vehicles/multirotor/firmwares/betaflight/BetaflightApi.hpp
```diff
@@ -0,0 +1,413 @@
...
+                }
+
+                recv_ret = udp_socket_->recv(&pkt, sizeof(pkt), 100);
+            }
+            for (auto i = 0; i < kBetaflightRotorControlCount; ++i) {
+                rotor_controls_[i] = pkt.pwm[i];
+            }
+
+            normalizeRotorControls();
+            Utils::log(Utils::stringf("pwm received: [ %f, %f, %f, %f ]", rotor_controls_[0], rotor_controls_[1], rotor_controls_[2], rotor_controls_[3]), Utils::kLogLevelInfo);
```
rajat2004, line 378: Should be removed once things are final, otherwise too much logging

AirLib/include/vehicles/multirotor/firmwares/betaflight/BetaflightApi.hpp
```diff
@@ -0,0 +1,413 @@
...
+private:
+    
+    struct fdmPacket{  // equivalent of fdm_packet in betaflight SITL
+        double timestamp;                   // in seconds
+        double imu_angular_velocity_rpy[3]; // rad/s -> range: +/- 8192; +/- 2000 deg/se
+        double imu_linear_acceleration_xyz[3];    // m/s/s NED, body frame -> sim 1G = 9.80665, FC 1G = 256
+        double imu_orientation_quat[4];     // w, x, y, z
+        double velocity_xyz[3];             // m/s, earth frame
+        double position_xyz[3];             // meters, NED from origin
+    } ;
```
rajat2004, line 390: Do confirm whether you want `__attribute__((__packed__))` to avoid padding issues when reading raw bytes directly into struct

AirLib/include/vehicles/multirotor/firmwares/betaflight/BetaflightApi.hpp
```diff
@@ -0,0 +1,413 @@
...
+    
+    static const int kBetaflightRotorControlCount = 4;
+    
+    struct servoPacket {  // equivalent of servo_packet in betaflight SITL
+        u_int16_t pwm[kBetaflightRotorControlCount]; //
```
rajat2004, line 395: What is this `u_int16_t`? Haven't seen this before...

rajat2004 (commented):
Just a brief look, would be great if you could also do the clang-formatting, setting up in IDE should be simple. The maintainers will need to grant access to run the Github workflows which would do some basic compilation and formatting checks

rajat2004:
Re JSON vs FDM packet, the benefit with JSON (or any kind of serialized format) is that the sender and receiver can add more types of data, extract info that is required and ignore the rest (e.g. if something is not supported, backwards compatibility). For e.g. with Lidar & RC control support, it would have worked even if either side didn't support it, it would just be ignored. For the fixed struct, variable data like Lidar won't work very nicely since the receiver would need to know exactly where the data starts and ends. New fields can only be added at the end, any change anywhere else in the struct would cause incorrect data being read if older struct is being used on other side. Again one of the reasons with ArudPilot is also moving more towards using JSON as a generic and flexible way to implement interfaces for multiple simulators

AirLib/include/vehicles/multirotor/firmwares/betaflight/BetaflightApi.hpp
nikhil-sethi: Yes. Will fix this.

AirLib/include/vehicles/multirotor/firmwares/betaflight/BetaflightApi.hpp
nikhil-sethi: Yes definitely. Will fix

AirLib/include/vehicles/multirotor/firmwares/betaflight/BetaflightApi.hpp
nikhil-sethi: Agreed. Will add the check for sensors also at the top

AirLib/include/vehicles/multirotor/firmwares/betaflight/BetaflightApi.hpp
nikhil-sethi: (today I learned) Apparently this is (one of) the typedefs in sys/types.h used for integers before C99 came up with the standards for uintXX_t. I must've added an underscore by mistake and let intellisense worry about the rest. Cursing myself anyways..

AirLib/include/vehicles/multirotor/firmwares/betaflight/BetaflightApi.hpp
nikhil-sethi: I'm not familiar with this attribute but I understand what it's doing. I'll read up + experiment with it before confirming. Will get back to you for sure.

AirLib/include/vehicles/multirotor/firmwares/betaflight/BetaflightApi.hpp
nikhil-sethi: Will correct it to the accepted standard.

nikhil-sethi:
> Re JSON vs FDM packet, the benefit with JSON (or any kind of serialized format) is that the sender and receiver can add more types of data, extract info that is required and ignore the rest (e.g. if something is not supported, backwards compatibility). For e.g. with Lidar & RC control support, it would have worked even if either side didn't support it, it would just be ignored. For the fixed struct, variable data like Lidar won't work very nicely since the receiver would need to know exactly where the data starts and ends. New fields can only be added at the end, any change anywhere else in the struct would cause incorrect data being read if older struct is being used on other side.
> 
> FDM struct would be faster in performance than serializing to string and later parsing it, it'll most likely not be the bottleneck for the simulation, there are other much more obvious issues like AirSim internal physics and UE4 rendering, firmware side handling, and then network communication. Another thing that was very useful with the JSON interface was the ease of debugging, the packet can be easily read manually in Wireshark, to spot any obvious errors.
> 
> ArduPilot is also moving towards using JSON for interfacing with multiple simulators, e.g. [this Discuss post](https://discuss.ardupilot.org/t/sitl-protocol-for-custom-simulators/60365/5)

Thanks for explaining this in detail. It makes sense now why JSON was used. I'll progress with FDM for current testing purposes and in the meantime discuss at the betaflight PR if I can add the JSON unpacking routine in their target.

AirLib/include/vehicles/multirotor/firmwares/betaflight/BetaflightApi.hpp
nikhil-sethi: 1. I agree that it's wrong and I knew it would get pointed out. But for some reason, the right implementation with division by 1E9 didn't seem to be working with betaflight. 
Upon further debugging I recently found this was a timeout issue in [this](https://github.com/betaflight/betaflight/blob/7b4415f062656218427ee160b776ac5488eca5f1/src/main/target/SITL/target.c#L94) line and the timeout should be increased to 600ms to calculate the correct simRate. There should be a general method to fix this on the betaflight side. Will work on it in the future.
Also betaflight needs relative time from start of simulation which I have changed.

2. I understand what you're saying. I could cast the time to a double before division like so:
```
pkt.timestamp = static_cast<double>(ClockFactory::get()->nowNanos()) / 1E9;
```
But this gives me the same result without casting and works the same with betaflight. This timestamp is used to find the simulation delta time which is exactly 0.003 sec. Am I understanding correctly that the lost ns info doesn't matter (when dividing by 1e9) because of this reason? I'll still add the casting to make it clearer though.

zhouqingf:
Hello, in the process of replicating airsim and bateflight, I couldn't implement communication between airsim and silt. I tried to configure setting.json, but once tested, UE waited for a communication connection and the program wouldn't work. I'd like to know how your setting.jsons are configured.

PR #4377, opened 2022-02-26
Pin Actions to Windows 2019

rajat2004:
Fixes: #    

## About

`windows-latest` recently changed to Windows 2022, pin the version to 2019 for CI. Separate work needs to be done to support VS 2022


zimmy87:
Thanks for catching this and fixing it @rajat2004!

PR #4385, opened 2022-03-02
fix scaling of pause period in car mode

### LINKED ISSUE #1719: continueForTime API is not accurate
Hello, It is after a long time of post.
I tried using AirSim for about half a year and found several problems. So I will report on the problem.
I am using AirSim for my research and I use AirSim and MATLAB by integrating it by calling Pythonclient from MATLAB.

I found that AirSim 's continuefortime API is incorrect.
This problem occurs in car mode.
The biggest problem is that updates are done for longer than the specified time.
We will update 0.01 seconds with simcontinuosfortime. Then, the difference between the moving distance obtained from the product of the speed and the step time and the difference between the previous position does not match. The verification script will be posted below.

```
import setup_path
import airsim

import time
import math


DELTA_T=0.01
#timerate=1.7

histTime=[0]
histVelocity=[0]
histPositionX=[0]
histPositionY=[0]
histPositionZ=[0]
calculatedStepDistFromVel=[0]
calculatedStepDistFromPosition=[0]

# connect to the AirSim simulator
client = airsim.CarClient()
client.confirmConnection()
client.enableApiControl(True)

car_controls = airsim.CarControls()


client.simPause(True)


for i in range(1, 1500):
    car_controls.throttle = 1.0
    car_controls.steering = 0
    client.setCarControls(car_controls)

    client.simContinueForTime(DELTA_T)

    car_state = client.getCarState()
    #get position and velocity
    histTime.append(i*DELTA_T)
    #histTime.append(i*DELTA_T*timerate)
    histVelocity.append(car_state.speed)
    histPositionX.append(car_state.kinematics_estimated.position.x_val)
    histPositionY.append(car_state.kinematics_estimated.position.y_val)
    histPositionZ.append(car_state.kinematics_estimated.position.z_val)

    #calculate step distance form velocity and position
    calculatedStepDistFromVel.append(histVelocity[i]*DELTA_T)
    #calculatedStepDistFromVel.append(histVelocity[i]*DELTA_T/timerate)
    calculatedStepDistFromPosition.append(math.sqrt( \
         math.pow((histPositionX[i]-histPositionX[i-1]),2) \
        +math.pow((histPositionY[i]-histPositionY[i-1]),2) \
        +math.pow((histPositionZ[i]-histPositionZ[i-1]),2)))


print(sum(calculatedStepDistFromVel))
print(sum(calculatedStepDistFromPosition))
print(sum(calculatedStepDistFromVel)/sum(calculatedStepDistFromPosition))
```
When executed, the output result is as follows.
```
Connected!

156.9931730871114
93.58373037297422
1.6775690866502262
```
The settings used in the program are as follows.
```
{
	"SeeDocsAt": "https://github.com/Microsoft/AirSim/blob/master/docs/settings.md", "SettingsVersion": 1.2, "SimMode": "Car",
	"ClockSpeed": 1.0,
	"ViewMode": "",
	"CameraDefaults": {
		"CaptureSettings":
		[
			{
				"ImageType": 0, "Width": 1024, "Height": 576, "FOV_Degrees": 90,
				"AutoExposureSpeed": 0, "AutoExposureBias": 0, "AutoExposureMaxBrightness": 2.00, "AutoExposureMinBrightness": 0.03,
				"MotionBlurAmount": 0, "TargetGamma": 1.5, "ProjectionMode": "", "OrthoWidth": 5.12
			},
			{
				"ImageType": 1, "Width": 512, "Height": 128, "FOV_Degrees": 100,
				"AutoExposureSpeed": 0, "AutoExposureBias": 0, "AutoExposureMaxBrightness": 2.00, "AutoExposureMinBrightness": 0.03,
				"MotionBlurAmount": 0, "TargetGamma": 1.5, "ProjectionMode": "", "OrthoWidth": 5.12
			}
		]
	},
	"SubWindows":
	[
		{"WindowID": 0, "ImageType": 0, "CameraName": "DCameraLeft", "Visible": false},
		{"WindowID": 1, "ImageType": 0, "CameraName": "DCameraFront", "Visible": false},
		{"WindowID": 2, "ImageType": 0, "CameraName": "ExtCamera", "Visible": false}
	],
	"Vehicles": {

"Car1": { "VehicleType": "PhysXCar", "X":1750.01, "Y":456.9992, "Z":0, "Pitch":0, "Roll":0, "Yaw":-4.5375
}

	}
}
```

This problem also occurs with a combination of simpause / time.sleep(python) or pause(MATLAB) / simpause.
Another problem related to the pause API is that there is a problem that RPM slowly fluctuates during pause. This problem has been reported in #1584.
Under such circumstances, it can not be used for strict physical simulation.
In addition to this, I found a problem that speed and acceleration information takes a large value. This has been reported in #1501.
We also found a problem that images at the same time do not match. This has been reported in #1494.

We will also report a different issue from the pause API.
When I request images to AirSim, I found a problem that processing becomes extremely slow. When processing time is measured while acquiring images, nearly 80% of processing time is used for image acquisition. It is considered that improvement is necessary.
This problem has already been reported by #1669 by other people, but I will report it again.

### LINKED ISSUE #4350: `timestamp` doesn't consider pause time
## Bug report 

- AirSim Version/#commit: 8057725712c0cd46979135396381784075ffc0f3
- UE/Unity version: 4.27
- autopilot version:
- OS Version: Windows 11

It seems that #4305 didn't fully solve the issue (@zimmy87 can you assist please?). Trying this piece of code:

```
import airsim

import time

client = airsim.CarClient()
client.confirmConnection()
client.enableApiControl(True)

prev = client.getCarState().timestamp

client.simPause(True)  # Pause the simulation
while True:
    curr = client.getCarState().timestamp
    delta = curr - prev
    prev = curr
    print(delta * 1e-9)

    time.sleep(2.0)  # Sleep for 2 seconds - should be ignored by the simulator
    client.simContinueForTime(1.0)  # Continue the simulation for 1 second
    while not client.simIsPause():
        time.sleep(0.1)

```

This code sleeps 2 seconds while the simulation is paused and continues the simulation for 1 second. So the expected output should be 1 second since the simulation was played for 1 second every time.

But got the output 
```
Connected!

0.0
5.6500268
3.0188154000000003
3.0312138
3.0336949
3.0241089000000003
3.0263058000000003
3.0252388000000003
3.0206698000000003
3.0262011
3.020782
3.0155367
3.0184487
...
```

If I ignore the first value, I don't fully understand why I get `5.65` (and other high numbers) in the first reading. But after that, I get 3 seconds constantly instead of 1 second as expected.

### Settings

```
{
  "SettingsVersion": 1.2,
  "SimMode": "Car"
}
```

1. Build AirSim with commit 8057725712c0cd46979135396381784075ffc0f3
2. Run the above code

No error, just unexpected result

zimmy87:
Fixes: #4350
Fixes: #1719    

## About
The PR allows for the clock speed to be correctly accounted for during a call to client.simContinueForTime() by scaling the pause period accordingly (faster clockspeed should result in longer period to allow for same wall time to elapse during API call).

## How Has This Been Tested?
The test script in #4350 was used along with a modified version for multirotor mode to ensure feature parity.

### DIFF
--- a/AirLib/include/common/ClockBase.hpp
+++ b/AirLib/include/common/ClockBase.hpp
@@ -48,6 +48,11 @@ namespace airlib
             return elapsed;
         }
 
+        virtual TTimePoint stepBy(TTimeDelta amount)
+        {
+            return step();
+        }
+
         virtual TTimePoint step()
         {
             //by default step doesn't do anything
--- a/AirLib/include/common/ClockBase.hpp
+++ b/AirLib/include/common/ClockBase.hpp
@@ -50,6 +50,7 @@ namespace airlib
 
         virtual TTimePoint stepBy(TTimeDelta amount)
         {
+            unused(amount);
             return step();
         }
 
--- a/AirLib/include/common/SteppableClock.hpp
+++ b/AirLib/include/common/SteppableClock.hpp
@@ -30,7 +30,7 @@ namespace airlib
 
         virtual ~SteppableClock() {}
 
-        TTimePoint stepBy(TTimeDelta amount)
+        virtual TTimePoint stepBy(TTimeDelta amount) override
         {
             current_ = addTo(current_, amount);
             return current_;
--- a/Unreal/Plugins/AirSim/Source/Vehicles/Car/SimModeCar.cpp
+++ b/Unreal/Plugins/AirSim/Source/Vehicles/Car/SimModeCar.cpp
@@ -28,7 +28,7 @@ void ASimModeCar::initializePauseState()
 void ASimModeCar::continueForTime(double seconds)
 {
     pause_period_start_ = ClockFactory::get()->nowNanos();
-    pause_period_ = seconds;
+    pause_period_ = seconds * current_clockspeed_;
     pause(false);
 }
 
@@ -53,6 +53,9 @@ void ASimModeCar::setupClockSpeed()
 void ASimModeCar::Tick(float DeltaSeconds)
 {
     Super::Tick(DeltaSeconds);
+
+    if (!isPaused())
+        ClockFactory::get()->stepBy(DeltaSeconds);
 
     if (pause_period_start_ > 0) {
         if (ClockFactory::get()->elapsedSince(pause_period_start_) >= pause_period_) {


jonyMarino:
Thanks, @zimmy87!

jonyMarino:
Thanks, @zimmy87!

PR #4393, opened 2022-03-05
Improve Codes & Fixing Narrowing Conversion.

BillyFrcs:
## About
- Use auto when initializing with a template cast to avoid duplicate type name.
- Fixed Narrowing conversion warning variable with static cast.
- Adding const & constexpr in some variable & parameter so they can't be modified.
- Comment on unused include library to avoid large binary file.

## How Has This Been Tested?
Everything looks really well as usual.

### DIFF
--- a/Examples/main.cpp
+++ b/Examples/main.cpp
@@ -149,7 +149,7 @@ void runDepthNavSGM()
     depthNav.initialize(client, request);
 
     SGMOptions params;
-    CStateStereo* p_state;
+    CStateStereo p_state;
 
     if (params.maxImageDimensionWidth != static_cast<int>(depthNav.params_.depth_width))
         printf("WARNING: Width Mismatch between SGM and DepthNav. Overwriting parameters.\n");
@@ -157,13 +157,9 @@ void runDepthNavSGM()
 
     params.Print();
 
-    p_state = new CStateStereo;
-    p_state->Initialize(params, static_cast<int>(depthNav.params_.depth_height), static_cast<int>(depthNav.params_.depth_width));
+    p_state.Initialize(params, static_cast<int>(depthNav.params_.depth_height), static_cast<int>(depthNav.params_.depth_width));
 
-    depthNav.gotoGoalSGM(goalPose, client, request, p_state);
-
-    //Cleanup
-    delete p_state;
+    depthNav.gotoGoalSGM(goalPose, client, request, &p_state);
 }
 
 int main(const int argc, const char* argv[])
--- a/Examples/main.cpp
+++ b/Examples/main.cpp
@@ -149,13 +149,15 @@ void runDepthNavSGM()
     depthNav.initialize(client, request);
 
     SGMOptions params;
-    auto* p_state = new CStateStereo();
+    CStateStereo* p_state;
 
     if (params.maxImageDimensionWidth != static_cast<int>(depthNav.params_.depth_width))
         printf("WARNING: Width Mismatch between SGM and DepthNav. Overwriting parameters.\n");
     params.maxImageDimensionWidth = static_cast<int>(depthNav.params_.depth_width);
 
     params.Print();
+
+    p_state = new CStateStereo;
     p_state->Initialize(params, static_cast<int>(depthNav.params_.depth_height), static_cast<int>(depthNav.params_.depth_width));
 
     depthNav.gotoGoalSGM(goalPose, client, request, p_state);
--- a/Examples/main.cpp
+++ b/Examples/main.cpp
@@ -5,10 +5,8 @@
 #include "GaussianMarkovTest.hpp"
 #include "DepthNav/DepthNavCost.hpp"
 #include "DepthNav/DepthNavThreshold.hpp"
-// #include "DepthNav/DepthNavOptAStar.hpp"
 #include <iostream>
 #include <string>
-// #include <sys/stat.h>
 #ifndef _USE_MATH_DEFINES
 #define _USE_MATH_DEFINES
 #endif
@@ -97,7 +95,7 @@ void runDepthNavGT()
     typedef ImageCaptureBase::ImageRequest ImageRequest;
     typedef ImageCaptureBase::ImageType ImageType;
 
-    const std::vector<ImageRequest> request = {
+    std::vector<ImageRequest> request = {
         ImageRequest("front_left", ImageType::DepthPlanar, true) /*,
         ImageRequest("front_left", ImageType::Scene),
         ImageRequest("front_left", ImageType::DisparityNormalized, true) */
--- a/Examples/main.cpp
+++ b/Examples/main.cpp
@@ -5,14 +5,14 @@
 #include "GaussianMarkovTest.hpp"
 #include "DepthNav/DepthNavCost.hpp"
 #include "DepthNav/DepthNavThreshold.hpp"
-#include "DepthNav/DepthNavOptAStar.hpp"
+// #include "DepthNav/DepthNavOptAStar.hpp"
 #include <iostream>
 #include <string>
-#include <sys/stat.h>
+// #include <sys/stat.h>
 #ifndef _USE_MATH_DEFINES
 #define _USE_MATH_DEFINES
 #endif
-#include <math.h>
+#include <cmath>
 
 int runStandAloneSensors(int argc, const char* argv[])
 {
@@ -40,8 +40,8 @@ int runStandAloneSensors(int argc, const char* argv[])
     //marymoore park
     //GeoPoint testLocation(47.662804385, -122.1167039875, 9.93f);
 
-    GeoPoint testLocation(47.7631699747, -122.0685655406, 9.93f); // woodinville
-    float yawOffset = 0; // static_cast<float>(91.27622  * M_PI / 180.0); // I was aligned with the road...
+    const GeoPoint testLocation(47.7631699747, -122.0685655406, 9.93f); // woodinville
+    constexpr float yawOffset = 0; // static_cast<float>(91.27622  * M_PI / 180.0); // I was aligned with the road...
 
     std::ofstream out_file(argv[1]);
     StandALoneSensors::generateImuStaticData(out_file, period, total_duration);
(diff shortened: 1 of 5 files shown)


HelloCar/main.cpp
```diff
@@ -12,7 +12,7 @@ STRICT_MODE_ON
 #include "vehicles/car/api/CarRpcLibClient.hpp"
 #include "common/common_utils/FileSystem.hpp"
 #include <iostream>
-#include <chrono>
```
rajat2004, line 15: Better to remove it

Examples/main.cpp
```diff
@@ -96,7 +97,7 @@ void runDepthNavGT()
     typedef ImageCaptureBase::ImageRequest ImageRequest;
     typedef ImageCaptureBase::ImageType ImageType;
 
-    std::vector<ImageRequest> request = {
```
rajat2004, line 99: Since this is being changed, let's remove the assignment as well, and directly use list initializer, like `const std::vector<ImageRequest> request {...}`
Similarly for other places

HelloDrone/main.cpp
```diff
@@ -33,7 +33,9 @@ int main()
         const vector<ImageResponse>& response = client.simGetImages(request);
         std::cout << "# of images received: " << response.size() << std::endl;
 
-        if (response.size() > 0) {
+        constexpr uint zero = 0;
+
+        if (response.size() > zero) {
```
rajat2004, line 38: How does this help?

HelloSpawnedDrones/HelloSpawnedDrones.cpp
```diff
@@ -13,42 +13,43 @@ STRICT_MODE_ON
...
+void runSingleClient(const uint16_t port, const int ordinal)
 {
     using namespace msr::airlib;
-    const char host[] = "localhost";
+
+    constexpr char host[] = "localhost";
     float timeout_s = 60;
 
     try {
         MultirotorRpcLibClient* client = new MultirotorRpcLibClient(host, port, timeout_s);
```
rajat2004, line 27: This can be changed to `unique_ptr` as well instead of raw pointer, if there's an exception thrown before the delete then memory will leak

HelloDrone/main.cpp
```diff
@@ -104,14 +106,14 @@ int main()
 
         std::cout << "Press Enter to fly in a 10m box pattern at 3 m/s velocity" << std::endl;
         std::cin.get();
-        // moveByVelocityZ is an offboard operation, so we need to set offboard mode.
+        // moveByVelocityZ is an off board operation, so we need to set offboard mode.
```
zimmy87, line 109: Any reason for adding the space in "off board"? Seems like most other references I see to "offboard" just write it as one word. Happens other places too.

Examples/main.cpp
```diff
@@ -5,14 +5,14 @@
 #include "GaussianMarkovTest.hpp"
 #include "DepthNav/DepthNavCost.hpp"
 #include "DepthNav/DepthNavThreshold.hpp"
-#include "DepthNav/DepthNavOptAStar.hpp"
+// #include "DepthNav/DepthNavOptAStar.hpp"
```
zimmy87, line 8: It's better to just delete unused includes rather than commenting them out. Superfluous code can always be found in source history if we need to look it up. This happens in other files as well.

zimmy87 (changes requested):
Thank you for submitting this PR @BillyFrcs! Just a few extra comments. I also agree with @rajat2004's feedback. I'd like to see his comments addressed as well before merging this.

HelloCar/main.cpp
BillyFrcs: uh alright.

HelloDrone/main.cpp
BillyFrcs: Actually, it will remove a warning message while do comparison between size of a vector with 0, so i created a variable to handle the 0 value.

HelloSpawnedDrones/HelloSpawnedDrones.cpp
BillyFrcs: um alright, so do i need to create another PR?

HelloDrone/main.cpp
BillyFrcs: ah my bad, i forgot if i adding a space between the word. i think offboard is the true word.

Examples/main.cpp
BillyFrcs: Alright thanks, i didn't know that.

HelloSpawnedDrones/HelloSpawnedDrones.cpp
zimmy87: no, feel free to just commit to this PR's branch with the change

BillyFrcs: Alright understood

Examples/main.cpp
zimmy87: This looks like this still needs to be updated in the latest revision. I see instances where the list initializer can be directly used on ln 98 & 128 of this file (as in HelloCar/main.cpp: ln 36 & HelloDrone/main.cpp: ln 32). For an example of what @rajat2004 is referring to, see the example initialization of `std::vector<std::pair<std::string, int>>` [here](http://gavinchou.github.io/experience/summary/syntax/initializer_list/#:~:text=std%3A%3Avector%20has%20a%20constructor%20of%20std%3A%3Ainitlizaer_list%2C%20possible%20initialization,thing%20to%20say%2C%20e.g.%2C%20there%20is%20a%20declaration). Basically, you can remove the `=` to simplify these lines.

HelloDrone/main.cpp
zimmy87: instead of creating a new variable, can we replace `zero` with `static_cast<uint>(0)`? The extra line needed for the variable declaration seems superfluous for what is ultimately a literal expression.

HelloSpawnedDrones/HelloSpawnedDrones.cpp
zimmy87: I'm confused by this part in latest revision. Looks like `MultirotorRpcLibClient* client` still needs to be updated to a [unique_ptr](https://en.cppreference.com/w/cpp/memory/unique_ptr), and `constexpr char host[]` got reverted back to `const char host[]` on ln 23. Any particular reason for reverting ln 23?

zimmy87:
Hey @BillyFrcs, thanks for responding to our feedback, I've gone through all the items and resolved the ones that look completed and updated the others that I think still require attention with comments specifying what remains. Once the remaining items are addressed, we can get this merged.

Examples/main.cpp
```diff
@@ -150,23 +149,22 @@ void runDepthNavSGM()
     depthNav.initialize(client, request);
 
     SGMOptions params;
-    CStateStereo* p_state;
+    auto* p_state = new CStateStereo();
```
rajat2004, line 152: This also shouldn't be a raw pointer. It can be changed to a normal local var,

HelloDrone/main.cpp
rajat2004: Rather than a zero var, let's just change it to `if (!response.empty())`, cleaner as well. Wdyt?

HelloCar/main.cpp
```diff
@@ -34,11 +33,13 @@ int main()
 
         std::cout << "Press Enter to get FPV image" << std::endl;
         std::cin.get();
-        vector<ImageRequest> request = { ImageRequest("0", ImageType::Scene), ImageRequest("1", ImageType::DepthPlanar, true) };
+        const vector<ImageRequest> request = { ImageRequest("0", ImageType::Scene), ImageRequest("1", ImageType::DepthPlanar, true) };
```
rajat2004, line 36: To be consistent (and good C++ practices as well), let's use `std::vector`

HelloDrone/main.cpp
zimmy87: I agree, `if (!response.empty())` seems like the best choice here.

HelloCar/main.cpp
BillyFrcs: Alright Sir, thanks for that review.

Examples/main.cpp
BillyFrcs: Ah i see.

HelloSpawnedDrones/HelloSpawnedDrones.cpp
BillyFrcs: actually, I'm not sure to use ``unique_ptr`` so i revert it back to ```const```, but anyway i'll try to use ```unique_ptr``` to updated it.

BillyFrcs: hey @zimmy87 i've already use ```unique_ptr``` so it'll looks like this ```std::unique_ptr<MultirotorRpcLibClient> client(new MultirotorRpcLibClient(host, port, timeout_s));``` and in line 113 i use ```client.reset(nullptr)``` instead of ```delete client``` cuz it will get an error while use it, & the error says something like this ```expression must be a pointer to a complete object type```. Wdyt?

rajat2004: No need for the `client.reset(nullptr)`, it'll automatically get cleaned up when the function exits, [this page](https://www.learncpp.com/cpp-tutorial/stdunique_ptr/) might be helpful to understand more of how things work (the entire site is great), and for more details you can check out cppreference as well

BillyFrcs: Alright understood!

Examples/main.cpp
```diff
@@ -152,21 +151,22 @@ void runDepthNavSGM()
...
-    p_state = new CStateStereo();
-    p_state->Initialize(params, depthNav.params_.depth_height, depthNav.params_.depth_width);
+
+    p_state = new CStateStereo;
+    p_state->Initialize(params, static_cast<int>(depthNav.params_.depth_height), static_cast<int>(depthNav.params_.depth_width));
 
     depthNav.gotoGoalSGM(goalPose, client, request, p_state);
 
     //Cleanup
     delete p_state;
```
rajat2004, line 166: Maybe my earlier suggestion of a normal local var wasn't very clear, and the changes got reverted somehow, I meant something like this (not tested for compilation) -
```cpp
CStateStereo p_state;
p_state.Initialize(params, depthNav.params_.depth_height, depthNav.params_.depth_width);

depthNav.gotoGoalSGM(goalPose, client, request, &p_state);
```

BillyFrcs: Ah no problem, so do i need to clean the variable with ```delete &p_state``` or remove it?

rajat2004: No, delete is not required after this, manual memory cleanup is not needed for stack vars. I would suggest checking out some resources if you're new to C++, such as https://www.learncpp.com/cpp-tutorial/dynamic-memory-allocation-with-new-and-delete/

BillyFrcs: Alright i got it, thanks for the resources.

zimmy87 (approved):
Latest revision looks good to me

PR #4408, opened 2022-03-12
Add support for building (x86_64) on Apple Silicon

mbullington:
Fixes: https://github.com/microsoft/AirSim/issues/3696 https://github.com/microsoft/AirSim/issues/3691

## About
- This PR adds a solution for the above tickets (which are closed but require out-of-tree script modifications). The goal is for AirSim to compile and run under x84_64 on Apple Silicon machines.
  * Use `brew --prefix` instead of `/usr/local` in case of M1, which resides at `/opt/homebrew`.
  * Specify CMake flag https://cmake.org/cmake/help/latest/variable/CMAKE_APPLE_SILICON_PROCESSOR.html, which basically just tells CMake to use x86_64 (Rosetta) on M1 machines. This allows linking to succeed with the x86 versions of UE and Unity.
- Additionally, in the test project, allow the user to specify `UnrealDir` and have that be used.

## How Has This Been Tested?

This was tested on my local machine, with the following specs:

- M1 MacBook Air
- macOS Monterey
- Unreal Engine 4.27

## Screenshots (if appropriate):

PR #4409, opened 2022-03-12
Add dedicated docs page for building on macOS

mbullington:
Fixes: #    

## About

This PR is related to, but does not depend on, https://github.com/microsoft/AirSim/pull/4408.

I helped a friend build AirSim on their M1 machine and found the combined page for Linux & macOS somewhat confusing. This PR treats them as separate pages and platforms.

I also updated the guidance to use UE 4.27, which includes fixes for XCode and PrivateFrameworks changes in macOS Big Sur+.

## How Has This Been Tested?

Following these instructions, along with #4408, led to a successful AirSim installation on the following specs:

M1 MacBook Air
macOS Monterey
Unreal Engine 4.27

N/A

### DIFF
--- a/docs/build_macos.md
+++ b/docs/build_macos.md
@@ -33,6 +33,8 @@ cd AirSim
 
 By default AirSim uses clang 8 to build for compatibility with UE 4.25. The setup script will install the right version of cmake, llvm, and eigen.
 
+CMake 3.19.2 is required for building on Apple Silicon.
+
 ```bash
 ./setup.sh
 ./build.sh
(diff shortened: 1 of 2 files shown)


rajat2004:
You'll also need to add the new page to [mkdocs.yml](https://github.com/microsoft/AirSim/blob/master/mkdocs.yml) so that it appears on the site. 
If yo want to see the generated page locally, try following the steps in [test_docs.yml worflow](https://github.com/microsoft/AirSim/blob/master/.github/workflows/test_docs.yml) to install mkdocs (virtualenv is recommended), and then run `build_docs.sh` (skip the API ref docs), and then run `mkdocs serve` to generate a live site locally

PR #4410, opened 2022-03-14
Update unity_api_support.md

zimmy87:
update documentation of unity API support

Fixes: #    

## About
This PR updates the documentation listing unity API support. All APIs defined in client.py should be included now.

## How Has This Been Tested?
Preview tab of web editor.

## Screenshots (if appropriate):

PR #4426, opened 2022-03-21
Fix the problem in ros: can't take off after reset

### LINKED ISSUE #4424: How can I get collision information with AirSim ROS Wrapper?
### What's your question?
How can I get collision information with AirSim ROS Wrapper?

### Include context on what you are trying to achieve

I want to use AirSim in a drl-based navigation system. I just want to get a collision trigger and the give my agent a negative reward , then reset the env.  

#### Context details
AIrSim version: #4410 
Unreal version: v1.7.0-linux
OS version: ubuntu18.04 with ros-melodic
Use default settiongs :AirSim/ros/settings/front_stereo_and_center_mono.json in #4410 

### Include details of what you already did to find answers
By the way, I test reset service manually use rqt in ros, it can't take off after reset service being called. I see a commit and merge in the early version, but it may dose't work?

LucasMa00:
Fixes: #4424    

can't take off after reset in ros

## About
https://github.com/microsoft/AirSim/issues/4424
 I test reset service manually use rqt in ros, it can't take off after reset service being called. I see a commit and merge in the early version, but it may dose't work?

## How Has This Been Tested?

According to the api doc:
`reset: This resets the vehicle to its original starting state. Note that you must call enableApiControl and armDisarm again after the call to reset.`

AIrSim version: https://github.com/microsoft/AirSim/pull/4410
Unreal version: v1.7.0-linux
OS version: ubuntu18.04 with ros-melodic
Use default settiongs :AirSim/ros/settings/front_stereo_and_center_mono.json in https://github.com/microsoft/AirSim/pull/4410

Screenshot from 2022-03-21 21-31-26


ros/src/airsim_ros_pkgs/src/airsim_ros_wrapper.cpp
```diff
@@ -441,10 +443,17 @@ bool AirsimROSWrapper::land_all_srv_cb(airsim_ros_pkgs::Land::Request& request,
 bool AirsimROSWrapper::reset_srv_cb(airsim_ros_pkgs::Reset::Request& request, airsim_ros_pkgs::Reset::Response& response)
 {
     std::lock_guard<std::mutex> guard(drone_control_mutex_);
-
-    airsim_client_->reset();
-
-    response.success = true;
+    if(airsim_client_!=nullptr){
```
rajat2004, line 446: Is there a reason to have this check? This is created in in the constructor itself, and the ROS service would have failed much earlier if it's null

ros/src/airsim_ros_pkgs/scripts/uav_joy.py
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env python
+import rospy
+from std_msgs.msg import Empty
+from geometry_msgs.msg import Twist
+from sensor_msgs.msg import Joy
+from airsim_ros_pkgs.msg import VelCmd
+from airsim_ros_pkgs.srv import TakeoffRequest,Takeoff
+
+msg = """
+Control the RoboMaster TT (Tello Talent) through Gamepad
```
zimmy87, line 10: Was this latest commit intentional? We're not sure if we want to merge gamepad support for a specific drone. If this script is only relevant to this one drone (Tello Talent), please move this latest commit to a new branch/PR. We'd still like to take the more general reset fix by itself, however.

zimmy87 (changes requested):

PR #4438, opened 2022-03-27
Fix namespace usages in AirLib

rajat2004:
Fixes: #4431    

## About

Removes the `using namespace msr::airlib;` in MultiRotorBase.hpp, and fixes the various compilation errors. Also removes `using namespace std;` from a few places

## How Has This Been Tested?

Compilation

## Screenshots (if appropriate):

PR #4446, opened 2022-03-29
Add ros2 CI

alonfaraj:
Fixes: #    

## About

Add ROS2 CI build to GitHub action, under Ubuntu build.
Notice - ROS2 support Ubuntu 20.04 only.

## How Has This Been Tested?

Run the CI and make sure it pass.


tools/install_ros2_deps.sh
```diff
@@ -0,0 +1,28 @@
...
+
+set -x
+
+DISTRO="$(lsb_release -sc)"
+if [[ "$DISTRO" == "focal" ]]; then
+    ROS_DISTRO="galactic"
+else
+    echo "ROS2 support only ubuntu focal"
+    # We return 0 so we won't fail the CI for Ubuntu 18
+    exit 0
```
rajat2004, line 13: Since we're only running the step for 20.04, we can change this back to exit 1 for correct reporting?

.github/workflows/test_ubuntu.yml
```diff
@@ -31,7 +31,16 @@ jobs:
...
           cd ros
           catkin build -DCMAKE_C_COMPILER=gcc-8 -DCMAKE_CXX_COMPILER=g++-8
-
+      
+      - name: Build ROS2 Wrapper
+        if: matrix.os == 'ubuntu-20.04'
+        run: |
+          chmod +x ./tools/install_ros2_deps.sh
+          ./tools/install_ros2_deps.sh
+          source /opt/ros/*/setup.bash
```
rajat2004, line 40: CI is working fine so isn't an issue, but just thinking whether this can cause issues with the ROS1 installation in the same folder as well

rajat2004 (commented):
Thanks a lot for this, will be great for testing #4438 as well! Just a small nitpick comment

tools/install_ros2_deps.sh
alonfaraj: absolutely, thanks

.github/workflows/test_ubuntu.yml
alonfaraj: Do you mean `source /opt/ros/*/setup.bash`?
ROS and ROS2 can coexistence on the same machine as far as I know

jonyMarino:
Thanks a lot @alonfaraj! In time and very helpful!

.github/workflows/test_ubuntu.yml
dzywater: Yes, I am sure ROS and ROS2 can be coexistent. I often brige ros1 msgs to ros2 for compatible with other ros1's algorithms.

PR #4457, opened 2022-04-04
Fix detection crash

alonfaraj:
Fixes: #   https://github.com/microsoft/AirSim/discussions/4439 

## About

Fix a crash in detection object filter.
More details can be found in the related issues.


jonyMarino:
Thanks, @alonfaraj!

PR #4458, opened 2022-04-04
Update Unity Car Demo Scene & Fixed Some Error and Warning ⚠ 

BillyFrcs:
Fixes: Error & Warning

1. Fixed ``MissingReferenceException: The object of type 'Car' has been destroyed but you are still trying to access it`` error.
2. Fixed "possible loss of fraction" warning!

## About
1. Convert parts of loop's body to LINKQ expression.
3. Avoid to use ``GetComponent<Rigidbody>()``
4. Use object initialize.
5. Set some variable to ``readonly``
6. Change some ``public`` method to ``private`` because it's not being use in the other class.
7. Fixed "possible loss of fraction" warning!
8. Convert to lambda expression instead of delegate.
9. Remove unused variable.
10. Fixed ``MissingReferenceException: The object of type 'Car' has been destroyed but you are still trying to access it`` error.

## How Has This Been Tested?

Seems like it should be fine without any error.

- From this

- To This


zimmy87:
Tested this locally for regressions and it worked fine for me, so I am merging this. Thanks for the contribution @BillyFrcs!

BillyFrcs:
No problem Sir!

PR #4460, opened 2022-04-05
Fix Unreal Engine deprecations

jmackay2:
Fixes: This doesn't fix all of the issues to upgrading to a newer UE version, particularly the upgrade form physx to chaos in UE 5.

## About
Fixing some of the deprecated features in UE 4.27 and 5. 

## How Has This Been Tested?
Except for physx issues on the car, this builds on UE 5 preview. 
This should be backward compatible with at least UE 4.26. I should go back and test with the current recommended version of 4.25. 

### DIFF
--- a/Unreal/Plugins/AirSim/Source/AirBlueprintLib.cpp
+++ b/Unreal/Plugins/AirSim/Source/AirBlueprintLib.cpp
@@ -455,8 +455,12 @@ std::vector<msr::airlib::MeshPositionVertexBuffersResponse> UAirBlueprintLib::Ge
 
         //Various checks if there is even a valid mesh
         if (!comp->GetStaticMesh()) continue;
-        if (!comp->GetStaticMesh()->RenderData) continue;
-        if (comp->GetStaticMesh()->RenderData->LODResources.Num() == 0) continue;
+        if (!comp->GetStaticMesh()->HasValidRenderData()) continue;
+        #if (ENGINE_MAJOR_VERSION >= 4 && ENGINE_MINOR_VERSION > 26)
+            if (comp->GetStaticMesh()->GetRenderData()->LODResources.Num() == 0) continue;
+        #else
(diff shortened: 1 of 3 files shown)


zimmy87:
It seems that your PR is failing the clang-format check. Please follow these steps to resolve this:

1. check out your pull request
1. run `git checkout master .clang-format` in your repo's root folder
1. if you don't have npm installed, download it from https://www.npmjs.com/get-npm
1. run `npm install -g clang-format`
1. run `clang-format -i path/to/file` for each failing file (failing files can be found in the details for the clang-format check)
1. run `git add -u` to track all style changes
1. run `git commit -am "apply style from clang-format"`
1. run `git merge master`
1. resolve other non-style conflicts if you have any
1. push changes

Thanks for helping us with these style changes!

zimmy87:
In addition to the clang-format check, it looks like this PR is breaking the 4.25 build with the following error:

```
C:/__w/13/s/Unreal/Environments/Blocks/Plugins/AirSim/Source/Recording/RecordingFile.cpp(4): fatal error C1017: invalid integer constant expression
```

This is happening on both Ubuntu & Windows, can you investigate this failure and resolve it with a new commit?

jmackay2:
@zimmy87 Thank you for pushing this PR across the finish line!

I am curious about the need to revert the RecordingFile.cpp. I was pretty confident that the header file capitalization changed in UE 5 and was pretty sure I tested that the change was required for UE 5.

[UE5 reference](https://docs.unrealengine.com/5.0/en-US/API/Runtime/Core/HAL/FPlatformFileManager/)

zimmy87:
Hey @jmackay2, we reverted RecordingFile.cpp because it was breaking the Azure build check with this error: `C:/__w/13/s/Unreal/Environments/Blocks/Plugins/AirSim/Source/Recording/RecordingFile.cpp(4): fatal error C1017: invalid integer constant expression`. We didn't realize the capitalization change at the time, so we felt it was a reasonable way of fixing the error in the Azure build check.

PR #4465, opened 2022-04-06
fix4445

### LINKED ISSUE #4445: dist_msg.max_range  is given the wrong value in ROS2
https://github.com/microsoft/AirSim/blob/78d8cf732ad08993183699cadf33d084bc5b176d/ros2/src/airsim_ros_pkgs/src/airsim_ros_wrapper.cpp#L760
It should be "dist_msg.max_range = dist_data.max_distance;"

dzywater:
fixes: #4445


dzywater:
fix #4445

alonfaraj:
Hi @dzywater,
Seems like a good fix.
I think it worth apply that fix to ROS1 as well.

dzywater:
> Hi @dzywater, Seems like a good fix. I think it worth apply that fix to ROS1 as well.

I not use ros1. If ros1 and ros2 have similar implementation, I want to optimize ros2 first.

zimmy87:
Thanks for submitting this fix @dzywater! Tested this locally and it works for me, so I am moving ahead with merging. Applying this fix to ROS1 can be handled separately in a different PR.

PR #4509, opened 2022-05-03
Fix missing vehicle name in ros camera msg

alonfaraj:
Fixes: #    

## About

This PR fix missing vehicle name in published Image camera header.
Applied for ros1 and ros2.

## How Has This Been Tested?

Without this fix, it's not possible to show the image in rviz2 on the map/world frame.
Run ros wrapper and show the image in rviz2 under the world_enu frame.

### DIFF
--- a/ros/src/airsim_ros_pkgs/src/airsim_ros_wrapper.cpp
+++ b/ros/src/airsim_ros_pkgs/src/airsim_ros_wrapper.cpp
@@ -1421,13 +1421,13 @@ void AirsimROSWrapper::process_and_publish_img_response(const std::vector<ImageR
         if (curr_img_response.pixels_as_float) {
             image_pub_vec_[img_response_idx_internal].publish(get_depth_img_msg_from_response(curr_img_response,
                                                                                               curr_ros_time,
-                                                                                              curr_img_response.camera_name + "_optical"));
+                                                                                              vehicle_name + "/" + curr_img_response.camera_name + "_optical"));
         }
         // Scene / Segmentation / SurfaceNormals / Infrared
         else {
             image_pub_vec_[img_response_idx_internal].publish(get_img_msg_from_response(curr_img_response,
                                                                                         curr_ros_time,
-                                                                                        curr_img_response.camera_name + "_optical"));
+                                                                                        vehicle_name + "/" + curr_img_response.camera_name + "_optical"));
         }
         img_response_idx_internal++;
     }
--- a/ros2/src/airsim_ros_pkgs/src/airsim_ros_wrapper.cpp
+++ b/ros2/src/airsim_ros_pkgs/src/airsim_ros_wrapper.cpp
@@ -1335,13 +1335,13 @@ void AirsimROSWrapper::process_and_publish_img_response(const std::vector<ImageR
         if (curr_img_response.pixels_as_float) {
             image_pub_vec_[img_response_idx_internal].publish(get_depth_img_msg_from_response(curr_img_response,
                                                                                               curr_ros_time,
-                                                                                              curr_img_response.camera_name + "_optical"));
+                                                                                              vehicle_name + "/" + curr_img_response.camera_name + "_optical"));
         }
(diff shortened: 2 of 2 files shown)


alonfaraj:
Hi @zimmy87, thanks for the review, I missed the camera info somehow.
I'm not entirely sure we get this warning after applying the fix, but according to ros guidelines `camera_info` should be publish under the camera name namespace, in our case `/airsim_node/Drone1/Test0/camera_info`.

I made a fix so camera info is under the camera name namespace.
Also, you right about the frame `Test0_optical/static`, it's should contain the vehicle name and I fixed it as well. It's actually look like a typo bug, you can see here https://github.com/microsoft/AirSim/blob/master/ros2/src/airsim_ros_pkgs/src/airsim_ros_wrapper.cpp#L1213-L1214



The only problem I encounter now regarding the camera info is in case we have 2 different `CaptureSettings `for different image types.
Ideally, we should have only one camera info under the camera name, but in case we have `Scene` and `Segmentation` with different resolutions, we end up with 2 publishers `/airsim_node/Drone1/Test0/camera_info` which publish inconsistent data.

So we do need to change the frame id of the image msg, then we have to change the camera info topic to under the camera name namespace, but it will cause a problem when we have different `CaptureSettings`.
What do you think?

jonyMarino:
Hi @alonfaraj!. Thanks for yet another contribution. If I understand what you are saying correctly, we should have a topic for each CaptureSettings of each camera. If so, why don't we add the ImageType of each CaptureSettings to the camera name? In this case, it would be /airsim_node/Drone1/Test0_0, where the last 0 would be the ImageType

alonfaraj:
Hi @jonyMarino, that's correct.
I have implemented the change you suggested which is probably the most reasonable.




2 things up to your decision:
1. To follow ros convection, I changed the raw image to be published under `/image`,  in our case `/airsim_node/Drone1/Test0_0/image`
2. Should we go with image type as number or explicit name? I think explicit name would be easier. Let me know wdyt

jonyMarino:
Excellent work @alonfaraj!
1. OK
2. Leave it as an explicit name. It would be good to keep correspondence with settings.json; maybe we can give the option to use explicit names also in the configuration.

EvelynM7:
Hello,

I am running into a similar error.

I tried to follow @zimmy87's repro steps as follows with ROS1:

1. Setup with ROS1 from WSL2
2. Edited lines 1424 and 1430 in ros/src/airsim_ros_pkgs/src/airsim_ros_wrapper.cpp by adding {`vehicle_name + "/" +` }
3. Ran {`catkin_make`} in AirSim/ros
4. Opened and played Unreal project with settings.json as @zimmy87 showed
5. Ran {`cd AirSim/ros`} and {`source devel/setup.bash`}
6. Ran {`roslaunch airsim_ros_pkgs airsim_node.launch`}
7. In separate ubuntu terminal, cd to AirSim/ ros, ran {`source devel/setup.bash`}, then ran {`roslaunch airsim_ros_pkgs rviz.launch`} after linking to XLaunch with {`export DISPLAY=$(cat /etc/resolv.conf | grep nameserver | awk '{print $2}'):0`}
8. Added Camera in rviz
9. Edited Camera Topic to be /airsim_node/Drone1/Test0/Scene

This gave me Status: Warn as well. The error for Image under Camera gives "No image received". The error for Camera Infor says "No CameraInfo received on [/airsim_node/Drone1/Test0/camera_info]. Topic may not exist."

@zimmy87 also showed a screenshot with Frames. I only have world_enu and world_ned as options. Could this be part of the error? If so, how do I add the other Frames?

It looks like @alonfaraj may have provided fixes to the errors @zimmy87 was running into, but I am not understanding what I should try to follow the steps.

I've listed the versions I'm using below as well as a couple screenshots that may help to understand what issues I am running into.

Any assistance would be greatly appreciated. Thank you!


Versions:
- Unreal Engine 4.27.2
- Ubuntu 18.04.5 LTS
- WSL 2
- ROS melodic 1

rvis camera warning

rostopic list

PR #4511, opened 2022-05-04
Publish image and camera_info to the same namespace

ajordan5:
Fixes: # 1 Now publishes camera_info to the same namespace as the corresponding namespace. For example, for a camera named 'bottom_center', the ros wrapper will publish 
`.../bottom_center/Scene` and `.../bottom_center/camera_info`

## About

Previously these topics were published as:
`.../bottom_center/Scene` and `.../bottom_center/Scene/camera_info`

Many ros packages expect the camera info to be in the same namespace as the image such as [apriltag_ros](https://github.com/AprilRobotics/apriltag_ros). In the case of apriltag it cannot find the camera info and will not calculate pose information.

## How Has This Been Tested?

I have run the airsim node in my own airsim environment that includes an apriltag on a moving platform. This new publishing format is indeed the format expected by apriltag_ros and correct pose info is calculated. I expect most other ROS interfaces will require the same.

image
image


dzywater:
I wanted to use one camera_info for one camera before.

But in settings.json, different image types can be configured different width/height/fov, so respective camera_info may be reasonable.
 
I don't how to get a good resolution.

PR #4520, opened 2022-05-12
Fix missing sensor name in ros PointCloud2 msg

alonfaraj:
## About
Fix missing sensor name in PointCloud2 frame ID.
Applied for ROS & ROS2.

Before: 
`my_car_name`

After: 
`my_car_name/my_lidar1`

## How Has This Been Tested?

Run ros wrapper with lidar configuration and echo the appropriate msg to check the frame id


zimmy87:
Thanks for fixing this @alonfaraj! I tested this locally and it's working for me, so I'm moving ahead with merging.

dzywater:
I wanted to pull request before, but this request is right only if the DataFrame filed is set as "SensorLocalFrame" in settings.json.

dzywater:
Moreover, like camera tf, the lidar also add static tf "VehicleName/LidarName/static"

PR #4527, opened 2022-05-17
API improvement: Independent gimbal angle control 

ibrahim-aydin:
# Context
Improving pythonclient, rpc and unreal apis to allow independent control of each gimbal angle. This update allows the control of each gimbal angle separately for a given camera. Therefore we won't need to pass a Pose struct for every and every gimbal control action. This will minorly improve performance and increase camera control flexibility. Docs update is also included in this PR

## How Has This Been Tested?
Tested through a simple gimbal control python script

### DIFF
--- a/AirLib/include/api/RpcLibClientBase.hpp
+++ b/AirLib/include/api/RpcLibClientBase.hpp
@@ -145,6 +145,9 @@ namespace airlib
         void simSetDistortionParam(const std::string& camera_name, const std::string& param_name, float value, const std::string& vehicle_name = "", bool external = false);
         std::vector<float> simGetDistortionParams(const std::string& camera_name, const std::string& vehicle_name = "", bool external = false);
         void simSetCameraPose(const std::string& camera_name, const Pose& pose, const std::string& vehicle_name = "", bool external = false);
+        void simSetCameraPitch(const std::string& camera_name, float pitch, const std::string& vehicle_name = "", bool external = false);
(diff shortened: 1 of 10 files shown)


rajat2004:
IMHO it's not really worth it adding extra APIs just for controlling one specific parameter, going this way will lead to explosion in number of APIs. Do you have any numbers for the performance improvement, doubt there would be much difference.
If the current pose API doesn't handle controlling independent angles, maybe it can be enhanced to only apply non-zero control values. But zero pitch or roll is also valid and something which the user wants to apply, maybe an optional argument (like a bitset), indicating which clause the user wants to control? Wdyt

xubeyb:
I agree with you. But for a gimbal, as most gimbal controller APIs allow, controlling each angle separately would be useful. Tracking an object during maneuver can be given as an example. This helped me in my application. I thought it would be good to help others too. 
By the way your idea also sounds positive to me, advancing the simSetCameraPose by adding a typemask argument to select control of each angle would also be good. But might increase complexity for the user. I can update my implementation that way. Your opinions would be appreciated

xubeyb:
@rajat2004 Any thoughts on this?

rajat2004:
Sorry, missed the reply. Bitmask seems to be the better option to me atleast. What do you think @zimmy87 @jonyMarino

jonyMarino:
Thanks for the contribution and the debate. I think the current implementation is ok. I would refactor the client APIs to get objects with specialized APIs (for example, for cameras). But that is out of the scope of this contribution. I will continue thinking about this.

PR #4541, opened 2022-05-23
Travis/move drone on path

tnilest:
updated path.py to allow a single drone agent to move to predetermined coordinates in hallway example environment.


jonyMarino:
Hi @tnilest! It seems that you have added some libraries by accident.

PR #4526, opened 2022-05-17
Improving build.sh to recursively update plugins of all environments

ibrahim-aydin:
## What is this about and what it fixes/improves
Improving environment plugin update by recursively updating all projects under Unreal/Environments folder. This small update fixes the [build.sh#L150](https://github.com/microsoft/AirSim/blob/e8c7ea362bcc34a045a5c70457ecc6a91a2d56ec/build.sh#L150).

## How Has This Been Tested?
This was tested by in a scenario where multiple folders were under Unreal/Environments and no folders at all.

PR #4546, opened 2022-05-26
update all environments during build in Windows

zimmy87:
Fixes: #    

## About
This PR applies the same fix in #4526 to the Windows build script (build.cmd). This also updates the documentation to remove an explicit call to update_from_git.bat in the Windows workflow.

## How Has This Been Tested?
A build was ran with the new build.cmd and two environments: Blocks and a copy of Blocks under another name. Both environments were confirmed to startup and run after the build.


jonyMarino:
Thanks, @zimmy87 !

PR #4551, opened 2022-05-30
Fix simLoadLevel and Levels OptionsMenu

### LINKED ISSUE #4432: simLoadLevel does not correctly switch between maps
## Bug report 

- AirSim Version/#commit: e692a6f81d9b9bc17a0cea59592d0504c6222575
- UE/Unity version: UE 4.27
- autopilot version: N/A
- OS Version: Ubuntu 20.04

I'm looking to get AirSim working with multiple environments in the same binary. I've followed the tutorials to get custom environments into AirSim and am able to run them through the editor. However, I would like to be able to switch between the environments without restarting AirSim.

I came across `simLoadLevel` which looks like it is supposed to do this but fails to do so. It seems like this is oriented towards streaming levels- so increasing the size of an already loaded persistent level rather than load and replace a currently loaded level.

There is an undocumented param in the settings.json file, "Default environment", which is supposed to be able to allow a specific environment to be loaded on AirSim startup but that does nothing either.

I have also tried calling the `simLoadLevel` function from the python client with the environment name. This works but loads the new environment on top of the current environment- like a streaming level.

1. Have multiple environments in a single project
2. Attempt to use `simLoadLevel` to load a map that is not yet loaded

No errors.

### LINKED ISSUE #4542: simLoadLevel (C++ API) doesn't destroy assets before loading new Level
## Bug report 

- AirSim Version/#commit: Release 1.7.0
- UE/Unity version: 4.25.4
- autopilot version:
- OS Version: Windows 10

Using the following command
```C++
    client.simLoadLevel("Acceleration");
```

Assets aren't destroyed in the new map

Before load:
BeforeLoading
After load
AfterLoading

Furthermore, the origin point after the loading isn't the origin of the new map but the origin of the previous map (tested loading another map)

### Settings

```json
{
  "SettingsVersion": 1.2,
  "SimMode": "Car",
  "ClockSpeed": 1,
  "PawnPaths": {
    "DefaultCar": { "PawnBP": "Class'/AirSim/VehicleAdv/Cars/TechnionCar/TechnionCarPawn.TechnionCarPawn_C'" }
  },

  "SubWindows": [
    {"WindowID": 0, "CameraName": "ZED2", "ImageType": 0, "VehicleName": "WShabtiCAR", "Visible": true, "External": false},
    {"WindowID": 2, "CameraName": "TestingCamera", "ImageType": 0, "VehicleName": "WShabtiCAR", "Visible": true, "External": false}
  ],
  "Recording": {
    "RecordOnMove": true,
    "RecordInterval": 0.05,
    "Folder": "",
    "Enabled": false,
    "Cameras": [
        { "CameraName": "ZED2", "ImageType": 0, "PixelsAsFloat": false,  "VehicleName": "WShabtiCAR", "Compress": true },
        { "CameraName": "TestingCamera", "ImageType": 0, "PixelsAsFloat": false,  "VehicleName": "WShabtiCAR", "Compress": true }
    ]
  },
  "Vehicles": {
    "WShabtiCAR": {
      "VehicleType": "PhysXCar",
      "DefaultVehicleState": "",
      "PawnPath": "DefaultCar",
      "EnableCollisionPassthrogh": false,
      "EnableCollisions": true,
      "AllowAPIAlways": true,
      "RC": {
        "RemoteControlID": -1
      },
      "Cameras": {
        "ZED2": {
          "CaptureSettings": [
          {
              "ImageType": 0,
              "Width": 256,
              "Height": 144,
              "FOV_Degrees": 110
          }
          ],
          "X": -1.3,
          "Y": 0,
          "Z": -1.2,
          "Pitch": 0,
          "Roll": 0.0,
          "Yaw": 0
        },
	"TestingCamera": {
          "CaptureSettings": [
          {
              "ImageType": 0,
              "Width": 640,
              "Height": 480,
              "FOV_Degrees": 110
          }
          ],
          "X": -5,
          "Y": 0,
          "Z": -2,
          "Pitch": 0.3,
          "Roll": 0.0,
          "Yaw": 0
        }  
      },
      "Sensors":
      {
        "Imu" : {
          "SensorType": 2,
          "Enabled": true
        },
        "Gps" : {
          "SensorType": 3,
          "Enabled": true
        },
        "Magnetometer": {
          "SensorType": 4,
          "Enabled" : true
      }
      }
    }
  }
}

```

Just use the C++ API calling the loadLevel command in the client RPC class

alonfaraj:
Fixes: https://github.com/microsoft/AirSim/issues/4432
Fixes:  https://github.com/microsoft/AirSim/issues/4542

## About

This PR fix load level mechanism, simplify and cleanup the whole process.  
`OptionsMenu` was fixed as well to list the available maps under `/Game` folder only.
Till now, load level was loading the level as streaming level on top of the existing level.  
As far as I understand, this logic assume a level is always exists and keep loading the rest of the levels as streaming levels which is not the desired behavior.  
I have used a simpler method, `UGameplayStatics::OpenLevel` instead of `AirsimLevelStreaming` class which is not needed. 
This method take care of unload and load the requested level, and make it Persistent. 

## How Has This Been Tested?
Run the blocks project and switch between 2 levels with different `PlayerStart` positions, as well as using the python API.
Tested on packaged game (windows) as well.
To test it on package, you should specify the desired maps:
`Project settings -> packaging -> List of maps to include in packages build
`

https://user-images.githubusercontent.com/15960497/171026001-03812a8c-be92-4d20-9bc2-7eccf8206f67.mp4


jonyMarino:
@alonfaraj Thanks for this contribution. I agree that this is the right approach. Thanks for the detailed explanation. I didn't remember where was the list to include other levels 😄

PR #4558, opened 2022-06-06
Fix settings FOV not applied

### LINKED ISSUE #4557: FOV from settings.json not applied
## Bug report 

- AirSim Version/#commit: b8b8e0d42 and release 1.7.0
- UE/Unity version: UE 4.27.2
- autopilot version: N/A
- OS Version: Ubuntu 20.04

Changing the ```FOV_Degrees``` attribute in ```Vehicles -> drone_1 -> Cameras -> img0 -> CaptureSettings``` in the settings.json file (see below) has no effect on the FOV during the simulation. I've set it to ```10``` and to ```100``` and the images produced inside the subwindow, as well as images captured with ```MultirotorClient.simGetImages()``` always produce the same image.

When executing the following python snipped you can see that the horizontal FOV is always set to 89.9036, no matter the value in the settings.json file.

```python
import airsim
client = airsim.VehicleClient()
print(client.simGetCurrentFieldOfView("img0"))
```

However, setting the FOV at runtime through the Python API does indeed change the FOV, as shown in the subwindow and images captures with ```MultirotorClient.simGetImages()```:

```python
import airsim
client = airsim.VehicleClient()
client.simSetCameraFov("img0", 10)
print(client.simGetCurrentFieldOfView("img0"))
```

### Attempts to fix it

I've changed attributes ```Width``` and ```Height``` in the settings.json file. These changes were instantly reflected after restarting the simulation. So that works.

Also, I've played around with the ```CameraDefaults``` attribute to set the FOV, but without any success.

We have another system running AirSim 1.4. I tested the FOV settings there and it worked.

### Settings

<details><summary>settings.json</summary>
<p>

```json
 {
  "SettingsVersion": 1.2,
  "ClockSpeed": 0.2,
  "SimMode": "Multirotor",
  "PhysicsEngineName": "ExternalPhysicsEngine",
  "ViewMode": "Manual",
  "DefaultSensors": {
    "imu_1": {
         "SensorType": 2,
         "Enabled" : true
    }
  },
  "Vehicles": {
    "drone_1": {
      "VehicleType": "SimpleFlight",
      "X": 0, "Y": 0, "Z": 0,
      "Cameras": {
        "img0": {
          "CaptureSettings": [
            {
              "ImageType": 0,
              "Width": 1024,
              "Height": 1024,
              "FOV_Degrees": 120
            }
          ],
          "X": 0.5, "Y": -0.06, "Z": 0.10,
          "Pitch": 0.0, "Roll": 0.0, "Yaw": 0.0
        }
      }
    }
  },
  "OriginGeopoint": {
      "Latitude": 0.0,
      "Longitude": 0.0,
      "Altitude": 0.0
    },
  "SubWindows": [
    {"WindowID": 0, "ImageType": 0, "CameraName": "img0", "Visible": true}
  ]
}
```

</p>
</details>

1. Open Blocks environment
2. Copy my settings.json file to your Documents/AirSim directory
3. Change FOV in settings.json to random number [10, 100]
4. Start simulation
5. Check captured image in subwindow
6. Stop simulation
7. Go to 3

### Others
Thank you for your help!

alonfaraj:
Fixes: https://github.com/microsoft/AirSim/issues/4557

## About
With this PR, an issue was fixed that caused the FOV from a setting to be overridden when creating a custom camera.
More details here https://github.com/microsoft/AirSim/issues/4557.

I'm not sure what the original intent of this line is, but it causes the requested FOV to be overridden by the default.
 At this point, `dst` already has set the right FOV while `src` is not.
https://github.com/microsoft/AirSim/blob/caa13182328e866564c09fdeebca89ba024bc323/Unreal/Plugins/AirSim/Source/PIPCamera.cpp#L781 

## How Has This Been Tested?
Run with the original issue settings and make sure the settings actually change the FOV.

### DIFF
--- a/Unreal/Plugins/AirSim/Source/PIPCamera.cpp
+++ b/Unreal/Plugins/AirSim/Source/PIPCamera.cpp
@@ -778,7 +778,6 @@ void APIPCamera::copyCameraSettingsToSceneCapture(UCameraComponent* src, USceneC
 {
     if (src && dst) {
         dst->SetWorldLocationAndRotation(src->GetComponentLocation(), src->GetComponentRotation());
-        dst->FOVAngle = src->FieldOfView;
 
         FMinimalViewInfo camera_view_info;
         src->GetCameraView(/*DeltaTime =*/0.0f, camera_view_info);


jonyMarino:
Thanks, @alonfaraj!

PR #4365, opened 2022-02-22
fix: compile on Ubuntu 21.10 and Unreal 4.27.2

russkel:
Fixes: #    

- Not compiling on Ubuntu 21.10 and Unreal 4.27.2

## How Has This Been Tested?
Successfully compiled on the machine described above.

## Screenshots (if appropriate):

PR #4561, opened 2022-06-08
Copy of fix: compile on Ubuntu 21.10 and Unreal 4.27.2

zimmy87:
Fixes: #    

## About
This PR recreates the changes in @russkel's #4365 along with fixes for the CR feedback. This PR also adds ubuntu 21.10 to the github checks so that we can test the build on that platform.

## How Has This Been Tested?
Built locally


russkel:
Thanks a lot for this @zimmy87. As a side note 22.04 should probably replace 21.10.

I do not use AirSim anymore so I cannot afford to maintain these patches.

PR #4710, opened 2022-09-29
Fix .binvox files being saved with the inverse scale

sotpapathe:
Fixes: no related issue

## About
According to the [binvox specification](https://www.patrickmin.com/binvox/binvox.html) the scale stored in binvox should be what the normalized voxel coordinates (in the interval [0, 1]) should be scaled by.

>binvox now includes two extra lines in the header (which may be omitted, viewvox and thinvox don't need them):
>translate \<t_x> <t_y> <t_z>
>scale \<scale factor>
>First scale (x_n, y_n, z_n) by the scale factor, then translate them by (t_x, t_y, t_z)

This is also how the scale is interpreted in [this](https://github.com/dimatura/binvox-rw-py/blob/public/binvox_rw.py#L79) Python binvox implementation.

## How Has This Been Tested?
Tested by using other implementations to read the resulting `.binvox` files.


Bartok9:
Thanks @sotpapathe — re-landed your binvox scale fix on current `main` as salvage PR https://github.com/microsoft/AirSim/pull/9792 with full credit to you/#4710 so maintainers have a fresh head. Happy to close mine in favor of yours if you prefer to push an update instead.

